PR #27, opened 2026-04-06
docs: add Simplified Chinese README

cyxer000:
## Summary
- add `README.zh-CN.md` as a Simplified Chinese translation of the project README
- add a language switcher to `README.md` so readers can discover the Chinese version
- keep the change documentation-only and aligned with the existing structure

## Test plan
- [x] `git diff --check`
- [x] Verified README links and file structure locally

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Agent Skills
 
+[English](README.md) | [简体中文](README.zh-CN.md)
+
 **Production-grade engineering skills for AI coding agents.**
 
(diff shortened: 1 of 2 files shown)


addyosmani:
Thank you very much for this contribution. It's greatly appreciated. I'm going to have to decline the contribution unfortunately as we will be unable to keep the internationalized version in sync without significant consistent community contributions. We'll definitely circle back here if we change positions on this.

PR #7, opened 2026-03-18
Add simplify-ignore hook for block-level code protection

federicobartoli:
## Summary

Implements the `simplify-ignore` hook requested in #2 — a single bash script that hides annotated code blocks from the model during `/code-simplify` sessions.

- **Hash-based tokens**: blocks become `BLOCK_<8-hex-hash>` placeholders (content-addressed, unambiguous round-trip)
- **Multi-block support**: arbitrary number of protected blocks per file
- **Optional reason strings**: `/* simplify-ignore-start: perf-critical */` → `/* BLOCK_de115a1d: perf-critical */`
- **Language-aware placeholders**: preserves comment syntax per language (`/* */`, `#`, `<!-- -->`)

### Hook flow

| Event | Action |
|---|---|
| `PreToolUse Read` | Backs up file, replaces blocks with `BLOCK_<hash>` placeholders in-place |
| `PostToolUse Edit\|Write` | Expands placeholders back, applies model's changes, re-filters |
| `Stop` | Restores all files from backup when session ends |

### Setup required

The hook script needs to be wired into `.claude/settings.json` to work. The `/code-simplify` command should register these hooks automatically, or users can add them manually:

```json
{
  "hooks": {
    "PreToolUse": [
      { "matcher": "Read", "hooks": [{ "type": "command", "command": "bash ${CLAUDE_PROJECT_DIR}/hooks/simplify-ignore.sh" }] }
    ],
    "PostToolUse": [
      { "matcher": "Edit|Write", "hooks": [{ "type": "command", "command": "bash ${CLAUDE_PROJECT_DIR}/hooks/simplify-ignore.sh" }] }
    ],
    "Stop": [
      { "hooks": [{ "type": "command", "command": "bash ${CLAUDE_PROJECT_DIR}/hooks/simplify-ignore.sh" }] }
    ]
  }
}
```

### Files

- `hooks/simplify-ignore.sh` — unified hook script (~275 lines)
- `hooks/SIMPLIFY-IGNORE.md` — setup, annotation syntax, crash recovery, known limitations
- `.gitignore` — added `.claude/.simplify-ignore-cache/`

Closes #2

PR #30, opened 2026-04-07
Fix simplify-ignore bugs from PR #7 review

Ic3b3rg:
## Summary

Addresses the three issues @addyosmani flagged in the [PR #7 review](https://github.com/addyosmani/agent-skills/pull/7#issuecomment-4146310951):

- **Silent file restoration (data-loss bug):** piping `{}` or any JSON without `tool_name` triggered the Stop/restore block, silently reverting all protected files. Fixed by parsing `hook_event_name` explicitly and adding a guard that warns instead of restoring on ambiguous input.
- **Single-line block fragility:** the nested `case` + `continue` pattern worked but relied on a non-obvious control flow. Added a `handled` flag as defense-in-depth so the end-marker check and the normal-line write are both skipped when a single-line block was already processed.
- **Trailing newline (`dd`):** already fixed before merge (uses`perl -pe 'chomp if eof'`). No changes needed.

## Test plan

- [x] 29 assertions pass on macOS Bash 3.2 and default bash
- [x] Manual verification: `echo '{}' | hook` no longer restores files
- [x] Manual verification: `{"hook_event_name":"Stop"}` still restores
- [x] Manual verification: TTY crash recovery (`bash hooks/simplify-ignore.sh`) still works

### DIFF
--- a/hooks/SIMPLIFY-IGNORE.md
+++ b/hooks/SIMPLIFY-IGNORE.md
@@ -73,8 +73,10 @@ Any comment style works (`//`, `/*`, `#`, `<!--`). Multiple blocks per file and 
 If Claude Code crashes without triggering the Stop hook, files on disk may still have `BLOCK_<hash>` placeholders. To restore manually:
 
 ```bash
-echo '{}' | bash hooks/simplify-ignore.sh
+bash hooks/simplify-ignore.sh
 ```
+
+Run from your project root, or set `CLAUDE_PROJECT_DIR` to the project path.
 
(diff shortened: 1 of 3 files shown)


addyosmani:
Thanks for the contribution here @Ic3b3rg! It's much appreciated and also to @federicobartoli for the review.

> Might be worth having Addy weigh in, but from what I can see the current code handles these cases correctly by design.

This is my observation as well. Given this, are we wanting to close this PR out?

Ic3b3rg:
Thanks to both of you, I'm closing the PR now

PR #45, opened 2026-04-08
feat: source-driven-development skill

federicobartoli:
## What this adds

A new **source-driven-development** skill that grounds every framework-specific implementation decision in official documentation.

The core idea: agents shouldn't implement from memory. They should verify, cite, and let the user see the sources behind every non-trivial pattern.

## The problem it solves

AI agents confidently produce code that looks correct but uses outdated patterns — because training data goes stale. The agent doesn't know it's wrong, and neither does the user until they hit a deprecation warning or a subtle runtime bug.

This is especially painful with fast-moving frameworks: the agent writes React 18-style code in a React 19 project, Django 4 patterns in a Django 5 project. The code runs, the tests pass, and the problem surfaces six months later.

## How it works

Four-step process: **DETECT → FETCH → IMPLEMENT → CITE**

1. Read the dependency file to identify exact versions (`package.json`, `go.mod`, `Cargo.toml`, etc.)
2. Fetch the specific documentation page for the feature — not the homepage, the relevant page
3. Implement following the documented patterns, surfacing conflicts with existing code
4. Cite every non-trivial decision with a full URL

Every pattern in the output traces back to a source the user can verify.

## What makes it different from just "read the docs"

- Framework-agnostic — works the same for React, Django, Rails, Go, Rust, Laravel
- Structured citation format in both code comments and conversation
- Explicit `UNVERIFIED:` flag when documentation can't be found
- `CONFLICT DETECTED:` surfacing when docs differ from existing codebase patterns
- Clear "When NOT to use" — doesn't slow down variable renames and file moves

## Integration

- Added to the `using-agent-skills` flowchart and lifecycle sequence
- Added to `CLAUDE.md` Build phase
- Added to `README.md` Build table (20 skills total)

## Demo

Built a SpaceX launch dashboard POC with the skill active. The agent:
- Detected React 19.1.0 from `package.json`
- Fetched `react.dev/reference/react/use` and `react.dev/blog/2024/12/05/react-19`
- Used `use()` for data fetching (not `useEffect`), `<Context>` without `.Provider` (React 19), `useOptimistic` for favorites
- Cited every pattern inline

The same prompt without the skill produces `useEffect` + `useState` — valid React 18 code, wrong for React 19.

## Future work

An earlier version included a local documentation cache (`.agent-sources/`) for reusing fetched docs across sessions. Removed from this PR to keep the skill focused on verification and citation. A caching mechanism could live in a dedicated skill or as an extension to `context-engineering`.

### DIFF
--- a/skills/source-driven-development/SKILL.md
+++ b/skills/source-driven-development/SKILL.md
@@ -1,0 +1,192 @@
+---
+name: source-driven-development
+description: Grounds every implementation decision in official documentation. Use when you want authoritative, source-cited code free from outdated patterns. Use when building with any framework or library where correctness matters.
+---
+
+# Source-Driven Development
+
+## Overview
+
+Every framework-specific code decision must be backed by official documentation. Don't implement from memory — verify, cite, and let the user see your sources. Training data goes stale, APIs get deprecated, best practices evolve. This skill ensures the user gets code they can trust because every pattern traces back to an authoritative source they can check.
+
+## When to Use
+
+- The user wants code that follows current best practices for a given framework
+- Building boilerplate, starter code, or patterns that will be copied across a project
+- The user explicitly asks for documented, verified, or "correct" implementation
+- Implementing features where the framework's recommended approach matters (forms, routing, data fetching, state management, auth)
+- Reviewing or improving code that uses framework-specific patterns
+- Any time you are about to write framework-specific code from memory
+
+**When NOT to use:**
+
+- Renaming variables, fixing typos, moving files
+- Language-level patterns that don't depend on any library (loops, conditionals, data structures)
+- Code that doesn't touch any external framework or library
+- The user explicitly wants speed over verification ("just do it quickly")
+
+## The Process
+
+```
+DETECT ──→ FETCH ──→ IMPLEMENT ──→ CITE
+  │          │           │            │
+  ▼          ▼           ▼            ▼
+ What       Get the    Follow the   Show your
+ stack?     relevant   documented   sources
+            docs       patterns
+```
+
+### Step 1: Detect Stack and Versions
+
+Read the project's dependency file to identify exact versions:
+
+```
+package.json    → Node/React/Vue/Angular/Svelte
+composer.json   → PHP/Symfony/Laravel
+requirements.txt / pyproject.toml → Python/Django/Flask
+go.mod          → Go
+Cargo.toml      → Rust
+Gemfile         → Ruby/Rails
+```
+
+State what you found explicitly:
+
+```
+STACK DETECTED:
+- React 19.1.0 (from package.json)
+- Vite 6.2.0
+- Tailwind CSS 4.0.3
+→ Fetching official docs for the relevant patterns.
+```
+
+If versions are missing or ambiguous, **ask the user**. Don't guess — the version determines which patterns are correct.
+
+### Step 2: Fetch Official Documentation
+
+Fetch the specific documentation page for the feature you're implementing. Not the homepage, not the full docs — the relevant page.
+
+**Source hierarchy (in order of authority):**
+
+| Priority | Source | Example |
+|----------|--------|---------|
+| 1 | Official documentation | react.dev, docs.djangoproject.com, symfony.com/doc |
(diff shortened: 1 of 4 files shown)


skills/source-driven-development/SKILL.md
```diff
@@ -0,0 +1,192 @@
...
+## When to Use
+
+- The user wants code that follows current best practices for a given framework
+- Building boilerplate, starter code, or patterns that will be copied across a project
+- The user explicitly asks for documented, verified, or "correct" implementation
+- Implementing features where the framework's recommended approach matters (forms, routing, data fetching, state management, auth)
+- Reviewing or improving code that uses framework-specific patterns
+- Any time you are about to write framework-specific code from memory
+
+**When NOT to use:**
```
addyosmani, line 21: The "When NOT to use" framing around language-level vs. library patterns feels a bit loose. 

A more reliable signal might be "does correctness depend on a specific version?" I think that might cover more edge cases cleanly.

skills/source-driven-development/SKILL.md
```diff
@@ -0,0 +1,192 @@
...
+
+- Use the API signatures from the docs, not from memory
+- If the docs show a new way to do something, use the new way
+- If the docs deprecate a pattern, don't use the deprecated version
+- If the docs don't cover something, flag it as unverified
+
+**When docs conflict with existing project code:**
+
+```
+CONFLICT DETECTED:
```
addyosmani, line 109: This is great, but it only handles code-vs-docs conflicts. What happens when the docs themselves are inconsistent across versions (e.g. a migration guide that contradicts the reference page)? Might be worth a note.

skills/source-driven-development/SKILL.md
```diff
@@ -0,0 +1,192 @@
...
+I'm using useActionState instead of manual useState for the
+form submission state. React 19 replaced the manual
+isPending/setIsPending pattern with this hook.
+
+Source: https://react.dev/blog/2024/12/05/react-19
+"useTransition now supports async functions [...] to handle
+pending states automatically"
+```
+
+**Citation rules:**
```
addyosmani, line 146: Small one: the citation examples feel like they might point to top-level pages, but docs restructure and anchors are more stable long-term. 

Recommending deep links where possible? (e.g. `/reference/react/useActionState#usage` over just `/reference/react/useActionState`) would make the citations more durable. None of these are blockers  solid skill overall, happy to see it land.

addyosmani (changes requested):
Really like the direction here. The flow gives the agent a clear mental model to follow, and the source hierarchy table does a lot of heavy lifting. 

I've left some minor review comments and I think we should be able to land this with just a little more work. Great stuff, @federicobartoli!

Sista21 (approved):

skills/source-driven-development/SKILL.md
federicobartoli: Great point , reframed around version-dependent correctness. Much cleaner.

skills/source-driven-development/SKILL.md
federicobartoli: Added a note after the source hierarchy: surface the discrepancy to the user and verify against the detected version.

skills/source-driven-development/SKILL.md
federicobartoli: Updated examples with anchors in the two links and added a citation rule to prefer deep links where possible.

federicobartoli:
> Really like the direction here. The flow gives the agent a clear mental model to follow, and the source hierarchy table does a lot of heavy lifting.
> 
> I've left some minor review comments and I think we should be able to land this with just a little more work. Great stuff, @federicobartoli!

Thanks for the review Addy, glad you liked the direction and the mental model approach! 🙏 ❤️ 
Addressed all three points:
1. Reframed "When NOT to use" around version-dependent correctness
2. Added guidance for conflicting official sources , surface the discrepancy, verify against the detected version
3. Updated citation examples with anchors and added a rule to prefer deep links where possible

Let me know if it's ready to land!

addyosmani (approved):

addyosmani:
Thank you! This looks good to land!

PR #51, opened 2026-04-09
feat: introduce evaluation framework for skill testing

Ic3b3rg:
## Summary

Introduce an evaluation framework for testing agent skills effectiveness. The repo had no way to measure whether skills actually improve agent performance - this addresses that gap.

Follows the bash + `claude -p` + prompt-fixtures pattern from [superpowers/tests](https://github.com/obra/superpowers/tree/main/tests).

## Context

This repo ships 21 skills but has no way to answer: *do they actually help?*

[SkillsBench](https://arxiv.org/abs/2602.12670) showed curated skills improve agent performance by +16.2pp on average - but also that poorly written ones hurt by -1.3pp. Without evaluation, we can't tell which category ours fall into.

This PR adds a lightweight, bash-based evaluation framework that measures skill effectiveness through scenario-based testing with baseline comparison.

## What this brings to the repo

Today, a contributor can edit a skill, pass `claude plugin validate`, and merge with no signal on whether the change helped or hurt agent behavior. This PR closes that gap:

- **Before/after comparison**: run any scenario with and without the skill, get a delta report
- **Regression safety net**: catch skill edits that degrade agent behavior before they land
- **Contributor testing path**: `CONTRIBUTING.md` now documents how to validate changes locally
- **Multi-model confidence**: same scenario across Haiku, Sonnet, and Opus via `EVAL_MODEL`
- **Structural guardrails**: automated checks for frontmatter, required sections, and description length, catches issues `claude plugin validate` doesn't cover

## What's included

- `test/validate-skills.sh` - structural validator (frontmatter, sections, description length)
- `test/scenarios/` - 6 scenarios for TDD, debugging, spec-driven-development
- `test/run-test.sh` / `test/run-all.sh` - single + batch runner, supports `--triggering`, `--baseline`, `EVAL_MODEL`
- `test/test-helpers.sh` - shared assertions and Claude execution wrappers
- `test/fixtures/` - minimal Node.js apps (zero deps, `node:test` only)
- `CONTRIBUTING.md` - added "Testing Your Changes" section

## Design decisions

- **Neutral prompts** — skill must be the sole source of process guidance for unbiased measurement
- **Bash scripts** — consistent with existing hooks (`simplify-ignore.sh`, `session-start.sh`)
- **`--plugin-dir` for with-skill runs** — loads only agent-skills without interference from other plugins
- **`--disallowedTools "Skill"` for baseline** — prevents agent from loading skills even if available
- **`EVAL_MODEL` env var** — multi-model testing (`claude -p --model`)
- **Prompts as `.txt` files** — aligned with superpowers `tests/skill-triggering/prompts/` pattern

## Test plan

- [ ] `claude plugin validate .` - existing CI unaffected by test/ additions
- [ ] `bash test/validate-skills.sh` - confirm 225 PASS, 0 FAIL, 3 WARN
- [ ] `bash test/run-all.sh --list` - shows all 6 scenarios with name and skill
- [ ] `bash test/run-all.sh --validate` - each scenario has prompt.txt, fixture/, and process-checks.sh
- [ ] `bash test/run-test.sh tdd-new-feature.json --triggering` - verify skill triggers from neutral prompt
- [ ] `bash test/run-test.sh tdd-new-feature.json` - grading output includes pass/fail per check
- [ ] `bash test/run-test.sh tdd-new-feature.json --baseline` - delta report shows skill vs no-skill difference
- [ ] Break a scenario on purpose (delete prompt.txt) → `--validate` catches it with actionable error

## Known warnings (3)

`bash test/validate-skills.sh` reports 3 expected warnings — these are documented design choices, not bugs:

| Warning | Skill | Reason |
|---------|-------|--------|
| description >250 chars (251) | `api-and-interface-design` | 1 character over Anthropic's truncation threshold. Minimal impact. |
| description >250 chars (268) | `browser-testing-with-devtools` | 18 characters over. Could be trimmed in a follow-up PR. |
| section missing: When to Use | `idea-refine` | Has `## Usage` instead, which describes invocation syntax rather than triggering conditions. Likely an omission — not a deliberate choice. |

Two additional skills are intentionally excluded from section checks:
- **`using-agent-skills`**: Meta-skill (router/discovery map), not a workflow skill - section checks skipped entirely
- **`idea-refine`**: Common Rationalizations check skipped - uses "Anti-patterns to Avoid" (bullet list) instead of the table format. @addyosmani chose this format in commit `ed34942`.

## What's next

- **More scenarios from real usage** — current 6 scenarios cover 3 skills. Additional scenarios should come from real contributor use cases, not invented ones.
- **Multi-model benchmark results** — `EVAL_MODEL` support is implemented but no benchmark data exists yet. Running across Haiku, Sonnet, and Opus would document how skill effectiveness varies by model.
- **More prompts per skill** — currently 1 prompt per scenario. Anthropic recommends at least 3 scenarios per skills.
- **Scenarios for remaining 18 skills** — framework supports any skill. Pattern: add a `.json` in `scenarios/`, a `.txt` in `prompts/`, and optionally a fixture.
- **Multi-turn conversation tests** — Current scenarios test single-prompt interactions. [Superpowers](https://github.com/obra/superpowers/tree/main/tests/explicit-skill-requests) includes multi-turn tests (`run-multiturn-test.sh`) to verify skill behavior across conversation turns — e.g., does the agent maintain TDD discipline when the user adds follow-up requests?

## What I'd like feedback on

- Scenario coverage - are TDD, debugging, and spec-driven-development the right starting three?
- Process checks granularity - currently binary pass/fail per check, considering weighted scoring
- Fixture complexity - kept minimal on purpose, open to expanding if too trivial

<details>
<summary>Sources that motivated this work</summary>

| Source | Key insight |
|--------|------------|
| [Anthropic Best Practices](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices) | *"Create evaluations BEFORE writing extensive documentation"* |
| [Anthropic Demystifying Evals](https://www.anthropic.com/engineering/demystifying-evals-for-ai-agents) | pass@k vs pass^k framework, *"grade what the agent produced, not the path it took"* |
| [agentskills.io](https://agentskills.io/skill-creation/evaluating-skills) | Evaluation pipeline with scenarios, verification scripts, baseline comparison |
| [SkillsBench](https://arxiv.org/abs/2602.12670) | Curated skills improve performance by +16.2pp across 7,308 trajectories |
| [Skills in the Wild](https://arxiv.org/abs/2604.04323) | Performance degrades as evaluation settings become more realistic |
| [Superpowers tests](https://github.com/obra/superpowers/tree/main/tests) | Bash + `claude -p` + stream-json, prompt fixtures, shared helpers |

</details>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,6 +36,37 @@ Every skill should include these sections:
 - Don't create supporting files unless content exceeds 100 lines
 - Don't put reference material inside skill directories — use `references/` instead
 
+## Testing Your Changes
+
+Before submitting a PR, validate your skill:
+
+```bash
+# Structural validation — checks format, sections, description
+bash test/validate-skills.sh
+```
+
+If you modified a skill's workflow, run the relevant evaluation scenario:
+
+```bash
+# List available scenarios
+bash test/run-all.sh --list
+
+# Fast check: does the skill activate on a neutral prompt?
+bash test/run-test.sh <scenario>.json --triggering
+
+# Full test: run agent with skill and grade the result
+bash test/run-test.sh <scenario>.json
+
+# Compare with-skill vs without-skill (measures the skill's value)
+bash test/run-test.sh <scenario>.json --baseline
+
+# Grade an existing workspace without re-running Claude
+bash test/run-test.sh <scenario>.json --grade-only
+
+# Test with a specific model (Anthropic recommends Haiku, Sonnet, and Opus)
+EVAL_MODEL=haiku bash test/run-test.sh <scenario>.json
+```
+
 ## Modifying Existing Skills
 
 - Keep changes focused and minimal
--- a/test/README.md
+++ b/test/README.md
@@ -1,0 +1,121 @@
+# Evaluation Framework
+
+Measures whether agent-skills actually improve agent performance. Aligned with [superpowers](https://github.com/obra/superpowers/tree/main/tests) testing patterns, [Anthropic's evaluation methodology](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices), and [SkillsBench](https://arxiv.org/abs/2602.12670).
+
+## Components
+
+### 1. Structural Validator
+
+Validates SKILL.md files against repo conventions that `claude plugin validate` does not check.
+
+```bash
+bash test/validate-skills.sh          # warnings don't fail
+bash test/validate-skills.sh --strict  # warnings = failures
+```
+
+Checks: description <250 chars, "Use when" trigger conditions, required sections (When to Use, Common Rationalizations, Red Flags, Verification), table format.
+
+### 2. Scenario-Based Evaluation
+
+Test scenarios for skills, using `claude -p` in headless mode. Two tiers:
+
+**Fast — Skill triggering (~2 min):** Does a neutral prompt activate the right skill?
+```bash
+bash test/run-test.sh tdd-new-feature.json --triggering
+```
+
+**Full — Integration + grading:** Does the agent produce correct results with the skill?
+```bash
+bash test/run-test.sh tdd-new-feature.json
+```
+
+**Full + baseline delta:** Compare with-skill vs without-skill.
+```bash
+bash test/run-test.sh tdd-new-feature.json --baseline
+```
+
+**Grade only:** Grade an existing workspace without re-running Claude.
+```bash
+bash test/run-test.sh tdd-new-feature.json --grade-only
+```
+
+**Model-specific testing:** Anthropic recommends testing skills with all models.
+```bash
+EVAL_MODEL=haiku bash test/run-test.sh tdd-new-feature.json    # enough guidance?
+EVAL_MODEL=sonnet bash test/run-test.sh tdd-new-feature.json   # clear and efficient?
+EVAL_MODEL=opus bash test/run-test.sh tdd-new-feature.json     # avoids over-explaining?
+```
+Source: "Test your Skill with all the models you plan to use it with." — platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices
+
+**Run all scenarios:**
+```bash
+bash test/run-all.sh                  # all with skill
+bash test/run-all.sh --triggering     # fast skill activation check
+bash test/run-all.sh --baseline       # all with delta
+bash test/run-all.sh --list           # list scenarios
+bash test/run-all.sh --validate       # validate JSON + prompt files
+```
+
+### 3. Process Graders
+
+Deterministic checks on the workspace after the agent finishes:
+
+```bash
+bash test/graders/process-checks.sh <workspace> <scenario.json>
+```
+
+Check types: `file_order`, `test_result`, `file_contains`, `file_exists`, `no_code_changes`.
+
+## Directory Structure
+
+```
+test/
+├── validate-skills.sh      # Repo convention validator
+├── test-helpers.sh          # Shared assertions + Claude execution helpers
+├── run-test.sh              # Run single scenario
+├── run-all.sh               # Run all scenarios
+├── scenarios/               # Scenario definitions (JSON metadata + checks)
+├── prompts/                 # One .txt file per scenario (neutral task prompt)
+├── fixtures/                # Mini-apps for agent to work on
+│   ├── feature-app/         # Clean app for TDD/spec scenarios
+│   └── buggy-app/           # App with known bugs for debugging scenarios
+└── graders/
+    └── process-checks.sh    # Deterministic grading
+```
+
+## Prompt Design
+
+Prompts in `prompts/*.txt` are **intentionally neutral** — they describe only the task, never the process. The skill is the sole source of process guidance.
+
+Sources:
+- Anthropic skill-creator: "Same prompt, no skill path" for baseline
+- SkillsBench: "Instructions must not reference which Skills to use"
+- agentskills.io: Baseline and with-skill runs use identical prompts
+
+## CLI Flags
+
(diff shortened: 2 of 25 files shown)


federicobartoli:
Thanks for putting this together! The instinct to ground skill quality in something measurable is absolutely the right one.
I’d like to push back gently on the overall approach, though. Before diving into the implementation details, I think we have a structural issue with how the scenarios are evaluated.

TL;DR: The current evaluation framework is prone to false positives because it checks static patterns in the final state, many of which are already satisfied by the initial fixtures.

To test the evaluation logic, I ran the graders against the untouched fixtures,simulating an agent that does absolutely nothing.
The results showed that the eval framework can't distinguish an agent that fixed a bug from an agent that ignored it:
debugging-null-ref-01: A no-op agent gets 3/3 PASS on the intact buggy fixture. (e.g., node --test output already contains the function name, and the "missing profile" string is already in the fixture).
tdd-new-feature-01: A no-op agent gets 2/4 PASS on the intact fixture. (The file order check trivially passes because git history on fixtures is identical, and node --test passes because pre-existing tests are green).

Given the context of the repo, I’m not sure a complex bash-based eval framework (with custom validators, AST-like parsing, and n=1 stochastic runs) justifies its maintenance burden, especially if the checks aren't grading on deltas (before vs. after).
Additionally, the validate-skills.sh script (180+ lines) feels a bit heavy for what it does. skill-anatomy.md already guides contributors well, and the existing 21 skills are already conformant. A 10-line grep script might do the same job.

Getting evals right for non-deterministic agents is genuinely hard. I wonder if it makes sense to pause this PR and open an issue to discuss the underlying question: What does "this skill works" actually mean for documentation-only skills, and what’s the smallest, most reliable way to measure it?
I suspect a reliable grader would need to snapshot initial and final states and assert strictly on the difference, rather than static string matching.
 have a few smaller notes on the script itself (e.g., stderr corrupting JSON logs, --baseline variance), but I think the methodological question is the one we should align on first. Let me know what you think!

Ic3b3rg:
Thanks @federicobartoli, useful comment. I think it's worth separating two things here:

1. **The quality of some specific graders**
2. **Whether introducing an eval harness is the right direction**

## On the first point: you're right.

Some checks were too permissive. I verified — a no-op on the fixture really does get 3/3 on `debugging-null-ref` and 2/4 on `tdd-new-feature`. Not defending that.

I've already pushed fixes:

- Assertions tightened on 3 scenarios with behavioral checks (`node -e`) that verify actual fix correctness, not just pattern presence
- Added `--null-test` mode — formalizes exactly your experiment. 6/6 scenarios now correctly reject the untouched fixture
- `validate-skills.sh` slimmed from 188 → 48 lines (removed everything `claude plugin validate` already covers in CI)
- stderr fix (`2>&1` → `2>/dev/null`)

## On the second point: I disagree with pausing.

The grader bugs are real, but I don't think they invalidate the framework. Every source I've looked at treats this as expected iteration, not a red flag:

- **agentskills.io** explicitly says: *"Review the assertions themselves... Remove or replace assertions that always pass in both configurations."* Your no-op test was step 3 of that cycle; the fix is step 4.
- **Anthropic best practices**: *"Create evaluations BEFORE writing extensive documentation... Start with 2-3 test cases."* — start small, iterate.
- **Anthropic's eval paper** describes the same dynamic in agent eval context — grader bugs get discovered and fixed, they don't invalidate the benchmark. It also warns that *"the compounding value of evals is invisible upfront while costs manifest immediately, leading teams to underinvest early."*
- **SkillsBench** (7,308 trajectories): curated skills improve pass rate by +16.2pp on average, but 16/84 tasks show negative deltas. Without evals you literally can't tell which category yours fall into.
- **"Skills in the Wild"** shows skill benefits degrade in realistic settings. The conclusion isn't "don't measure" — it's "measure more realistically."

### How I read this PR

Not as a framework that scientifically proves skill effectiveness, but as **minimal infrastructure to start measuring repeatably** isolated fixtures, neutral prompts, baseline comparison, deterministic graders. The kind of thing that gets better by iterating, not by waiting.

## What should come next

- Real delta checks over static patterns (partly done with `--null-test` and `node -e`)
- Distinguish smoke/regression checks from statistical claims
- Multi-trial where we want to talk about actual improvement
- Hybrid grading for qualitative skills (mechanical checks + human review)
- Multi-trial per scenario to reduce --baseline variance — n=1 is a smoke check, not a statistical claim

Happy to discuss any specific grader design. But I think the right move is to tighten the checks and keep going, not pause.

Ic3b3rg:
I have also added `files_modified` check: compares against the initial fixture commit with `git diff`, so a no-op agent always fails regardless of pattern matching. Also prints the actual diff to show what the agent changed. This should address your point about snapshotting initial vs final state :)

federicobartoli:
Don’t get me wrong: I’m not saying a proper evaluation framework isn’t needed. What I’m questioning is the architecture itself .I’m not fully convinced by it. You mentioned Superpowers, but if you look at the tests, they follow a different conceptual model in that repo. I also ran a few experiments locally and I’m still hitting some errors. I can keep investigating this further, but I’d prefer to keep the comment concise for now

ravibharathivs:
Hello, Federico Bartoli Thank you for your valuable feedback. I understand
your concerns regarding the architecture of the evaluation framework. I am
open to discussing this further to ensure that the design is robust and
aligns well with the overall goals of the project. Please let me know your
suggestions or if you would like to schedule a call to dive deeper into the
architectural considerations. best regards, Ravibharathi VS

addyosmani:
The eval framework has since landed (evals/, three tiers, running in CI). Closing as superseded, but thanks for the early push on this, it was the right idea.

PR #52, opened 2026-04-10
feat: add delegation-provenance skill

asiridalugoda:
## What this adds

This adds a new **delegation-provenance** skill for agent workflows that cross trust boundaries.

The simple idea is: once work starts bouncing between agents and tools, human approval gets fuzzy fast. It ends up living in chat history, UI state, or in one orchestrator that "knows" what was allowed. This skill pushes back on that. It teaches agents to treat approval as something they need to verify, carry forward, and refresh when the scope changes.

## The problem it solves

Agent systems are getting very good at delegation, but the accountability story still gets messy once work fans out.

After a few handoffs, it becomes surprisingly hard to answer basic questions:

- who actually authorized the final action
- what scope was granted
- whether the current step is still inside that scope
- whether the workflow quietly picked up new powers halfway through
- whether the downstream tool or agent had any way to check this before acting

That gap matters most when the workflow touches sensitive data, persistent state, money, or anything with real-world consequences.

## How it works

Four-step process: **CLASSIFY → VERIFY → PROPAGATE → REAUTHORIZE**

1. Classify the trust boundary before each agent handoff or tool call
2. Verify a durable authorization artifact with principal, scope, expiry, and delegation limits
3. Propagate provenance on every hop with an append-only delegation record
4. Require fresh human approval whenever scope expands

The skill also makes two things explicit:

- fail closed when authorization is missing, expired, ambiguous, or out of scope
- surface mismatches between declared authorization and runtime permissions instead of hand-waving them away

## What makes it different from generic security guidance

This is not just "be careful with auth."

It focuses on **multi-hop authorization continuity**:
- approval should survive delegation
- downstream actors should be able to inspect scope before acting
- scope expansion should be visible, not something that just sort of happens

It also stays implementation-agnostic, so teams can apply it across MCP tools, A2A-style agent handoffs, hosted services, or internal multi-agent systems.

## Integration

- Added to the `using-agent-skills` discovery flow and lifecycle sequence
- Added to `CLAUDE.md` Build phase
- Added to `README.md` Build table and project structure (`21` skills total)

## Background

This skill is informed by the Human Delegation Provenance (HDP) work:

- Paper: https://arxiv.org/abs/2604.04522
- IETF Internet-Draft: https://datatracker.ietf.org/doc/draft-helixar-hdp-agentic-delegation/

HDP frames human authorization as a verifiable chain-of-custody for multi-agent systems. This skill borrows that design insight, but keeps the workflow broad enough to be useful even if a team is not using HDP directly.

## Demo

Example workflow with the skill active:

- `human-support-manager -> triage-agent`
- `triage-agent -> policy-agent`
- `policy-agent -> refund-tool`

The agent:
- classifies the refund path as a financial trust boundary
- verifies allowed actions, delegatees, expiry, and delegation depth
- carries the authorization context forward on each hop
- asks for fresh approval before adding a new delegatee (`comms-agent`) to send a customer-facing email

Without the skill, that email step is easy to smuggle in as "basically the same task." With the skill, it gets surfaced for what it is: a scope change.

## Future work

A natural follow-up would be implementation-specific examples for ADK callbacks, MCP middleware, or A2A boundaries.

I kept this PR focused on the workflow itself so the skill stays generally useful and easier to review.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,7 +17,7 @@ docs/         → Setup guides for different tools
 
 **Define:** spec-driven-development
 **Plan:** planning-and-task-breakdown
-**Build:** incremental-implementation, test-driven-development, context-engineering, source-driven-development, frontend-ui-engineering, api-and-interface-design
+**Build:** incremental-implementation, test-driven-development, context-engineering, source-driven-development, delegation-provenance, frontend-ui-engineering, api-and-interface-design
 **Verify:** browser-testing-with-devtools, debugging-and-error-recovery
 **Review:** code-review-and-quality, code-simplification, security-and-hardening, performance-optimization
 **Ship:** git-workflow-and-versioning, ci-cd-and-automation, deprecation-and-migration, documentation-and-adrs, shipping-and-launch
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ The commands above are the entry points. Under the hood, they activate these 21 
 | [test-driven-development](skills/test-driven-development/SKILL.md) | Red-Green-Refactor, test pyramid (80/15/5), test sizes, DAMP over DRY, Beyonce Rule, browser testing | Implementing logic, fixing bugs, or changing behavior |
 | [context-engineering](skills/context-engineering/SKILL.md) | Feed agents the right information at the right time - rules files, context packing, MCP integrations | Starting a session, switching tasks, or when output quality drops |
 | [source-driven-development](skills/source-driven-development/SKILL.md) | Ground every framework decision in official documentation - verify, cite sources, flag what's unverified | You want authoritative, source-cited code for any framework or library |
-| [delegation-provenance](skills/delegation-provenance/SKILL.md) | Enforce scoped authorization, auditable handoffs, and reauthorization across agent and tool boundaries | Building agents that delegate, call tools, or act across trust boundaries |
+| [delegation-provenance](skills/delegation-provenance/SKILL.md) | Preserve scoped authorization, auditable handoffs, and reauthorization across delegated execution | Building multi-agent or delegated tool workflows where authorization must survive multiple hops |
 | [frontend-ui-engineering](skills/frontend-ui-engineering/SKILL.md) | Component architecture, design systems, state management, responsive design, WCAG 2.1 AA accessibility | Building or modifying user-facing interfaces |
 | [api-and-interface-design](skills/api-and-interface-design/SKILL.md) | Contract-first design, Hyrum's Law, One-Version Rule, error semantics, boundary validation | Designing APIs, module boundaries, or public interfaces |
 
--- a/README.md
+++ b/README.md
@@ -117,9 +117,9 @@ Skills are plain Markdown - they work with any agent that accepts system prompts
 
 ---
 
-## All 20 Skills
+## All 21 Skills
 
-The commands above are the entry points. Under the hood, they activate these 20 skills — each one a structured workflow with steps, verification gates, and anti-rationalization tables. You can also reference any skill directly.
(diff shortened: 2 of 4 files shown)


federicobartoli:
Thanks for putting this together!The four-step framing is clear and feels like a solid foundation.
I do want to flag a concern about the worked examples. As written, they may lead an agent applying this skill to produce code that resembles authorization without necessarily enforcing it in a robust way.

In particular, the JSON receipt and authorize() function don’t seem to include a notion of issuer signature or verification. The receipt is a plain object, so an agent in the chain could construct one with something like allowed_actions: ["*"], and the verification function would accept it. There isn’t an obvious trust anchor, signature check, or separation between issuer and consumer.
Given that LLMs often mirror structural examples quite closely, I’d expect an agent following this pattern to end up implementing something that looks like a delegation token, but without the properties that make it secure.

I wonder if this could be strengthened by anchoring the examples to an existing standard : for example RFC 8693 (OAuth 2.0 Token Exchange), signed JWTs, UCAN, or Macaroons. Each of these addresses parts of this problem space and brings well-understood cryptographic guarantees. That might help avoid the “looks right but isn’t safe” middle ground the current example could fall into.

asiridalugoda:
Thanks @federicobartoli, thats a really good catch.

You’re right, the earlier JSON + authorize() example could push an agent toward something that looks like a delegation token without any real trust anchor or verification behind it.

I’ve updated the PR to remove that example and reframe the section around verifying an existing authorization artifact instead of inventing one ad hoc. I also added explicit guidance around issuer verification, signatures/MACs, trust anchors, and pointed to existing standards/protocols like RFC 8693, signed JWTs, UCAN, Macaroons, and HDP.

Really appreciate you flagging it. That “looks right but isn’t actually safe” middle ground is exactly what I want to avoid here. 🥇

addyosmani:
The rewrite in response to Federico's feedback is strong. Thank you.

Main question I'd want @asiridalugoda to think about: is this the right repo for this skill? :) Most existing skills are about engineering workflow an AI coding agent follows when writing software - TDD, debugging, specs, reviews, git. This one is about a *property of the software the agent is building* - specifically, multi-agent or multi-tool systems with trust boundaries. 

That's a narrower activation surface than everything else in the Build phase. Someone building a React component or a REST endpoint will never hit it. I'm not saying that's disqualifying, but the description should make the narrow applicability loud so it doesn't misfire on ordinary work. Something like "Use ONLY when building systems where the software itself delegates across agents or tools" rather than the current opening, which an LLM could plausibly trigger on any mention of tools or permissions.

One concrete code-level thing: the `authorize()` example still hinges entirely on `verifyAuthorizationArtifact(artifact, trustAnchors)` - which is the one function that actually matters and is hand-waved. That's fine as a sketch but I'd add a sentence right under it saying "this skill does not implement verifyAuthorizationArtifact - use a vetted library for your chosen artifact format (jose for JWTs, etc), and treat rolling your own as a red flag."

asiridalugoda:
@addyosmani Thanks for the detailed feedback really appreciate it!

I’ll address the points today (tightening the “When to Use” section to make the narrow applicability much louder, plus adding the clear disclaimer under the authorize() example and using vetted libraries).

I believe this is the right repo for the skill as it improves how agents handle real-world authorization continuity in multi-agent/tool delegation scenarios. exactly the kind of edge that current workflows often miss and where subtle security gaps creep in. 

Happy to iterate until it fits cleanly with the rest of the skills.

Will push an updated PR shortly once I’m back in front of a proper PC. Thanks again!

asiridalugoda:
@addyosmani Just pushed an update with those changes. tightened the activation criteria and added the verifier note under `authorize()`. Thanks again for the feedback.

addyosmani:
Thanks for this. Tracking delegation provenance is niche enough that it doesn't fit the core lifecycle pack. Closing as out of scope.

PR #59, opened 2026-04-11
Add observability skill, cross-references, and persona integration

googlarz:
## Summary

- **New skill: `observability-and-monitoring`** — fills the only confirmed lifecycle gap in the collection. Covers structured logging (with cardinality rules), four golden signals metrics, OpenTelemetry-standard distributed tracing (W3C `traceparent`/`tracestate`), SLO-based alerting (burn rate, not raw thresholds), and dashboard design. Follows `skill-anatomy.md` format exactly: Overview, When to Use, stepwise process, Common Rationalizations, Red Flags, Verification checklist.
- **Cross-references added to isolated skills** — `idea-refine` now points forward to `spec-driven-development`; `deprecation-and-migration` now references `incremental-implementation`, `test-driven-development`, and `git-workflow-and-versioning` at the natural touch points in the migration process.
- **`/ship` command updated** — `observability-and-monitoring` is wired into the launch checklist so the skill is reachable via the `/ship` entrypoint, not just via the meta-skill decision tree.
- **`using-agent-skills` updated** — new skill added to decision tree, lifecycle sequence (before `shipping-and-launch`, not after), and quick reference table.
- **README updated** — skill count 20 → 21, new entry in Ship section and directory tree.

## How findings were addressed

This PR went through both `/codex:adversarial-review` and `/codex:review` before submission:

| Finding | Fix |
|---|---|
| Metric `path` label causes cardinality explosion | Changed to `route_template`; added explicit cardinality rule banning user IDs, tenant IDs, query strings from labels |
| Tracing example used custom `X-Trace-Id` headers (non-interoperable) | Replaced with OpenTelemetry `inject()` + W3C `traceparent`/`tracestate` |
| New skill not reachable via `/ship` entrypoint | Added as step 6 in ship command |
| Skill count stale in README | Updated to 21, added to table and directory tree |
| Lifecycle placed observability after deploy | Moved before `shipping-and-launch` in decision tree, lifecycle, and quick reference |

## Test plan

- [ ] Verify `skills/observability-and-monitoring/SKILL.md` exists and passes frontmatter lint
- [ ] Verify skill count in README matches `ls skills/ | wc -l`
- [ ] Confirm `/ship` command references `observability-and-monitoring`
- [ ] Confirm lifecycle in `using-agent-skills` has observability before shipping

### DIFF
--- a/.claude/commands/ship.md
+++ b/.claude/commands/ship.md
@@ -11,7 +11,8 @@ Run through the complete pre-launch checklist:
 3. **Performance** — Core Web Vitals good, no N+1 queries, images optimized, bundle sized
 4. **Accessibility** — Keyboard nav works, screen reader compatible, contrast adequate
 5. **Infrastructure** — Env vars set, migrations ready, monitoring configured
-6. **Documentation** — README current, ADRs written, changelog updated
+6. **Observability** — Invoke the agent-skills:observability-and-monitoring skill to verify structured logging, metrics (four golden signals), SLOs, and alerting are in place before going live
+7. **Documentation** — README current, ADRs written, changelog updated
 
 Report any failing checks and help resolve them before deployment.
 Define the rollback plan before proceeding.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,10 +24,32 @@ The agent should automatically map user intent to skills:
 - Feature / new functionality → `spec-driven-development`, then `incremental-implementation`, `test-driven-development`
 - Planning / breakdown → `planning-and-task-breakdown`
 - Bug / failure / unexpected behavior → `debugging-and-error-recovery`
-- Code review → `code-review-and-quality`
+- Code review → `code-review-and-quality` + load `agents/code-reviewer.md` persona for the review subagent
+- Security review / hardening → `security-and-hardening` + load `agents/security-auditor.md` persona
+- Test strategy / writing tests → `test-driven-development` + load `agents/test-engineer.md` persona for bug reproduction subagents
 - Refactoring / simplification → `code-simplification`
 - API or interface design → `api-and-interface-design`
 - UI work → `frontend-ui-engineering`
+- Production visibility / logging / alerting → `observability-and-monitoring`
+
+### Agent Personas
+
+Three reusable agent personas live in `agents/`. Load them when a skill calls for a specialized review or testing subagent — they provide structured output formats, severity classifications, and rules the base agent lacks.
+
+| Persona | File | Load when |
+|---------|------|-----------|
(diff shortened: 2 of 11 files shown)


nucliweb:
Thanks @googlarz for the thorough work here — the adversarial review pass before submitting is exactly the right discipline, and it shows. The technical decisions in the observability skill are solid: route_template cardinality rule, OTel + W3C traceparent, burn-rate alerting over raw thresholds, and the "When NOT to use" section are all production-grade choices that less experienced contributors miss.

A few things worth discussing before merge:

### 1. Two independent concerns in one PR

**The observability skill and the agent persona integration are unrelated changes**. The persona wiring touches code-review-and-quality, security-and-hardening, test-driven-development, AGENTS.md, and CLAUDE.md — none of which have anything to do with observability. Bundling them increases blast radius and makes the diff harder to review. If either half needs to be reverted, you're reverting both. These should be two PRs.

### 2. The decision tree trigger is reactive

```
├── Need production visibility? ──→ observability-and-monitoring
```

The skill itself says "build observability before you need it, not after your first outage", but this trigger only fires when you already feel the pain. Something like "Instrumenting a service / setting up monitoring?" would match the proactive framing the skill is actually teaching.

### 3. Lifecycle placement is right, but incomplete

Putting observability before shipping-and-launch (step 11 of 12) is correct. The problem is it implies you start thinking about it at the end. A small note in incremental-implementation, something like "when defining a slice, define what you'll measure" — would anchor observability to where it actually starts, without moving the full skill.

### 4. Missing references/observability-checklist.md

Every other skills covering a technical domain has a corresponding checklist in references/:

- references/security-checklist.md
- references/testing-patterns.md
- references/performance-checklist.md

This skill doesn't create one and doesn't reference one. A checklist covering "do I have the four golden signals? trace_id on every log line? SLOs defined?" would be the most immediately useful artifact for someone running this skill on a real service — and it would make this PR consistent with the rest of the repo.

### 5. No Step 0 for existing services

The "When to Use" mentions "taking ownership of an existing one" but the steps assume you're starting from scratch. There's no audit step, "what instrumentation do you already have?", before adding more. That's the most common real-world case.


> None of these are blockers on the core skill, which is good work. The main structural concern is point 1 — the persona integration deserves its own PR so both changes can be reviewed and reverted independently.

googlarz:
Thanks for the thorough review — all points taken.

Closing this in favour of two focused PRs:

- **#61** — observability skill only, with all four content fixes applied (proactive trigger wording, brownfield Step 0, `references/observability-checklist.md`, `incremental-implementation` anchor)
- **#62** — persona integration only (zero observability changes)

The CI validator is in its own separate **#60**.

PR #60, opened 2026-04-11
Add CI skill validator

googlarz:
## Summary

- **`scripts/validate-skills.js`** — zero-dependency Node.js script that validates every skill on every PR. Checks: SKILL.md exists, frontmatter has `name` + `description`, `name` matches directory, description ≤ 1024 chars, all five required sections present (Overview, When to Use, Common Rationalizations, Red Flags, Verification), and cross-skill references point to real skills.
- **CI job `validate-skills`** — runs before the existing manifest-validation job and blocks merge on any error. Uses Node 20, no `npm install` required.
- **Validator-owned allowlist** — `using-agent-skills` and `idea-refine` are exempt from section checks via an explicit `SECTION_EXEMPT_SKILLS` map in the script (with documented reasons). Skills cannot self-declare exemptions via frontmatter — if they try, CI fails loud.

## Adversarial review finding addressed

Initial implementation used `type: meta` / `exempt: sections` frontmatter fields to opt skills out of section checks. Adversarial review correctly flagged this as a bypass vector — any contributor could add `exempt: sections` to their own skill and pass CI. Fixed by moving the allowlist entirely into the validator script. Non-allowlisted skills that declare these fields now fail with an explicit error.

## Test plan

- [ ] `node scripts/validate-skills.js` exits 0 on current main
- [ ] Introducing a skill with missing frontmatter exits 1
- [ ] Introducing a skill with missing `## Verification` section exits 1  
- [ ] Introducing a skill with `exempt: sections` that is not in the allowlist exits 1
- [ ] CI `validate-skills` job passes on this branch

### DIFF
--- a/.github/workflows/test-plugin-install.yml
+++ b/.github/workflows/test-plugin-install.yml
@@ -6,8 +6,23 @@ on:
   workflow_dispatch:
 
 jobs:
+  validate-skills:
+    name: Validate skill content
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Validate all skills
+        run: node scripts/validate-skills.js
+
   validate:
     name: Validate plugin structure
+    needs: validate-skills
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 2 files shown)


googlarz:
Hi @addyosmani — gentle ping on this one. Happy to rebase or address any feedback whenever you get a chance.

addyosmani:
Thanks for your patience. Nice work.

This is probably what the validator needed to be with some real checks blocking in CI. Keeping the allowlist centralized in the validator instead of the frontmatter is super smart and I think it closes an easy loophole and forces any exemptions to go through a proper PR review. Merging this now and closing #153 (which tried to do the same thing but referenced a missing script). Great work, @googlarz!

PR #68, opened 2026-04-13
docs(cursor): update setup guide to use native Agent Skills

ArthurMor4is:
## What

Cursor 3.x+ (stable) ships native Agent Skills support — skills are placed in `~/.cursor/skills/` (personal) or `.cursor/skills/` (project) and loaded dynamically by the agent when relevant. The previous workaround of copying `SKILL.md` files into `.cursor/rules/` or inlining them into `.cursorrules` is no longer needed.

## Changes

- **README.md**: replaced the one-liner Cursor section with clear personal/project install instructions
- **docs/cursor-setup.md**: full rewrite — explains the native skills directory, dynamic loading vs always-on rules, one-liner install, and a migration guide from `.cursor/rules/` using Cursor's built-in `/migrate-to-skills` command

## Why it matters

The old setup (rules directory) loads all skill content into every conversation — consuming context window regardless of relevance. Native skills load descriptions at startup and pull the full `SKILL.md` only when the agent decides it's needed, which is exactly the behavior this project's skills are designed for.

Verified on Cursor 3.0.12 (stable, macOS arm64).

PR #69, opened 2026-04-13
Fix: Remove deprecated Option 3 (Notepads) from cursor setup documentation

sbusanelli:
Fixes #44

## Changes
- Remove deprecated Option 3 (Notepads) section from cursor setup documentation
- Update Phase-Specific Skills section to use on-demand rule files instead of notepads
- Update Usage Tips to remove all references to deprecated notepad functionality
- Maintain existing functionality with Rules Directory and .cursorrules options

## Impact
This fix removes outdated documentation that references deprecated Cursor functionality, ensuring users only see current, working setup options.

### DIFF
--- a/docs/cursor-setup.md
+++ b/docs/cursor-setup.md
@@ -29,15 +29,6 @@ echo "\n---\n" >> .cursorrules
 cat /path/to/agent-skills/skills/code-review-and-quality/SKILL.md >> .cursorrules
 ```
 
-### Option 3: Notepads
-
-Cursor's Notepads feature lets you store reusable context. Create a notepad for each skill you use frequently:
-
-1. Open Cursor → Settings → Notepads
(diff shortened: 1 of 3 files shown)


rachelslurs:
Hi there! I think Cursor offers support for `~/.cursor/skills` for this use case instead actually. Might be worth looking into instead of asking people to add and remove to save tokens.

addyosmani:
Closing as superseded — the same Notepads-removal change is included in #70 as the first commit, and #70 is the one moving forward to merge. Thanks for the contribution!

Also flagging @rachelslurs's good point above for anyone landing here later: Cursor 3.x+ supports `~/.cursor/skills/` natively, so the longer-term fix for cursor-setup.md is to switch from "rules files you add and remove" to native skill loading. That's tracked in #68.

PR #70, opened 2026-04-13
Fix: Add guidance to prevent redundant build verification commands

### LINKED ISSUE #47: Redundant build verification commands during scaffolding tasks
## Observed behavior

When using Claude Code with the agent-skills plugin to scaffold projects, the agent sometimes runs build verification commands (e.g. `npx tsc --noEmit`, `npx vite build`) multiple times in a row after file modifications, even when the first run already succeeded with no errors.

```
⏺ Bash(npx tsc --noEmit 2>&1)
  ⎿  (No output)

⏺ Bash(npx tsc --noEmit 2>&1)
  ⎿  (No output)

⏺ Bash(npx tsc --noEmit 2>&1)
  ⎿  (No output)
```

This is a Claude Code agent behavior, not something caused by the plugin. However, it could potentially be mitigated at the plugin level.

## Possible plugin-level mitigations

- A `PostToolUse` hook that detects and skips duplicate consecutive Bash commands
- Guidance in the `incremental-implementation` or `test-driven-development` skills (e.g. "verify once after changes, do not repeat on success")

## Has anyone else seen this?

If you've encountered this pattern, please share:
- Which model (Sonnet, Opus, etc.)
- What kind of task triggered it
- How often it happens

Would be useful to understand if this is widespread before investing in a fix.

sbusanelli:
Fixes #47

## Changes
- Add verification guidance to incremental-implementation skill checklist
- Add test guidance to test-driven-development skill verification section  
- Add common rationalizations about redundant verification in both skills
- Add red flags for running same command multiple times after success

## Impact
This addresses the issue where agents run build verification commands (npx tsc --noEmit, npx vite build, npm test) multiple times consecutively after successful runs, providing clear guidance to trust the first successful verification.

## Technical Details
- Added notes in verification checklists to run commands once per increment/cycle
- Added rationalizations table entries addressing the "run again just to be sure" mindset
- Added red flags to make redundant verification patterns more visible
- Maintains existing verification requirements while preventing unnecessary repetition

### DIFF
--- a/docs/cursor-setup.md
+++ b/docs/cursor-setup.md
@@ -29,15 +29,6 @@ echo "\n---\n" >> .cursorrules
 cat /path/to/agent-skills/skills/code-review-and-quality/SKILL.md >> .cursorrules
 ```
 
-### Option 3: Notepads
-
-Cursor's Notepads feature lets you store reusable context. Create a notepad for each skill you use frequently:
-
-1. Open Cursor → Settings → Notepads
-2. Create a new notepad named "swe: Test-Driven Development"
-3. Paste the content of `skills/test-driven-development/SKILL.md`
-4. Reference it in chat with `@notepad swe: Test-Driven Development`
-
 ## Recommended Configuration
 
 ### Essential Skills (Always Load)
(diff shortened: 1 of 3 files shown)


addyosmani (changes requested):
Thank you for the PR!

On the redundant-verification change: the instinct is right and this does address a real behavior, but the wording across the three additions is a little inconsistent.

The note on the verification checklist says "once per increment" which is correct - the right rule is "run once per change, then stop." But the rationalization table then says "running it again provides no additional value" and the red flag says "running the same command multiple times after a successful run" both of which are stronger claims that are only true if nothing has changed. 

Taken literally, an agent could read the rationalization and skip verification after a subsequent edit because it ran the command once before. I'd tighten all three to use the same framing - maybe something like "after a successful run, don't repeat the same command unless the code has changed since." Wdyt?

PHILADEL27-alt (commented):
Approve

PR #74, opened 2026-04-14
Add citation cache to source-driven-development

federicobartoli:
## Why

source-driven-development is about grounding every framework decision in official docs, not memory. But that means every session re-fetches the same React 19 form handler docs, the same Django auth patterns. It works, but it's repetitive.

This PR adds a simple idea: write down what you've verified, so the next session can look it up instead of re-verifying.

## How it works

When the agent implements a feature, it now:
1. Checks `.claude/sdd-citations.md` — have we already verified React 19 useActionState?
2. If yes (and the version matches), reuse the citation and skip the fetch
3. If no, fetch as usual, then add a row to the cache

The cache is a simple markdown table with columns: framework, version, pattern, url, verified_at, notes.

**Who sees it?** By default, gitignored and private per developer. But teams can opt in to commit it for a shared audit trail of what's been verified.

## Example

Your project uses React 19.1. First implementation of a form pattern:
- Fetch react.dev/reference/react/useActionState — 5 tokens, cache the result

Next session, same project:
- Check `.claude/sdd-citations.md` — find the row — reuse the citation, skip fetch
- Result: 0 tokens wasted on repeat verification

## Changes

- **Step 2 (Fetch)** now has a cache-check substep before any documentation fetch
- **Step 4 (Cite)** now records the citation to the cache
- **New section** "Citation Cache Protocol" with format, strict lookup rules, and what not to cache
- **Rationalization** for "re-fetching just to be safe" (don't bypass a cache you don't trust — delete the row and re-verify instead)
- **Red flags** for cache misuse (approximate version matching, caching non-authoritative sources, patching rows in place)
- **.gitignore** updated to ignore the cache by default

## Testing

Any implementation using source-driven-development will now:
1. Read `.claude/sdd-citations.md` before any fetch (creates it if missing)
2. Add a row after each successful fetch
3. Verify strict version matching (react@19.0 and react@19.1 are separate entries)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules/
 .env.*
 *.log
 .claude/.simplify-ignore-cache/
+.claude/sdd-citations.md
--- a/skills/source-driven-development/SKILL.md
+++ b/skills/source-driven-development/SKILL.md
@@ -27,12 +27,13 @@ Every framework-specific code decision must be backed by official documentation.
 ## The Process
 
 ```
-DETECT ──→ FETCH ──→ IMPLEMENT ──→ CITE
-  │          │           │            │
-  ▼          ▼           ▼            ▼
- What       Get the    Follow the   Show your
- stack?     relevant   documented   sources
-            docs       patterns
+DETECT ──→ RECALL ──→ FETCH ──→ IMPLEMENT ──→ CITE + CACHE
+  │          │          │           │              │
+  ▼          ▼          ▼           ▼              ▼
+ What       Cache      Get the     Follow the    Show your
+ stack?     hit?       relevant    documented    sources and
+            Skip       docs        patterns      append the
+            fetch                                citation
 ```
 
 ### Step 1: Detect Stack and Versions
@@ -60,9 +61,11 @@ STACK DETECTED:
 
 If versions are missing or ambiguous, **ask the user**. Don't guess — the version determines which patterns are correct.
 
-### Step 2: Fetch Official Documentation
+### Step 2: Check the Citation Cache, Then Fetch
 
-Fetch the specific documentation page for the feature you're implementing. Not the homepage, not the full docs — the relevant page.
+**Check the cache first.** Read `.claude/sdd-citations.md` if it exists. If it contains a row matching `framework + version + pattern` for the exact version detected in Step 1, reuse that citation and skip the fetch. Match is **strict**: `react@19.1` does not satisfy `react@19.0`. See the Citation Cache Protocol section below for the format and lookup rules.
+
+If no cache hit, fetch the specific documentation page for the feature you're implementing. Not the homepage, not the full docs — the relevant page.
 
 **Source hierarchy (in order of authority):**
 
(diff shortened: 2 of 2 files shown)


addyosmani (commented):
Hey @federicobartoli - thanks for putting this together! Still thinking on it, but some initial thoughts:

Repeated doc fetches across sessions is genuinely annoying, and I like that you're thinking about how to make the workflow more efficient.

**A few things I'm thinking about**

**Staleness risk** - I think the whole point of SDD is "don't trust memory, verify against current docs." A persistent cache is kind of a local memory of what docs said at some point, which (I think?) cuts against that philosophy a bit. Docs can change within the same version (corrections, restructured pages, new examples). Curious how you're thinking about that tradeoff?

**Complexity budget** - The skill is pretty lean right now (~195 lines, one clean workflow). Adding cache lookup, recording, invalidation rules, gitignore guidance, and team-sharing options is a fair amount of new surface area. I wonder if there's a lighter-weight version that captures most of the value?

**Cache invalidation** - The "delete, don't patch" rule is smart, but I'm thinking about edge cases. When a project upgrades from React 19.0 to 19.1, does the agent know to invalidate the 19.0 entries? Who cleans up stale rows over time?

**Alternative shape?** could this be solved with a shorter addition to Step 2 saying "if you already fetched this doc earlier in the current session, reuse it"? That handles the most common case (repeated fetches in one sitting) without needing a persistent file or invalidation logic. The heavier cross-session caching might fit better at the tool level (HTTP caching, `WebFetch` deduplication) rather than in the skill itself.

None of this is a hard no of course just thinking out loud. What's your take?

federicobartoli:
> Hey @federicobartoli - thanks for putting this together! Still thinking on it, but some initial thoughts:
> 
> Repeated doc fetches across sessions is genuinely annoying, and I like that you're thinking about how to make the workflow more efficient.
> 
> **A few things I'm thinking about**
> 
> **Staleness risk** - I think the whole point of SDD is "don't trust memory, verify against current docs." A persistent cache is kind of a local memory of what docs said at some point, which (I think?) cuts against that philosophy a bit. Docs can change within the same version (corrections, restructured pages, new examples). Curious how you're thinking about that tradeoff?
> 
> **Complexity budget** - The skill is pretty lean right now (~195 lines, one clean workflow). Adding cache lookup, recording, invalidation rules, gitignore guidance, and team-sharing options is a fair amount of new surface area. I wonder if there's a lighter-weight version that captures most of the value?
> 
> **Cache invalidation** - The "delete, don't patch" rule is smart, but I'm thinking about edge cases. When a project upgrades from React 19.0 to 19.1, does the agent know to invalidate the 19.0 entries? Who cleans up stale rows over time?
> 
> **Alternative shape?** could this be solved with a shorter addition to Step 2 saying "if you already fetched this doc earlier in the current session, reuse it"? That handles the most common case (repeated fetches in one sitting) without needing a persistent file or invalidation logic. The heavier cross-session caching might fit better at the tool level (HTTP caching, `WebFetch` deduplication) rather than in the skill itself.
> 
> None of this is a hard no of course just thinking out loud. What's your take?

Hey Addy, I was thinking about this PR again this morning while running (lol), and I’m not entirely sure anymore. Beyond all the valid points you raised, I’m a bit concerned the agent might hallucinate the cache. Maybe I should explore something lighter-weight, possibly tied to a hook so it stays deterministic.

In my tests it did work, but I was using Opus 4.6 with max thinking/effort, so not exactly a baseline scenario.

If you’re okay with it, we could pause this for now and I’ll think it through over the next few days. Otherwise I’m also happy to close it and reopen once things are clearer. What do you think?

Thanks again for the really insightful feedback 🙏

federicobartoli:
Hi @addyosmani  , I reworked this from the ground up based on your feedback.

Instead of introducing caching logic into the skill, the cache now lives entirely as a pair of opt-in Claude Code hooks on `WebFetch`. The skill itself remains unchanged.

On reuse, content is served from cache **only after the origin confirms `304 Not Modified` via conditional revalidation**, so SDD’s “verify against current docs” rule still holds on every access.

Opening a fresh PR ([#80](https://github.com/addyosmani/agent-skills/pull/80)) in draft since the approach is completely different, both in terms of architecture and trade-offs.

Curious if this direction feels more aligned with the original concerns, especially keeping caching at the tool layer rather than inside the skill.

PR #58, opened 2026-04-11
Fix button accessibility guidance in frontend-ui-engineering skill

yaphi1:
Native buttons support both "enter" and "space", and any JavaScript simulations of buttons should account for that when handling keyboard events.

This PR fixes the fact that "space" was missing in the frontend accessibility guidance for keyboard event handlers.

I hope this PR helps! 😄

PR #79, opened 2026-04-16
docs(frontend-ui): prevent page scroll in custom button example

### LINKED ISSUE #78: Custom button example scrolls the page on Space key
## Problem

PR #58 correctly added `Space` activation to the custom button example in `skills/frontend-ui-engineering/SKILL.md`, but the current snippet still allows the browser's default `Space` behavior on a non-native button.

Current example:

```tsx
<div role="button" tabIndex={0} onClick={handleClick}
     onKeyDown={e => (e.key === 'Enter' || e.key === ' ') && handleClick()}>
  Click me
</div>
```

In a local browser reproduction, pressing `Space` once on that exact pattern:
- fired the click handler once, **and**
- scrolled the page from `scrollY = 0` to `scrollY = 660`

When I added `e.preventDefault()` for `e.key === ' '`, the click still fired once and `scrollY` stayed at `0`.

## Why this matters

This repo positions `frontend-ui-engineering` as WCAG 2.1 AA guidance and presents the skill set as production-grade engineering workflows. For a custom element with `role="button"`, the example should model the full keyboard interaction, not a version that also scrolls the page.

## Proposed fix

Update the example to prevent the default `Space` action before calling `handleClick()`:

```tsx
onKeyDown={e => {
  if (e.key === ' ') e.preventDefault();
  if (e.key === 'Enter' || e.key === ' ') handleClick();
}}
```

## References

- PR #58: https://github.com/addyosmani/agent-skills/pull/58
- WAI APG Button Pattern: https://www.w3.org/WAI/ARIA/apg/patterns/button/
- MDN button role: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/button_role

shaun0927:
## Summary

Prevent page scroll in the custom button keyboard example by calling `e.preventDefault()` for `Space` before triggering `handleClick()`.

## Problem

PR #58 correctly added `Space` activation to the non-native button example, but the current snippet still allows the browser's default `Space` action on a custom `role="button"` element.

That means the example can both:
- activate the handler, and
- scroll the page

In a local browser reproduction of the current pattern:
- one `Space` press triggered the click handler once
- `window.scrollY` changed from `0` to `660`

With the updated snippet in this PR:
- one `Space` press still triggered the click handler once
- `window.scrollY` stayed at `0`

## Fix

Keep the change docs-only and limited to the existing example in `skills/frontend-ui-engineering/SKILL.md`.

The updated snippet:
- prevents the default `Space` action on the custom button
- preserves `Enter` and `Space` activation behavior
- stays aligned with the accessibility intent of the skill

## Verification

- Local WKWebView/browser reproduction of the current snippet
- Local WKWebView/browser reproduction of the patched snippet
- `git diff --check`

Closes #78

### DIFF
--- a/skills/frontend-ui-engineering/SKILL.md
+++ b/skills/frontend-ui-engineering/SKILL.md
@@ -174,8 +174,11 @@ Every component must meet these standards:
 <div onClick={handleClick}>Click me</div>               // ✗ Not focusable
 <div role="button" tabIndex={0} onClick={handleClick}    // ✓ But prefer <button>
      onKeyDown={e => {
+       if (e.key === 'Enter') handleClick();
        if (e.key === ' ') e.preventDefault();
-       if (e.key === 'Enter' || e.key === ' ') handleClick();
+     }}
+     onKeyUp={e => {
+       if (e.key === ' ') handleClick();
      }}>
   Click me
 </div>
--- a/skills/frontend-ui-engineering/SKILL.md
+++ b/skills/frontend-ui-engineering/SKILL.md
@@ -173,7 +173,10 @@ Every component must meet these standards:
 <button onClick={handleClick}>Click me</button>        // ✓ Focusable by default
 <div onClick={handleClick}>Click me</div>               // ✗ Not focusable
 <div role="button" tabIndex={0} onClick={handleClick}    // ✓ But prefer <button>
-     onKeyDown={e => (e.key === 'Enter' || e.key === ' ') && handleClick()}>
+     onKeyDown={e => {
+       if (e.key === ' ') e.preventDefault();
+       if (e.key === 'Enter' || e.key === ' ') handleClick();
+     }}>
   Click me
 </div>
 ```


addyosmani (commented):
Nice catch. Space-key default scroll on a focused role="button" div is one of those subtle a11y bugs that only shows up when you actually try to use the keyboard, and the original single-line ternary version papered right over it. Sorry bout that! The fix is correct and the repro in the issue is convincing.

Two small things worth considering while you're in there. First, Enter doesn't technically need preventDefault (no default action to suppress on a div), so guarding it only for Space would keep the intent clearer - something like if (e.key === ' ') e.preventDefault() before the handler runs. Second, native buttons fire on Space keyup, not keydown, because Space-down on a real <button> shows the pressed state and only activates on release. If the goal is to mimic <button> behavior faithfully, moving the Space branch to onKeyUp matches that, though plenty of component libraries cheat and fire on keydown for both. Not a blocker, just a note.

If you're open to that change go for it, otherwise I can merge this change in in the next few days.

shaun0927:
Thanks — I went ahead and applied the non-blocking refinement you suggested.

Updated example behavior:
- `Enter` still activates on `onKeyDown`
- `Space` now only calls `preventDefault()` on `onKeyDown`
- `Space` activation moves to `onKeyUp` to better match native `<button>` timing

Kept the scope docs-only and re-checked with `git diff --check`.

If you prefer the simpler keydown version for brevity, I’m also happy to leave it there.

addyosmani (approved):
LGTM

PR #84, opened 2026-04-18
docs: expand INP coverage in performance checklist

nucliweb:
## Summary

- Add an INP measurement workflow (field data → DevTools → device testing) referencing CrUX Vis as the recommended field data source
- Extend the `web-vitals` snippet with the attribution build to expose `inputDelay`, `processingDuration`, and `presentationDelay` per interaction
- Add checklist item for deferring non-critical work out of event handlers (e.g. analytics, logging) to avoid delaying interaction response

## Test plan

- [x] Review checklist items for accuracy and completeness
- [x] Verify CrUX Vis link resolves correctly
- [x] Confirm attribution snippet matches current `web-vitals` API


addyosmani (approved):
Changes LGTM.

At some point, I wonder when/if we should lean into linking more deeply into web.dev documentation but from a token optimization perspective the smallest possible exact recommendations are also valuable on their own.

Issue #85, opened 2026-04-18
[RFC] Add web-performance-auditor agent


nucliweb:
## Motivation

This project already has a `performance-optimization` skill that guides the *process* of fixing performance problems — measure, identify, fix, verify. That covers new development well.

What's missing is a **reviewer persona** for existing products: something you invoke to audit what's already there and get a structured report of what's wrong and why. The same distinction that exists between writing secure code (a skill) and running a security audit (`security-auditor` agent).

A `web-performance-auditor` agent would fill this gap.

## Proposed use cases

**1. Pre-release gate**
Before deploying to production, invoke the agent to audit the current state of the product. Acts as a performance checkpoint in the shipping process — similar to how `security-auditor` could be used before a release.

**2. Diagnosing an existing product**
On a project with years of accumulated code, invoke the agent to surface performance issues across the codebase. Useful when metrics drop or when onboarding to a legacy project.

## Proposed scope

- **Core Web Vitals** — patterns that affect LCP, INP, CLS
- **Loading & resources** — images, fonts, scripts, CSS, preload/lazy strategies
- **Rendering & JS** — main thread, long tasks, bundle size, code splitting, re-renders
- **Network / TTFB** — caching headers, compression, API response times

## Proposed output format

Unlike the `security-auditor` (which uses Critical/High/Medium/Low), this agent would produce:

1. **Scorecard** — current state: LCP, INP, CLS values + Lighthouse score
2. **Prioritized findings** — each with location, description, user impact, and specific recommendation

The agent would work on both **source code** (static analysis for anti-patterns) and **tool output** (interpreting Lighthouse reports, DevTools traces, Chrome DevTools MCP, CrUX data).

## Relationship to existing content

The `performance-optimization/SKILL.md` is a *workflow* — it guides how to optimize. This agent is a *reviewer* — it identifies what needs optimizing and reports it. Same separation of concerns as `code-reviewer` vs the implementation skills.

The skill answers: *"I need to fix a performance problem — what's the process?"*
The agent answers: *"Is this product performant? What's broken and how bad is it?"*

## Open questions for discussion

- Does this fit the intended scope of the `agents/` directory, or is the scope too broad for a single agent?
- Should the output format follow the `security-auditor` pattern (Critical/High/Medium/Low) for consistency, or does a scorecard-first approach make more sense for performance?
- Is there value in splitting this into a narrower agent (e.g., CWV-only)?

jingchang0623-crypto:
[...]

PR #86, opened 2026-04-19
Orchestration guidance for personas, /ship, and Claude Code interop

addyosmani:
I've been getting asked about how `agent-skills` work when you're using orchestration of multiple agents so thought I'd put together a PR to solidify my thoughts.

The repo has a few personas (`code-reviewer`, `security-auditor`, `test-engineer`) and seven slash commands, but nothing in the docs said how they should fit together. `/ship` was possibly the worst offender: a six-item checklist that one agent ran top to bottom, when three of those items obviously belonged to specialist personas running in parallel.

This PR documents how the layers compose, rewrites `/ship` to actually do the parallel thing, and checks the whole design against orchestration, Claude Code's subagent and Agent Teams docs. 

### The changes

1. `agents/README.md` - a persona index and a decision matrix for picking direct invocation, a slash command, or an orchestrator. Includes a valid orchestration example (parallel fan-out) and the invalid one I most wanted to warn against: a "router" persona whose only job is to decide which other persona to call.
2. AGENTS.md orchestration section - makes the three-layer model (skill / persona / command) explicit for contributors.
3. `/ship` refactor - three phases: parallel fan-out, merge in the main context, go/no-go decision with a mandatory rollback plan.
4. Composition blocks on each persona - three bullets per file: when to invoke directly, which slash commands wrap it, and the rule that personas don't call other personas.
5. `references/orchestration-patterns.md` - five endorsed patterns, four anti-patterns, a decision flow, and a gate for when to add a new pattern (so this catalog doesn't bloat).
6. Claude Code compatibility - checked against the official docs. Mapping is clean. Updates name the platform plumbing directly: the Agent tool, `subagent_type`, and the plugin frontmatter fields that are silently ignored (`hooks`, `mcpServers`, `permissionMode`).
7. Agent Teams worked example - a competing-hypothesis debugging walkthrough showing when Agent Teams beats `/ship`'s fan-out, plus a warning against wrapping it as a `/debug` slash command (you'd lose the inter-teammate messaging that makes it work).

### Why should we bother?

Mostly to prevent one specific bad outcome. Without explicit guidance, somebody eventually builds a "router" persona that does nothing but decide which other persona to call. It's pure overhead, and once it's in the repo you basically can't take it back out. The docs now state the rule plainly and explain it. Conveniently, Claude Code enforces the same rule at the platform level - subagents can't spawn subagents - so the constraint holds whether or not anyone reads the docs.

The `/ship` change is the part you'll actually feel. Three reviewers in parallel produces noticeably better reports than one agent doing six things sequentially, because each specialist gets its own context and doesn't get distracted. Run a real PR through the new `/ship` and you get the security findings, the test coverage gaps, and the code-review nits returned in one merged decision with a rollback plan, instead of needing three separate sessions to get the same picture.

### Compatibility

Verified against Claude Code's published docs for [subagents](https://code.claude.com/docs/en/sub-agents) and [agent teams](https://code.claude.com/docs/en/agent-teams). Personas auto-discover from `agents/` when the plugin loads - no path config needed. Agent Teams is still experimental: requires `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS=1` and Claude Code v2.1.32+. 

### Test plan

- [x] Read the merged docs in order; cross-references resolve.
- [x] Cross-checked the Claude Code claims against the linked official docs.
- [x] Working tree clean after all seven commits.
- [ ] (Reviewer) eyeball the `/ship` rewrite for accuracy against the existing personas' output formats.

### DIFF
--- a/.claude/commands/ship.md
+++ b/.claude/commands/ship.md
@@ -23,6 +23,8 @@ Constraints (from Claude Code's subagent model):
 - Each subagent gets its own context window and returns only its report to this main session.
 - If you need teammates that talk to each other instead of just reporting back, use Claude Code Agent Teams and reference these personas as teammate types (see `references/orchestration-patterns.md`).
 
+**Persona resolution.** If you've defined your own `code-reviewer`, `security-auditor`, or `test-engineer` in `.claude/agents/` or `~/.claude/agents/`, those take precedence over this plugin's versions — `/ship` picks up your customizations automatically. This is intentional: plugin subagents sit at the bottom of Claude Code's scope priority table, so user-level definitions win by design.
+
 ## Phase B — Merge in main context
 
 Once all three reports are back, the main agent (not a sub-persona) synthesizes them:
--- a/.claude/commands/ship.md
+++ b/.claude/commands/ship.md
@@ -8,13 +8,20 @@ Invoke the agent-skills:shipping-and-launch skill.
 
 ## Phase A — Parallel fan-out
 
-Spawn the following sub-agents concurrently. Each should run in its own context and return a structured report:
+Spawn three subagents concurrently using the Agent tool. **Issue all three Agent tool calls in a single assistant turn so they execute in parallel** — sequential calls defeat the purpose of this command.
 
-1. **code-reviewer** — Run a five-axis review (correctness, readability, architecture, security, performance) on the staged changes or recent commits. Output the standard review template.
-2. **security-auditor** — Run a vulnerability and threat-model pass. Check OWASP Top 10, secrets handling, auth/authz, dependency CVEs. Output the standard audit report.
-3. **test-engineer** — Analyze test coverage for the change. Identify gaps in happy path, edge cases, error paths, and concurrency scenarios. Output the standard coverage analysis.
+In Claude Code, each call passes `subagent_type` matching the persona's `name` field:
 
-Do not run these sequentially. Do not let one persona call another. Each persona sees the same diff and produces its own perspective.
+1. **`code-reviewer`** — Run a five-axis review (correctness, readability, architecture, security, performance) on the staged changes or recent commits. Output the standard review template.
+2. **`security-auditor`** — Run a vulnerability and threat-model pass. Check OWASP Top 10, secrets handling, auth/authz, dependency CVEs. Output the standard audit report.
+3. **`test-engineer`** — Analyze test coverage for the change. Identify gaps in happy path, edge cases, error paths, and concurrency scenarios. Output the standard coverage analysis.
+
+In other harnesses without an Agent tool, invoke each persona's system prompt sequentially and treat their outputs as if returned in parallel — the merge phase still works.
+
+Constraints (from Claude Code's subagent model):
+- Subagents cannot spawn other subagents — do not let one persona delegate to another.
+- Each subagent gets its own context window and returns only its report to this main session.
+- If you need teammates that talk to each other instead of just reporting back, use Claude Code Agent Teams and reference these personas as teammate types (see `references/orchestration-patterns.md`).
 
 ## Phase B — Merge in main context
 
--- a/.claude/commands/ship.md
+++ b/.claude/commands/ship.md
@@ -1,17 +1,63 @@
 ---
-description: Run the pre-launch checklist and prepare for production deployment
+description: Run the pre-launch checklist via parallel fan-out to specialist personas, then synthesize a go/no-go decision
 ---
 
 Invoke the agent-skills:shipping-and-launch skill.
 
-Run through the complete pre-launch checklist:
+`/ship` is a **fan-out orchestrator**. It runs three specialist personas in parallel against the current change, then merges their reports into a single go/no-go decision with a rollback plan. The personas operate independently — no shared state, no ordering — which is what makes parallel execution safe and useful here.
 
-1. **Code Quality** — Tests pass, build clean, lint clean, no TODOs, no console.logs
-2. **Security** — npm audit clean, no secrets in code, auth in place, headers configured
-3. **Performance** — Core Web Vitals good, no N+1 queries, images optimized, bundle sized
-4. **Accessibility** — Keyboard nav works, screen reader compatible, contrast adequate
-5. **Infrastructure** — Env vars set, migrations ready, monitoring configured
-6. **Documentation** — README current, ADRs written, changelog updated
+## Phase A — Parallel fan-out
 
-Report any failing checks and help resolve them before deployment.
-Define the rollback plan before proceeding.
+Spawn the following sub-agents concurrently. Each should run in its own context and return a structured report:
+
+1. **code-reviewer** — Run a five-axis review (correctness, readability, architecture, security, performance) on the staged changes or recent commits. Output the standard review template.
+2. **security-auditor** — Run a vulnerability and threat-model pass. Check OWASP Top 10, secrets handling, auth/authz, dependency CVEs. Output the standard audit report.
+3. **test-engineer** — Analyze test coverage for the change. Identify gaps in happy path, edge cases, error paths, and concurrency scenarios. Output the standard coverage analysis.
+
+Do not run these sequentially. Do not let one persona call another. Each persona sees the same diff and produces its own perspective.
+
+## Phase B — Merge in main context
+
+Once all three reports are back, the main agent (not a sub-persona) synthesizes them:
+
+1. **Code Quality** — Aggregate Critical/Important findings from `code-reviewer` and any failing tests, lint, or build output. Resolve duplicates between reviewers.
+2. **Security** — Promote any Critical/High `security-auditor` findings to launch blockers. Cross-reference with `code-reviewer`'s security axis.
+3. **Performance** — Pull from `code-reviewer`'s performance axis; cross-check Core Web Vitals if applicable.
+4. **Accessibility** — Verify keyboard nav, screen reader support, contrast (not covered by the three personas — handle directly here, or invoke the accessibility checklist).
+5. **Infrastructure** — Env vars, migrations, monitoring, feature flags. Verify directly.
+6. **Documentation** — README, ADRs, changelog. Verify directly.
+
+## Phase C — Decision and rollback
+
+Produce a single output:
+
+```markdown
+## Ship Decision: GO | NO-GO
+
+### Blockers (must fix before ship)
+- [Source persona: Critical finding + file:line]
+
+### Recommended fixes (should fix before ship)
+- [Source persona: Important finding + file:line]
+
+### Acknowledged risks (shipping anyway)
+- [Risk + mitigation]
+
+### Rollback plan
+- Trigger conditions: [what signals would prompt rollback]
+- Rollback procedure: [exact steps]
+- Recovery time objective: [target]
+
+### Specialist reports (full)
+- [code-reviewer report]
+- [security-auditor report]
+- [test-engineer report]
+```
+
+## Rules
+
+1. The three Phase A personas run in parallel — never sequentially.
+2. Personas do not call each other. The main agent merges in Phase B.
+3. The rollback plan is mandatory before any GO decision.
+4. If any persona returns a Critical finding, the default verdict is NO-GO unless the user explicitly accepts the risk.
+5. If the change is small enough that fan-out adds more overhead than value (e.g. a single typo fix), skip the fan-out and run a single review pass — `/ship` is designed for production-bound changes, not trivial edits.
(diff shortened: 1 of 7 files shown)


addyosmani:
cc @federicobartoli in case you have thoughts on this :)

.claude/commands/ship.md
```diff
@@ -1,17 +1,70 @@
...
-4. **Accessibility** — Keyboard nav works, screen reader compatible, contrast adequate
-5. **Infrastructure** — Env vars set, migrations ready, monitoring configured
-6. **Documentation** — README current, ADRs written, changelog updated
+## Phase A — Parallel fan-out
 
-Report any failing checks and help resolve them before deployment.
-Define the rollback plan before proceeding.
+Spawn three subagents concurrently using the Agent tool. **Issue all three Agent tool calls in a single assistant turn so they execute in parallel** — sequential calls defeat the purpose of this command.
+
+In Claude Code, each call passes `subagent_type` matching the persona's `name` field:
```
addyosmani, line 13: Note to self: Right now, I would say Claude Code is one of the most popular solutions and so adding specifics for it may be fine. At the same time, we probably want to be mindful of adding tool-specific guidance so something to discuss.

agents/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+Plugin agents do not support `hooks`, `mcpServers`, or `permissionMode` frontmatter — those fields are silently ignored. Avoid relying on them when authoring new personas here.
+
+## Adding a new persona
+
+1. Create `agents/<role>.md` with the same frontmatter format used by existing personas.
+2. Define the role, scope, output format, and rules.
+3. Add a **Composition** block at the bottom (Invoke directly when / Invoke via / Do not invoke from another persona).
+4. Add the persona to the table at the top of this file.
+5. If the persona enables a new orchestration pattern, document it in `references/orchestration-patterns.md` rather than inventing the pattern in the persona file itself.
```
addyosmani, line 120: Side-note: I know these are quite a few additions. If we end up on the fence, one alternative is to have a workflow path just for multi-agent/orchestration and do a break out.

federicobartoli:
> cc @federicobartoli in case you have thoughts on this :)

Checking right away 💪

.claude/commands/ship.md
```diff
@@ -1,17 +1,70 @@
...
+- [test-engineer report]
+```
+
+## Rules
+
+1. The three Phase A personas run in parallel — never sequentially.
+2. Personas do not call each other. The main agent merges in Phase B.
+3. The rollback plan is mandatory before any GO decision.
+4. If any persona returns a Critical finding, the default verdict is NO-GO unless the user explicitly accepts the risk.
+5. If the change is small enough that fan-out adds more overhead than value (e.g. a single typo fix), skip the fan-out and run a single review pass — `/ship` is designed for production-bound changes, not trivial edits.
```
federicobartoli: What do you think about making it “deterministic” when it’s small enough?
For example:
```
Before shipping, ask yourself:
- Does this change touch 2+ files? (Yes -> use fan-out)
-  Is this change 50+ lines? (Yes -> use fan-out)
```

agents/README.md
```diff
@@ -0,0 +1,120 @@
...
+2. **Personas do not invoke other personas.** Composition is the job of slash commands or the user. On Claude Code this is also a hard platform constraint — *"subagents cannot spawn other subagents"* — so the rule is enforced for you.
+3. A persona may invoke skills (the *how*).
+4. Every persona file ends with a "Composition" block stating where it fits.
+
+## Claude Code interop
+
+The personas in this repo are designed to work as Claude Code subagents and as Agent Teams teammates without modification:
+
+- **As subagents:** auto-discovered when this plugin is enabled (no path config needed). Use the Agent tool with `subagent_type: code-reviewer` (or `security-auditor`, `test-engineer`). `/ship` is the canonical example.
+- **As Agent Teams teammates** (experimental, requires `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS=1`): reference the same persona name when spawning a teammate. The persona's body becomes the teammate's system prompt.
```
federicobartoli: Small accuracy issue: the persona body doesn't *become* the teammate's system prompt — it's appended to it.
Per the [Agent Teams docs](https://code.claude.com/docs/en/agent-teams#use-subagent-definitions-for-teammates):

> The teammate honors that definition's `tools` allowlist and `model`, and the definition's body is **appended to the teammate's system prompt as additional instructions rather than replacing it**.

This matters for persona authors: they'll still inherit the lead's team-coordination instructions (SendMessage, task-list tools, etc.) on top of whatever they wrote. A persona that assumes it fully owns the prompt might be surprised.

Suggested wording:
> reference the same persona name when spawning a teammate. The persona's body is **appended to** the teammate's system prompt as additional instructions (not a replacement).

references/orchestration-patterns.md
```diff
@@ -0,0 +1,368 @@
...
+                       and any synchronous network calls added recently
+  - test-engineer  — propose tests that would distinguish between the
+                     hypotheses and check coverage gaps in checkout
+
+Have them message each other directly to challenge each other's
+theories. Update findings as consensus emerges. Only converge when
+two teammates agree they can disprove the others'.
+```
+
+The lead spawns three teammates referencing the existing persona names. The persona body becomes each teammate's system prompt; the prompt above becomes their task.
```
federicobartoli: Same issue as agents/README.md:108 — the persona body is appended to the teammate's system prompt, not a replacement.

Per the [Agent Teams docs](https://code.claude.com/docs/en/agent-teams#use-subagent-definitions-for-teammates):

> the definition's body is **appended to the teammate's system prompt as additional instructions rather than replacing it**.

Suggested rewrite:
> The lead spawns three teammates referencing the existing persona names. The persona body is appended to each teammate's system prompt as additional instructions; the trigger prompt above becomes their task.

references/orchestration-patterns.md
```diff
@@ -0,0 +1,368 @@
...
+
+| | Subagents | Agent Teams |
+|--|-----------|-------------|
+| Coordination | Main agent fans out, sub-agents only report back | Teammates message each other, share a task list |
+| Context | Own context window per subagent | Own context window per teammate |
+| When to use | Independent tasks producing reports | Collaborative work needing discussion |
+| Status | Stable | Experimental — requires `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS=1` |
+| Cost | Lower | Higher — each teammate is a separate Claude instance |
+
+**The personas in this repo work in both modes.** When spawned as subagents (e.g. by `/ship`), they report findings to the main session. When spawned as teammates (`Spawn a teammate using the security-auditor agent type…`), they can challenge each other's findings directly. The persona definition is the same; only the spawning context changes.
```
federicobartoli, line 137: Worth qualifying "The persona definition is the same; only the spawning context changes." A few frontmatter fields are actually dropped when the persona runs as a teammate vs. as a subagent.

Per the [Agent Teams docs](https://code.claude.com/docs/en/agent-teams#use-subagent-definitions-for-teammates):

> The `skills` and `mcpServers` frontmatter fields in a subagent definition are **not applied when that definition runs as a teammate**. Teammates load skills and MCP servers from your project and user settings, the same as a regular session.

So if a future persona declares `skills: [...]` in its frontmatter, that block works as a subagent (inside /ship) and silently no-ops as a teammate. Suggested addition:

> The persona definition is the same; only the spawning context changes. Note that `skills` and `mcpServers` frontmatter in a persona are honored as a subagent but ignored as a teammate (teammates load those from project/user settings) — require persona-specific skills at the session level if both modes matter.

federicobartoli (commented):
I left a few small comments, hope they’re helpful. Otherwise, everything looks great,a really awesome addition 💪

federicobartoli:
> I left a few small comments, hope they’re helpful. Otherwise, everything looks great,a really awesome addition 💪

Small addition: worth documenting this edge case. Per the [scope table](https://code.claude.com/docs/en/sub-agents#choose-the-subagent-scope), plugin subagents are the lowest priority , so a user-defined code-reviewer in .claude/agents/ or ~/.claude/agents/ will silently override the plugin's version when /ship runs.

Probably worth a one-liner in ship.md framing this as intentional (user customization wins), so people don't spend time debugging "why isn't /ship using the persona I expected?

nucliweb:
Related Issue #85 👀

addyosmani:
> > I left a few small comments, hope they’re helpful. Otherwise, everything looks great,a really awesome addition 💪
> 
> Small addition: worth documenting this edge case. Per the [scope table](https://code.claude.com/docs/en/sub-agents#choose-the-subagent-scope), plugin subagents are the lowest priority , so a user-defined code-reviewer in .claude/agents/ or ~/.claude/agents/ will silently override the plugin's version when /ship runs.
> 
> Probably worth a one-liner in ship.md framing this as intentional (user customization wins), so people don't spend time debugging "why isn't /ship using the persona I expected?

Good point. The priority ordering is easy to forget here and plugin subagents sit at the bottom of the scope table, not the top, and the override is silent. I'll add a one-liner to ship.md framing it as intentional so nobody loses an hour debugging it.

references/orchestration-patterns.md
addyosmani: Ah! Yeah, this is a real gap. 

I had the plugin-level frontmatter note earlier in the file (hooks / mcpServers / permissionMode silently ignored for plugin subagents) but missed the second axis - that skills and mcpServers get dropped when any subagent runs as a teammate, plugin or not. 

Taking your addition almost verbatim. Realistic failure mode: someone reuses a subagent-with-skills as a teammate and then spends half an hour wondering why their skills never loaded.

references/orchestration-patterns.md
addyosmani: Same mistake, same fix - taking your wording. Thanks for catching both instances in one review; would've been annoying to ship this with only one of them corrected.

agents/README.md
addyosmani: You're right "becomes" overstates it. I clocked the appending bit when I first read the Agent Teams docs and then wrote it loosely here. Taking your wording. It's a useful one to get right bc a persona author assuming they own the whole prompt would definitely be surprised by the SendMessage / task-list instructions sitting on top. Good catch.

.claude/commands/ship.md
addyosmani: Good call on making this concrete. How about anchoring to size AND surface area?

Something like:

Skip fan-out only if ALL of the following are true:

- 2 files or fewer
- Under 50 lines of diff
- No touches to auth, payments, data access, or config/env
- Otherwise, default to fan-out.

That keeps it mostly deterministic but doesn't let a small-but-risky change slip through. Happy to adjust the thresholds - 2 files / 50 lines was me following your numbers.

addyosmani:
Thanks for the review, @federicobartoli! Hopefully addressed: 

SHA | Commit | Addresses
-- | -- | --
93dd9b4 | docs: correct Agent Teams system-prompt handling (appended, not replaced) | Points 1 + 2 (same fix, two files)
0bf7398 | docs(references): note skills/mcpServers frontmatter dropped as teammate | Point 3
de62bb9 | refactor(ship): replace hand-wavy skip-fan-out rule with concrete thresholds | Point 4
2bee9b3 | docs(ship): note user-defined personas override plugin versions | Point 5

addyosmani:
> Related Issue #85 👀

Good shout. #85 could naturally slot in as another persona. Happy for us to revisit the orchestration docs once that one lands to make sure the decision matrix accommodates the personas cleanly.

.claude/commands/ship.md
federicobartoli: Love the sensitive-paths addition , a 10-line auth change is genuinely riskier than a 500-line test-file rename, so gating on blast radius alongside size is the right call. The inverted framing ("skip only if ALL true") also makes fan-out the safe default, which matches /ship's intent.

2 files / 50 lines feels right for a production-bound command , erring on the conservative side is the correct bias here. Happy to revisit if real usage surfaces a case where the thresholds need tuning.

federicobartoli:
> > > I left a few small comments, hope they’re helpful. Otherwise, everything looks great,a really awesome addition 💪
> > 
> > 
> > Small addition: worth documenting this edge case. Per the [scope table](https://code.claude.com/docs/en/sub-agents#choose-the-subagent-scope), plugin subagents are the lowest priority , so a user-defined code-reviewer in .claude/agents/ or ~/.claude/agents/ will silently override the plugin's version when /ship runs.
> > Probably worth a one-liner in ship.md framing this as intentional (user customization wins), so people don't spend time debugging "why isn't /ship using the persona I expected?
> 
> Good point. The priority ordering is easy to forget here and plugin subagents sit at the bottom of the scope table, not the top, and the override is silent. I'll add a one-liner to ship.md framing it as intentional so nobody loses an hour debugging it.

Verified all four against the doc quotes and the scope-table mapping ,clean fixes.

Quick callout on de62bb9: your sensitive-paths carve-out is a better rule than the raw size thresholds I'd suggested. A "small" change to auth or config/env isn't actually small, and the carve-out catches that cleanly. Nice upgrade.

Ready to merge from my side.

addyosmani:
Thanks for the review, @federicobartoli! Sounds good to me. Merging away :)

PR #82, opened 2026-04-17
feat: add Codex plugin setup and installer

WooRaZil-Boy:
Thank you for putting this project together. I have been using it heavily, and it has been genuinely useful in day-to-day coding work.

I personally use Codex often in addition to Claude Code, so I ended up building a Codex plugin workflow for my own setup. After iterating on it in practice, I wanted to contribute back a cleaned-up version that might also be useful to other Codex users.

## Summary

This PR adds a Codex-specific installation path while keeping the existing shared content as the single source of truth.

### What this change adds

- a Codex-specific plugin scaffold under `.codex/`
- a dedicated setup guide in `docs/codex-setup.md`
- an installer that builds a self-contained personal Codex plugin bundle
- wrapper skills for the main lifecycle entrypoints:
  - `spec`
  - `plan`
  - `build`
  - `test`
  - `review`
  - `ship`
  - `code-simplify`

## Design

The goal was to support Codex without introducing a second maintained copy of the shared skill set inside the repository.

The existing repository structure remains the source of truth:

- `skills/`
- `references/`
- `agents/`
- `.claude/commands/`

At install time, the Codex installer assembles a personal plugin bundle by combining:

- the shared skills from `skills/`
- the shared reference docs from `references/`
- the shared agent personas from `agents/`
- freshly generated Codex-only wrapper skills based on `.claude/commands/`

## Notes

This PR is based on OpenAI's official Codex plugin documentation:

- https://developers.openai.com/codex/plugins/build
- https://developers.openai.com/codex/plugins

I used those docs as the reference for the plugin structure, local marketplace flow, and installation model.

I also tried to keep the Codex integration isolated from the existing Claude Code workflow, and kept the setup details in a dedicated `docs/codex-setup.md` guide rather than expanding the main README.

In practice, Codex needs to be fully restarted before newly added marketplace or plugin entries appear. For the CLI, that may require fully stopping the running process and starting a fresh terminal session.

## Test plan

- `./.codex/scripts/install-plugin.sh`
- verified that the generated personal plugin bundle includes:
  - `skills/`
  - `references/`
  - `agents/`
- verified that installation does not mutate the repository worktree
- verified the installation flow on a fresh machine
- verified that Codex needs to be fully restarted before the plugin appears

## Screenshots

Codex recognizing the installed plugin and invoking the wrapper skills successfully:

app_0

cli_0

PR #88, opened 2026-04-19
feat: add Codex plugin support without duplicating skills

federicobartoli:
## Summary

Make this repo installable as a Codex plugin with a single command, while keeping `skills/` as the single source of truth shared with Claude Code. No files are copied, no installer script, no wrapper skills.

```bash
codex marketplace add addyosmani/agent-skills
```

## What this adds

- `codex/.codex-plugin/plugin.json` — Codex manifest (`"skills": "./skills/"`).
- `codex/skills` — git-tracked symlink to `../skills` (9 bytes, mode `120000`). Keeps the plugin directory self-contained without duplicating any skill content.
- `.agents/plugins/marketplace.json` — marketplace entry declaring the plugin at `./codex`.
- `docs/codex-setup.md` — install and usage guide, including a fallback that replaces the symlink with a local copy for Windows users.

Net change: 4 files, ~78 lines, 0 duplicated skill content.

## Why no duplication

Codex and Claude Code share the **same `SKILL.md` format** (YAML frontmatter with `name` + `description`). The [Codex plugin docs](https://developers.openai.com/codex/plugins/build) define `skills` as a relative path to a folder; pointing it at the existing `skills/` via a symlink means one file serves both platforms.

`.claude/commands/` and `agents/` stay Claude-specific — Codex natively supports only skills, apps, and MCP servers, so bundling commands/personas inside a Codex plugin isn't meaningful. Codex users invoke skills directly.

## Why a `codex/` subdirectory

Codex's manifest validator rejects plugins at the marketplace root: `source.path` must start with `./` and stay inside the marketplace root. `codex/` is that subdirectory.

## Verified

Tested with `codex-cli 0.121.0`:
- `codex marketplace add federicobartoli/agent-skills@feat/codex-plugin-support` succeeds.
- Plugin registered in `~/.codex/config.toml`, cached under `~/.codex/plugins/cache/agent-skills/`.
- All 21 skills visible in the plugin's Skills list via the Codex UI.

## Open to alternatives

If a symlink in the repo is undesirable, I'm happy to switch to either:
- **Restructure** — `git mv skills codex/skills` and point Claude Code's plugin.json at the new path (no symlink, but larger diff moving 21 directories), or
- **Manual step** — drop the symlink and document a `cp -R skills codex/skills` step after `git clone` (no symlink, but install is no longer a single `codex marketplace add`).

Just let me know which direction you prefer.

## Relation to #82

Lighter-weight alternative to #82 (which adds ~900 lines: installer script, wrapper skills regenerated from `.claude/commands/`, bundled copies of `skills/`/`references/`/`agents/`). Same intent; here Codex reads the existing `skills/` directly. Happy to defer to whichever direction the maintainer prefers.

## Test plan

- [x] `codex marketplace add federicobartoli/agent-skills@feat/codex-plugin-support` succeeds
- [x] Marketplace appears in `~/.codex/config.toml`
- [x] Plugin cache created under `~/.codex/plugins/cache/agent-skills/`
- [x] All 21 skills visible in Codex UI under the plugin's Skills list
- [ ] Skills invokable in chat after UI install (pending maintainer verification)

Screenshot 2026-04-19 alle 16 43 59
Screenshot 2026-04-19 alle 16 45 08
Screenshot 2026-04-19 alle 16 45 18

Next ideas: consider auto-generating (?) or at least avoiding repetition of the manifest; also consider enriching the manifest with marketplace metadata (https://developers.openai.com/codex/plugins/build#marketplace-metadata
).


brandonxiang:
why not to have claude commands like "spec", "plan", "build", "test", "review", "ship" inside codex？

ryankert01:
support to add skill in codex more smoothly. I personally want to use it.

docs/codex-setup.md
```diff
@@ -0,0 +1,43 @@
+# Using agent-skills with Codex
+
+This repository is also a [Codex plugin](https://developers.openai.com/codex/plugins/build). The same `skills/` directory used by Claude Code is consumed by Codex — no files are copied or duplicated.
+
+## Install (one command)
+
+```bash
+codex marketplace add addyosmani/agent-skills
```
ryankert01: nit: I think also need to update readme for QuickStart?

federicobartoli: yes, thank u man, i'll make the change as soon as I can :D

federicobartoli:
Pushed `d70d7a8` to update the install command for Codex CLI v0.122+.

In Codex CLI 0.122 the marketplace subcommand moved from `codex marketplace` to `codex plugin marketplace` (verified on 0.128.0). The README and `docs/codex-setup.md` now use `codex plugin marketplace add addyosmani/agent-skills`. The plugin manifest and marketplace.json schemas are unchanged — only the user-facing install command was renamed.

Test plan still valid; just substitute the new subcommand:

```bash
codex plugin marketplace add federicobartoli/agent-skills@feat/codex-plugin-support
```

federicobartoli:
Hi @addyosmani ,gentle ping when you have a chance.

This keeps `skills/` as the single source of truth while making the repo installable as a Codex plugin via `codex marketplace add addyosmani/agent-skills`. No duplicated skill content, and I verified the flow end-to-end with `codex-cli 0.121.0`.

Happy to adapt the implementation if you’d prefer a different structure.

addyosmani:
Federico, first: sorry for the silence on this one, you pinged twice and it slipped through while the repo was moving fast. The design intent here is exactly right (skills/ stays the single source of truth, no copies, no installer scripts), and I want to land Codex support built on this PR. Two things changed since April that need a rework before I can merge it, and they're both quick.

The bigger one is the symlink. `codex/skills -> ../skills` is clever, but on Windows git checks out symlinks as plain text files by default, so `codex/skills` becomes a file literally containing "../skills" and any Codex install resolving that path breaks. We already have one Windows path bug on file (#214) and this repo gets cloned across 70+ agents, so I don't want to bake in a whole class of issues I can't reproduce locally. The good news is the symlink turns out to be unnecessary: since #227 landed, the repo's convention is root-level per-tool manifests (`.claude-plugin/`, root `plugin.json` for Antigravity, root `commands/`). If you move your manifest to `.codex-plugin/plugin.json` at the root, `"skills": "./skills/"` just works with no symlink and no `codex/` directory at all, and it matches how every other tool integrates now.

One more heads-up: #332 opened in June covering the same ground with the root layout, so you two collided (that one's on me too, the PR-overlap guideline didn't exist when you opened this). You were here first, and yours is the stronger PR because of `docs/codex-setup.md`, which #332 doesn't have. So the ask is: root `.codex-plugin/`, drop the symlink and `codex/` dir, point the marketplace source at `./`, keep your docs, and refresh the README entry against the current Quick Start layout. Do that and I'll merge it and close #332 with credit to @kacperklusek for independently validating the root approach. If you're short on time, say the word and I'll push the rework to your branch instead. Thanks for the patience on this one.

federicobartoli:
Thanks again for the review @addyosmani ! I pushed an update addressing the Codex packaging feedback:

- Moved the Codex manifest to the repo root at `.codex-plugin/plugin.json`.
- Removed the `codex/skills -> ../skills` symlink and the `codex/` wrapper layout.
- Updated `.agents/plugins/marketplace.json` so the Codex marketplace source points at `./`.
- Refreshed the Codex setup docs to describe the root plugin layout.

One small implementation note: once the Codex plugin source points at the repo root, Codex can also auto-discover the existing `hooks/hooks.json` file as a plugin-bundled lifecycle hook. That hook is Claude-oriented, so I avoided changing the hook scripts themselves and instead made the Codex manifest declare an explicit empty hook config:

```json
"hooks": {
  "hooks": {}
}
```

This prevents Codex from falling back to the default `hooks/hooks.json` discovery path, while leaving the existing Claude hook files untouched.

I verified the local Codex plugin install after this change:

- `bash hooks/session-start-test.sh`
- local `codex exec` invocation of `$agent-skills:spec-driven-development`

The skill loads successfully in Codex, and the Claude-oriented SessionStart hook no longer runs in Codex.

addyosmani:
That's exactly the rework, thank you @federicobartoli. Root manifest, no symlink, and the empty hooks guard is a genuinely sharp catch: I hadn't considered that a root-scoped Codex plugin would auto-discover the Claude-oriented hooks/hooks.json, and heading that off in the manifest is the right fix.

Verified locally on top of green CI: manifests parse, no symlinks anywhere in the diff, zero conflicts against main, and the full validator plus eval suite passes. Merging. Codex folks: `codex plugin marketplace add addyosmani/agent-skills` and see docs/codex-setup.md. Thanks for sticking with this one through a three-month detour.

PR #90, opened 2026-04-22
fix: escape session-start hook JSON

### LINKED ISSUE #89: session-start hook outputs invalid JSON when `SKILL.md` contains quotes/newlines
**Affected file:** `sk/code-simplification/hooks/session-start.sh`

### What I hit
I ran this hook and tried to parse the output as JSON and got parse errors.

I’m thinking it is because `SKILL.md` is injected raw into:

`"message": "agent-skills loaded...\\n\\n$CONTENT"`

so quoted markdown and newlines can break JSON syntax.

### Repro
1. Run `bash hooks/session-start.sh | python3 -c 'import sys, json; json.loads(sys.stdin.read())'`.
2. Keep `SKILL.md` with quote-heavy markdown or control chars.
3. Hook output fails to parse as JSON.

### Expected
Startup output should be valid JSON for all markdown contents.

### Actual
The startup payload can be malformed and unusable.

### Suggested fix
JSON-escape the `SKILL.md` content before interpolation.

Bortlesboat:
## Summary
- Generate the SessionStart hook payload with `JSON.stringify` so Markdown quotes, backslashes, and newlines cannot corrupt the JSON output.
- Add an executable hook test that parses the payload and checks the injected using-agent-skills content is present.

Fixes #89.

## Verification
- Red repro before fix: `bash hooks/session-start.sh | python3 -c 'import sys,json; json.loads(sys.stdin.read())'` failed with `Invalid control character`.
- `git diff --check --cached`
- LF-normalized staged snapshot: `bash -n hooks/session-start.sh && bash -n hooks/session-start-test.sh && bash hooks/session-start-test.sh`
- LF-normalized staged snapshot: `bash hooks/session-start.sh | python3 -c 'import sys,json; payload=json.loads(sys.stdin.read()); assert payload["priority"] == "IMPORTANT"; assert "# Using Agent Skills" in payload["message"]'`

### DIFF
--- a/hooks/session-start-test.sh
+++ b/hooks/session-start-test.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+# session-start-test.sh - Tests for the SessionStart hook JSON payload
+
+set -euo pipefail
+
+tmp_payload="$(mktemp)"
+trap 'rm -f "$tmp_payload"' EXIT
+
+payload="$(bash hooks/session-start.sh)"
+printf '%s' "$payload" > "$tmp_payload"
+
+PAYLOAD_PATH="$tmp_payload" node <<'NODE'
+const fs = require('fs');
+
+const payload = JSON.parse(fs.readFileSync(process.env.PAYLOAD_PATH, 'utf8'));
+
+if (payload.priority !== 'IMPORTANT') {
+  throw new Error(`expected IMPORTANT priority, got ${payload.priority}`);
+}
+
+if (!payload.message.includes('agent-skills loaded.')) {
+  throw new Error('message is missing startup preface');
+}
+
+if (!payload.message.includes('# Using Agent Skills')) {
+  throw new Error('message is missing using-agent-skills content');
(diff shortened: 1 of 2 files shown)


addyosmani (approved):
Good catch! this is a real bug, and JSON.stringify is the right fix. Thanks especially for adding a test alongside it.

addyosmani:
After reviewing this further, the bug got fixed in main while your PR was open - the test file is great though, would you mind opening a follow-up PR with just that? I'm going to close this PR just to keep that focused if you do decide to submit a follow-up rebased against main. 

More context:

main got a parallel fix for the same bug, using jq instead of node. Both implementations correctly escape SKILL.md content -  main's version additionally has the graceful "if jq isn't installed, fall back to INFO message" pattern.

Bortlesboat:
Thanks again for the guidance here. I opened the focused follow-up with just the test, rebased on `main`: #132

PR #73, opened 2026-04-13
Feature: Standardize spec directory structure to prevent overwriting

sbusanelli:
Fixes #50

## Changes
- Update spec command to save specs to docs/features/[feature-name]/spec.md
- Update plan command to read specs from and save plans to feature directories
- Update getting-started documentation to reflect new directory structure
- Enable multiple simultaneous features without overwriting artifacts

## New Directory Structure
docs/features/[feature-name]/
  - spec.md     # Feature specification
  - plan.md     # Implementation plan
  - todo.md     # Task list

## Benefits
- Prevents spec overwriting when working on multiple features
- Enables simultaneous feature development
- Organizes all feature artifacts in one location
- Maintains backward compatibility with existing workflows

## Implementation Details
- Feature name derived from objective (e.g., 'user-authentication')
- Commands updated to use feature-specific paths
- Documentation updated to reflect new structure
- Maintains existing skill workflows while improving organization

PR #93, opened 2026-04-23
Feature: spec directory

jsls2517:
source from https://github.com/addyosmani/agent-skills/pull/73/changes/87de68f85dbb61e2c2509a41aa2c0329477bb161

## Changes
- Update spec command to save specs to docs/features/[feature-name]/spec.md
- Update plan command to read specs from and save plans to feature directories
- Update getting-started documentation to reflect new directory structure
- Enable multiple simultaneous features without overwriting artifacts

## New Directory Structure
docs/features/[feature-name]/
  - spec.md     # Feature specification
  - plan.md     # Implementation plan
  - todo.md     # Task list

### DIFF
--- a/.claude/commands/plan.md
+++ b/.claude/commands/plan.md
@@ -13,4 +13,8 @@ Read the existing spec (SPEC.md or equivalent) and the relevant codebase section
 5. Add checkpoints between phases
 6. Present the plan for human review
 
-Save the plan to tasks/plan.md and task list to tasks/todo.md.
+Save the plan and task list to the same feature directory:
+- docs/features/[feature-name]/plan.md
+- docs/features/[feature-name]/todo.md
+
+This keeps all feature artifacts together and enables multiple simultaneous features.
--- a/.claude/commands/spec.md
+++ b/.claude/commands/spec.md
@@ -12,4 +12,8 @@ Begin by understanding what the user wants to build. Ask clarifying questions ab
 
 Then generate a structured spec covering all six core areas: objective, commands, project structure, code style, testing strategy, and boundaries.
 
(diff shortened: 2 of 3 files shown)


nucliweb:
Thanks for this, and the multi-feature motivation is real, root-level `SPEC.md` does get overwritten when several features are in flight. The problem is that this changes only the producers (`/spec`, `/plan`) and leaves the consumers on the old paths, so it desyncs the workflow contract rather than moving it.

Concretely, `/build` breaks. `.claude/commands/build.md` looks for the spec only at `SPEC.md` / `docs/SPEC.md` / `spec/*` and requires the plan at `tasks/plan.md` (steps 1–3), so after this change it won't find what `/spec` and `/plan` now write under `docs/features/[name]/`. Step 2 also treats anything outside the expected artifact paths as stray uncommitted work, so `/build` would actually stop. On top of that, `skills/planning-and-task-breakdown` and `skills/spec-driven-development` both pin `tasks/plan.md` / `tasks/todo.md` as "the convention downstream commands expect," and `docs/adoption-guide.md` documents the old paths, so the commands and skills would now contradict each other. (CI's command-parity check compares descriptions, not artifact paths, so green CI doesn't cover this.)

To land the per-feature layout it needs to be one consistent convention change across the whole chain: `build.md`, `planning-and-task-breakdown/SKILL.md`, `spec-driven-development/SKILL.md`, and `adoption-guide.md`, alongside the two command files here. It's also adjacent to the existing spec-directory PRs (#73 and siblings), so it'd be worth coordinating on one agreed layout rather than landing a partial rename. As it stands I'd hold it.

addyosmani:
Thanks. Same area as the spec-structure work now in #383 and #431, so closing this in favor of that direction.

PR #94, opened 2026-04-23
Migrate repository to native OpenCode assets

diegaccio:
## Summary
- migrate the repository from Claude-specific packaging to native OpenCode commands, skills, agents, and plugins
- move shared references under `.opencode/` and add repository validation for the new layout
- add Unix and PowerShell bootstrap scripts for importing these reusable OpenCode assets into existing repositories
## Verification
- `node scripts/validate-opencode.mjs`


diegaccio:
Sorry about that, I opened the pull request against the upstream repository instead of my own fork by mistake. That was an error on my side. I’ll use the fork as the source for follow-up changes and keep future PRs scoped correctly.

PR #100, opened 2026-04-24
Add Gemini CLI slash command support

monraza:
Introduces `.gemini/commands/` with TOML definitions for 7 slash commands, enabling Gemini CLI to auto-discover and register them from the project root.

## Key Changes

- New `.gemini/commands/` directory with TOML files for `build`, `spec`,
  `planning`, `review`, `test`, `ship`, and `code-simplify`
- Each command invokes the corresponding agent-skills workflow automatically
- Named `/planning` instead of `/plan` — `/plan` conflicts with a Gemini CLI
  internal command name
- Updated `README.md` project structure tree to include `.gemini/commands/`
- Added a Slash Commands reference table to `docs/gemini-cli-setup.md`

## Design

Commands mirror the existing `.claude/commands/` set - same prompts, same lifecycle mapping - so Gemini CLI users get the same structured workflows Claude Code users already have. No skill content is duplicated; commands delegate directly to the existing skills.

> **Note:** Use `/planning` instead of `/plan` when working in Gemini CLI.

### DIFF
--- a/.gemini/commands/build.toml
+++ b/.gemini/commands/build.toml
@@ -1,8 +1,7 @@
-name = "build"
 description = "Implement the next task incrementally — build, test, verify, commit"
 
 prompt = """
-Invoke the agent-skills:incremental-implementation skill alongside agent-skills:test-driven-development.
+Invoke the incremental-implementation skill alongside test-driven-development.
 
 Pick the next pending task from the plan. For each task:
 
@@ -15,5 +14,5 @@ Pick the next pending task from the plan. For each task:
 7. Commit with a descriptive message
 8. Mark the task complete and move to the next one
 
-If any step fails, follow the agent-skills:debugging-and-error-recovery skill.
+If any step fails, follow the debugging-and-error-recovery skill.
 """
--- a/.gemini/commands/build.toml
+++ b/.gemini/commands/build.toml
@@ -1,0 +1,19 @@
+name = "build"
+description = "Implement the next task incrementally — build, test, verify, commit"
+
+prompt = """
+Invoke the agent-skills:incremental-implementation skill alongside agent-skills:test-driven-development.
+
+Pick the next pending task from the plan. For each task:
+
+1. Read the task's acceptance criteria
+2. Load relevant context (existing code, patterns, types)
+3. Write a failing test for the expected behavior (RED)
+4. Implement the minimum code to pass the test (GREEN)
+5. Run the full test suite to check for regressions
+6. Run the build to verify compilation
+7. Commit with a descriptive message
+8. Mark the task complete and move to the next one
+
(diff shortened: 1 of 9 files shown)


.gemini/commands/build.toml
```diff
@@ -0,0 +1,19 @@
+name = "build"
```
addyosmani: Afaik:

**1. `name` field is not part of the Gemini CLI TOML schema** (all 7 files)

Per the [[official docs](https://geminicli.com/docs/cli/custom-commands/)](https://geminicli.com/docs/cli/custom-commands/) and [[Google's gemini-cli docs](https://google-gemini.github.io/gemini-cli/docs/cli/custom-commands.html)](https://google-gemini.github.io/gemini-cli/docs/cli/custom-commands.html), the only valid top-level keys are:
- `prompt` (required)
- `description` (optional)

Command names are **derived from the filename** (`build.toml` → `/build`). Every file in this PR has a leading `name = "build"` etc. that Gemini CLI silently ignores. Harmless but should be removed for correctness.

**2. `subagent_type` reference in [[ship.toml](https://github.com/addyosmani/agent-skills/pull/100/files#diff-ship)](https://github.com/addyosmani/agent-skills/pull/100/files#diff-ship)**

The line `"In Gemini CLI, each call passes subagent_type matching the persona's name field"` looks like a copy-paste from the Claude Code version — `subagent_type` is a Claude Code Agent-tool parameter. Gemini CLI's subagent model uses a different invocation surface. This phrasing should be reworded or removed; the fallback paragraph about "Agent Teams / orchestration-patterns.md" already handles non-Claude harnesses.

**3. `agent-skills:` namespace prefix in every prompt body**

Lines like `Invoke the agent-skills:incremental-implementation skill` use Claude Code's `plugin:skill` syntax. Gemini CLI's skill system uses bare names (e.g., `incremental-implementation`) — when installed via `gemini skills install`, no `agent-skills:` prefix exists. Likely interpreted as natural language by the model, but technically inaccurate. Consider stripping the prefix in the Gemini variants.

**4. Minor: missing trailing newlines** on all 7 TOML files (POSIX nit).

## What's good
- `/plan` → `/planning` rename is correct — `/plan` does conflict
- Persona-resolution paragraph in ship.toml correctly swapped `.claude/agents/` → `.gemini/agents/`
- README tree disambiguation `(Claude Code)` / `(Gemini CLI)` is clean
- `docs/gemini-cli-setup.md` table is accurate

addyosmani (changes requested):
This is largely pretty good!

To move this forward:

"Drop the `name = ` line from all 7 TOMLs (filename is authoritative), reword the `subagent_type` sentence in ship.toml, and consider whether `agent-skills:` prefix is meaningful for Gemini's skill system or should be plain skill names."

monraza:
Thanks for resolving and the merge.

PR #101, opened 2026-04-25
Fix: JSON-escape hook content in session-start hook

### LINKED ISSUE #89: session-start hook outputs invalid JSON when `SKILL.md` contains quotes/newlines

ltianyi992:
Fixes #89

## Problem
The session-start hook was outputting invalid JSON when SKILL.md contained quotes or control characters like newlines. This broke JSON parsing for Claude Code hook consumers.

## Solution
Use `jq -cn` with `--arg` to properly construct and escape the JSON output. This ensures all special characters in the markdown content are safely escaped in the JSON string.

## Verification
- Tested with `bash hooks/session-start.sh | jq .` - now produces valid JSON
- Original issue: `json.JSONDecodeError: Invalid control character` - now fixed
- Works with markdown containing quotes, newlines, and other special characters

### DIFF
--- a/hooks/session-start.sh
+++ b/hooks/session-start.sh
@@ -8,13 +8,12 @@ META_SKILL="$SKILLS_DIR/using-agent-skills/SKILL.md"
 
 if [ -f "$META_SKILL" ]; then
   CONTENT=$(cat "$META_SKILL")
-  # Output as JSON for Claude Code hook consumption
-  cat <<EOF
-{
-  "priority": "IMPORTANT",
-  "message": "agent-skills loaded. Use the skill discovery flowchart to find the right skill for your task.\n\n$CONTENT"
-}
-EOF
+  # Use jq to properly escape and construct valid JSON
+  jq -cn \
+    --arg message "agent-skills loaded. Use the skill discovery flowchart to find the right skill for your task.
+
+$CONTENT" \
+    '{priority: "IMPORTANT", message: $message}'
 else
   echo '{"priority": "INFO", "message": "agent-skills: using-agent-skills meta-skill not found. Skills may still be available individually."}'
 fi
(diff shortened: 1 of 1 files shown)


addyosmani (approved):
Verified - current code emits invalid JSON against the real SKILL.md, your fix produces valid JSON. Happy to merge. 

One small ask: jq becomes a hard dependency for the hook to function. The original used pure shell. jq is not preinstalled on macOS (Homebrew) and isn't universal on Linux either. If a user runs Claude Code without jq on PATH, this hook will now fail rather than emit malformed JSON.

Could you either add a command `-v jq` guard with a graceful fallback message, or document the dependency in the hooks README? Just want users without jq to get a useful error rather than a silent failure on every session start.

smth like:

`if ! command -v jq >/dev/null; then echo '{"priority":"INFO","message":"agent-skills: jq required for session-start hook"}'; exit 0; fi ` so users get a useful message instead of a cryptic failure.

PR #108, opened 2026-04-26
add https installation guide in README

devmarkpro:
The README solution for installing the skill with HTTPS is a bit unnecessarily complex. Claude Code allows users to add marketplace plugins and install plugins over HTTPS, not SSH, if the full repo URL is provided.

ScreenShot 2026-05-01 at 04 19 43@2x

After installing the plugin, running ⁠git remote -v in the directory where Claude installed the skill shows that the repository was cloned using ⁠https instead of ⁠ssh.

ScreenShot 2026-05-01 at 04 20 41@2x

Here's also screenshot of the `/plugins menu` 

ScreenShot 2026-05-01 at 04 29 25@2x

And it can be updated using the menu just as any other plugin.

ScreenShot 2026-05-01 at 04 31 01@2x

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,12 +51,6 @@ Skills also activate automatically based on what you're doing — designing an A
 > claude plugin install agent-skills@addy-agent-skills
 > ```
 
-```
-claude plugin marketplace add https://github.com/addyosmani/agent-skills.git
-
-```
-
-
 **Local / development:**
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -45,10 +45,10 @@ Skills also activate automatically based on what you're doing — designing an A
 /plugin install agent-skills@addy-agent-skills
 ```
 
-> **SSH errors?** The marketplace clones repos via SSH. If you don't have SSH keys set up on GitHub, either [add your SSH key](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account) or run the commands below in your terminal (not Claude Code):
+> **SSH errors?** The marketplace clones repos via SSH. If you don't have SSH keys set up on GitHub, either [add your SSH key](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account) or use the full HTTPS URL to force the HTTPS cloning:
 > ```bash
-> claude plugin marketplace add https://github.com/addyosmani/agent-skills.git
-> claude plugin install agent-skills@addy-agent-skills
+> /plugin marketplace add https://github.com/addyosmani/agent-skills.git
+> /plugin install agent-skills@addy-agent-skills
 > ```
 
 **Local / development:**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,9 +45,10 @@ Skills also activate automatically based on what you're doing — designing an A
 /plugin install agent-skills@addy-agent-skills
 ```
 
-> **SSH errors?** The marketplace clones repos via SSH. If you don't have SSH keys set up on GitHub, either [add your SSH key](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account) or switch to HTTPS for fetches only:
+> **SSH errors?** The marketplace clones repos via SSH. If you don't have SSH keys set up on GitHub, either [add your SSH key](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account) or run the commands below in your terminal (not Claude Code):
 > ```bash
-> git config --global url."https://github.com/".insteadOf "git@github.com:"
+> claude plugin marketplace add https://github.com/addyosmani/agent-skills.git
+> claude plugin install agent-skills@addy-agent-skills
```
addyosmani: Thanks the instinct is right and the global `git config insteadOf` is genuinely a heavy hammer. Two issues with the proposed fix though:

Afaik 1. There's no `claude plugin marketplace add` shell subcommand in Claude Code's CLI — `/plugin marketplace add` is in-session only. The 'run in your terminal (not Claude Code)' framing inverts where the commands actually run. Could you keep them as slash commands?

2. Could you confirm empirically that passing a full HTTPS URL to `/plugin marketplace add` (rather than the `owner/repo` shorthand) actually forces HTTPS cloning? That's the premise this PR rests on, but it's not documented and I'd want a quick test result in the PR description before we recommend it.

If both check out, I'd suggest just swapping the shorthand for the full HTTPS URL inside the existing slash-command block, keeps the install model consistent, drops the global git-config workaround, no extra prose needed.

addyosmani (changes requested):

README.md
creelronnie10-lab: I have no idea when it comes to this stuff. I’m just trying to figure it out as I go.. Did someone teach you whenever you were educated on the matter

devmarkpro: Thanks @addyosmani that makes perfect sense. I've updated the PR to keep the command as a slash command and removed the extra prose.

Regarding the empirical test: I can confirm this works! I am actually working on a managed laptop where the SSH port to GitHub is closed by my organization. Using the shorthand owner/repo fails for me with an SSH error, but passing the full HTTPS URL to /plugin marketplace add bypasses the issue and clones successfully.

I have updated the command in the README to use `/plugin` version as you suggested. I've added the screenshots to the PR description as well to document the successful installation from my environment.

Regarding the last point, I still believe it makes more sense to keep the shorthand version as default installation guide. It aligns better with how the Claude Code marketplace works overall and will likely feel more familiar to the community. In any case, most users probably won’t encounter this issue.

addyosmani: Thanks for confirming that and I really appreciate you running the empirical test! The context about your managed laptop and the blocked SSH port is super helpful - that's exactly the kind of friction we want to help folks avoid. Going to go ahead with the merge!

PR #126, opened 2026-05-05
docs: add global installation guide for OpenCode

juanitourquiza:
OpenCode's built-in skill tool only recognizes a small set of natively registered skills. The skills in this repo are not automatically discoverable by that tool, making the per-workspace setup the only option.

This adds a 'Global Installation (All Projects)' section to the OpenCode setup guide that explains how to use the skills across all projects by:

1. Copying skills to ~/.agents/skills/
2. Creating a global ~/.claude/AGENTS.md with intent mapping
3. Referencing it from ~/.claude/CLAUDE.md (loaded in every session)

The agent then reads skill definitions on-demand via the read tool instead of the skill tool.

Also updates the README to mention both workspace-local and global options in the OpenCode quick-start summary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,10 @@ Add skill contents to your Windsurf rules configuration. See [docs/windsurf-setu
 
 Uses agent-driven skill execution via AGENTS.md and the `skill` tool.
 
-See [docs/opencode-setup.md](docs/opencode-setup.md).
+**Workspace-local:** Copy `AGENTS.md` + `skills/` into your project.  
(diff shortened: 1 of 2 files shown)


BennyAlex:
this only works with claude i guess but open code has many more models

addyosmani:
The OpenCode setup docs have been rewritten on main since April (docs/opencode-setup.md). Closing as superseded. If there's specific global-install content still missing, a small targeted PR is welcome.

PR #132, opened 2026-05-07
test: add session-start JSON regression coverage

Bortlesboat:
## Summary
- Add the focused `session-start` hook regression test from #90 on top of current `main`.
- Parse the hook output as JSON and assert the meta-skill content is injected when `jq` is available.
- Keep the test compatible with `main`'s INFO fallback when `jq` is absent.

Follow-up to #90 per @addyosmani's request.

## Verification
- Red check: LF-normalized temp fixture using pre-fix `a63f54b` fails with `JSON.parse` on the unescaped hook payload.
- `git diff --check HEAD~1..HEAD`
- LF-normalized `HEAD` fixture with temporary `jq` shim: `bash -n hooks/session-start.sh && bash -n hooks/session-start-test.sh && bash hooks/session-start-test.sh`
- LF-normalized `HEAD` fixture without `jq` in `PATH`: `bash hooks/session-start-test.sh`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,6 +47,35 @@ New skills should generally follow the standard anatomy:
 - Preserve the existing structure and tone
 - Test that YAML frontmatter remains valid after edits
 
+## Testing Hooks
+
+The session-start hook (`hooks/session-start.sh`) injects the `using-agent-skills` meta-skill into every new Claude Code session. A regression test at `hooks/session-start-test.sh` validates the hook's JSON payload — both when `jq` is available and when it isn't.
+
+Run it before opening any PR that touches:
+
+- `hooks/session-start.sh`
+- `skills/using-agent-skills/SKILL.md` (the meta-skill content embedded by the hook)
+
+```bash
+bash hooks/session-start-test.sh
+```
+
+Expected output: `session-start JSON payload OK`. The script exits non-zero on any assertion failure.
+
+### Reproducing the no-jq fallback
+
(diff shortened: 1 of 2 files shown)


AcererakTheMighty:
Looks good overall. I like that it keeps compatibility with the INFO fallback when jq isn’t available—that feels like a solid safety net.

federicobartoli:
Verified the test passes against `upstream/main` (which uses jq). Solid regression coverage for #89 — `set -euo pipefail`, `mktemp` + `trap`, both code paths exercised. A few nits:

1. **Node.js dependency for assertions.** The test uses `node` only for `JSON.parse` + `String.includes`. Since the hook itself relies on bash + jq, doing the assertions in jq would avoid pulling in a Node toolchain just for tests:
   ```bash
   echo "$payload" | jq -e '.priority == "IMPORTANT" and (.message | contains("agent-skills loaded."))' > /dev/null
   ```
2. **Not wired into CI.** Without a GitHub Actions workflow invoking this, regression coverage only exists when contributors manually run it. Worth either adding a workflow or documenting how/when to run it in CONTRIBUTING.
3. **No-jq path is hard to reproduce.** The verification mentions running "without jq in PATH" but doesn't show how to stub it. Could be useful to document (e.g. `PATH=$(pwd)/empty bash hooks/session-start-test.sh`).

addyosmani:
Thanks for the changes @Bortlesboat and the review, @federicobartoli. I'll tackle 1 and perhaps 2 in edit before landing this.

PR #128, opened 2026-05-05
Fix formatting in README workflow diagram

kigiri:
This was bothering me

```
  DEFINE         PLAN          BUILD        VERIFY        REVIEW         SHIP
 ┌──────┐      ┌──────┐      ┌──────┐      ┌──────┐      ┌──────┐      ┌──────┐
 │ Idea │ ──▶ │ Spec │ ──▶ │ Code │ ──▶ │ Test │ ──▶ │  QA  │ ──▶ │  Go  │
 │Refine│      │  PRD │      │ Impl │      │Debug │      │ Gate │      │ Live │
 └──────┘      └──────┘      └──────┘      └──────┘      └──────┘      └──────┘
  /spec         /plan         /build        /test         /review       /ship
```

Current version:
```
  DEFINE          PLAN           BUILD          VERIFY         REVIEW          SHIP
 ┌──────┐      ┌──────┐      ┌──────┐      ┌──────┐      ┌──────┐      ┌──────┐
 │ Idea │ ───▶ │ Spec │ ───▶ │ Code │ ───▶ │ Test │ ───▶ │  QA  │ ───▶ │  Go  │
 │Refine│      │  PRD │      │ Impl │      │Debug │      │ Gate │      │ Live │
 └──────┘      └──────┘      └──────┘      └──────┘      └──────┘      └──────┘
  /spec          /plan          /build        /test         /review       /ship
```

PR #140, opened 2026-05-07
Update workflow diagram formatting in README

Lovish-Mehra-06:
fix: correct ASCII diagram alignment

Fixed misalignment in the ASCII workflow diagram by replacing inconsistent spacing and improving terminal-safe formatting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,12 +5,12 @@
 Skills encode the workflows, quality gates, and best practices that senior engineers use when building software. These ones are packaged so AI agents follow them consistently across every phase of development.
 
 ```
-  DEFINE          PLAN           BUILD          VERIFY         REVIEW          SHIP
- ┌──────┐      ┌──────┐      ┌──────┐      ┌──────┐      ┌──────┐      ┌──────┐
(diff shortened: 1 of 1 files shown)


federicobartoli:
Thanks for the contribution! A couple of concerns:

1. **Regression Unicode → ASCII**: the original used `───▶` (Unicode arrow) which renders cleanly on GitHub markdown. Replacing it with `--->` is an aesthetic downgrade — the "terminal-safe" framing in the PR description doesn't really apply since this lives in a README rendered by GitHub, not in a terminal.
2. **Wider boxes don't improve alignment**: the original labels (DEFINE, PLAN, etc.) were already approximately centered over the boxes; the new version has the same approximate centering with wider boxes, so the underlying issue isn't actually addressed.
3. **Overlap with #128**, which makes a more conservative pass on the same diagram (keeps Unicode arrows, just trims one `─` and tightens label spacing).

Would suggest closing in favor of #128, or rebasing this to keep Unicode arrows if there's a specific alignment fix that #128 doesn't cover.

Lovish-Mehra-06:
Thank you for the detailed feedback. I'll close this PR in favor of that approach. Appreciate the review.

PR #142, opened 2026-05-07
docs: list using-agent-skills in README

aqilaziz:
## Summary
- Update the README skill count from 20 to 21
- Add the using-agent-skills meta-skill to the skill list

## Verification
- Counted 21 directories under skills/
- Checked README local Markdown links resolve
- Ran git diff --check

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ Skills are plain Markdown - they work with any agent that accepts system prompts
 
 ## All 21 Skills
 
-The pack includes these 21 skills — each one a structured workflow with steps, verification gates, and anti-rationalization tables. You can also reference any skill directly.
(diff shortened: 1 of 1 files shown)


federicobartoli:
Verified the count is correct (21 skills under `skills/` on main, including `using-agent-skills/`). Placement of the new "Meta" section before "Define" makes sense since it's about routing into the others.

One optional nit: the previous wording tied the skill list back to the slash commands ("the commands above are the entry points... they activate these skills"). The new sentence drops that framing. Could keep both — e.g. *"The commands above are entry points. The pack includes 21 skills total — each a structured workflow..."* — but minor, not a blocker.

addyosmani:
Thanks for the review, @federicobartoli and agreed! I'm happy to make a minor edit for your optional nit too before landing.

PR #143, opened 2026-05-07
Add six Python simplification idioms to code-simplification skill

milasaurus:
Extends the Python language-specific section with classic behavior-preserving idiom upgrades:

- enumerate() over manual index counter
- zip() over parallel index iteration
- with-statement over try/finally for resource cleanup
- dict.get(default) over membership-then-lookup
- collections.Counter over manual default-init dict
- any() over loop-with-flag boolean reduction

Each follows the existing "# SIMPLIFY: ... # Before / # After" style already used by the Python and TypeScript subsections.

### DIFF
--- a/skills/code-simplification/SKILL.md
+++ b/skills/code-simplification/SKILL.md
@@ -268,6 +268,65 @@ def process(data):
     if not data.has_permission():
         raise PermissionError("No permission")
     return do_work(data)
+
+# SIMPLIFY: Manual index counter
+# Before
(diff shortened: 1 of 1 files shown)


milasaurus:
Closing — on reflection, this drifts the skill toward a Python idiom cheatsheet, and most of the additions cover patterns modern coding agents already produce unprompted. The existing two Python examples cover the cases agents actually flub. Sorry for the noise!

PR #151, opened 2026-05-09
fix: resolve #150 — Recommended way to handle V2 features without polluting V1 spec?

### LINKED ISSUE #150: Recommended way to handle V2 features without polluting V1 spec?
During /spec I have lots of ideas, many of them I need to leave behind for the V1. 
1. Should I mention in the spec that this is a V1, even mention the V2 backlog?
2. After /build and possibly the others including /ship, what is the process to work on the V2 features? Hand-edit spec.md with a following /plan? Lets say I built ontop of Sqlite and now want to move to PostGreSQL. Or I want to remove a manual user step and use OCR instead.

I understand that its probalby best to bring in single new requirements or changes and work on them instead of changing too many at once. Still I am trying to figure out the questions above... Thanks so much for this awesome plugin!

andoan16:
## Summary

fix: resolve #150 — Recommended way to handle V2 features without polluting V1 spec?

## Problem

**Severity**: `Medium` | **File**: `.claude/commands/spec.md`

The spec.md file should include guidance on how to handle version progression from V1 to V2 while maintaining clean separation between versions. This addresses the user's concern about not polluting V1 specs with V2 features.

## Solution

Add a section to spec.md that explains: 1) How to clearly mark specifications as V1, 2) Recommendations for maintaining a V2 backlog or roadmap, 3) Process for evolving from V1 to V2 including technology migrations (SQLite to PostgreSQL example), and 4) Best practices for incremental feature development.

## Changes

- `.claude/commands/spec.md` (modified)
- `docs/getting-started.md` (modified)
- `skills/deprecation-and-migration/SKILL.md` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/.claude/commands/spec.md
+++ b/.claude/commands/spec.md
@@ -13,3 +13,12 @@ Begin by understanding what the user wants to build. Ask clarifying questions ab
 Then generate a structured spec covering all six core areas: objective, commands, project structure, code style, testing strategy, and boundaries.
 
 Save the spec as SPEC.md in the project root and confirm with the user before proceeding.
+
+## Version Management
+
+When evolving specifications from V1 to V2:
+
+1. **Version Labeling**: Clearly mark specifications with version identifiers (V1, V2) in the document title and filename
+2. **Separation**: Maintain distinct specification documents for each major version to avoid feature pollution
+3. **Migration Planning**: Document technology migrations (e.g., SQLite to PostgreSQL) and architectural changes in a separate V2 backlog
+4. **Incremental Development**: Add new features incrementally with clear version tracking rather than mixing versions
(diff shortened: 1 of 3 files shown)


addyosmani:
Same pattern as #157  with respect to issues I'm seeing - splits the word "delegates" into "deleg" / "ates" and injects a markdown ## Version Boundary Management heading inside a TypeScript code comment in the deprecation skill.

I have also not personally discussed any specific version numbers outside of the published scheme I've been publishing for the last few months so I have no idea why you're referring to a V1 and V2 in this way.

PR #154, opened 2026-05-09
fix: resolve #122 — /build不能让agent触发git

### LINKED ISSUE #122: /build不能让agent触发git
希望作者可以让agent始终触发git-worktree-flow.. 技能，因为目前来看，在编码阶段我还得主动告诉它才启动git工作流，git工作流不应该是在ship阶段，而是在构建或者叫编码阶段

andoan16:
## Summary

fix: resolve #122 — /build不能让agent触发git

## Problem

**Severity**: `High` | **File**: `.claude/commands/build.md`

The build command should automatically trigger the git worktree flow instead of requiring manual activation. This addresses issue #122 where users want git workflow to be part of the build/encoding stage rather than just the ship stage.

## Solution

Modify the build.md file to include instructions or automation that triggers the git-worktree-flow skill automatically when the build command is executed. This could involve adding a call to initialize git workflow or ensuring the skill is activated as part of the build process.

## Changes

- `.claude/commands/build.md` (modified)
- `.gemini/commands/build.toml` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/.claude/commands/build.md
+++ b/.claude/commands/build.md
@@ -11,7 +11,7 @@ Pick the next pending task from the plan. For each task:
 3. Write a failing test for the expected behavior (RED)
 4. Implement the minimum code to pass the test (GREEN)
 5. Run the full test suite to check for regressions
-6. Run the build to verify compilation
+6. Run the build to verify compilation and automatically trigger the git workflow
 7. Commit with a descriptive message
 8. Mark the task complete and move to the next one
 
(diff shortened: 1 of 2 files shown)


addyosmani (changes requested):
The commit step is already here, but making /build automatically push to trigger CI isn't something I want it doing on its own; pushing should stay deliberate. The .claude and .gemini command files also drift apart in this change. Happy to take a version that stops at commit, or puts the push behind an explicit opt-in.

Issue #32, opened 2026-04-07
Can't install in Claude Code


Gnex77:
Followed the getting started instructions. I added the marketplace but when I try to install the skill I get the following error. 

```bash
Error: Failed to install: Failed to clone repository: Cloning into '/Users/guario/.claude/plugins/cache/temp_github_1775568069794_xfvinv'...
  git@github.com: Permission denied (publickey).                    
  fatal: Could not read from remote repository.     
                  
  Please make sure you have the correct access rights
  and the repository exists.
```

Marketplace config:
```bash
  Manage marketplaces  

  ❯ + Add Marketplace

    ● addy-agent-skills
      addyosmani/agent-skills
      1 available • Updated 4/7/2026

    ● ✻ claude-plugins-official ✻
      anthropics/claude-plugins-official
      123 available • 11 installed • Updated 4/7/2026
```

Issue #77, opened 2026-04-16
Plugin install fails inside Docker/devcontainer due to SSH authentication


Ic3b3rg:
## Description

`claude plugin install` fails when running inside a Docker container (e.g., VS Code devcontainer) because the git clone uses SSH (`git@github.com:`) and the container typically doesn't have SSH keys configured.

## Steps to reproduce

1. Open a project inside a devcontainer (Docker)
2. Add a marketplace:
   ```bash
   claude plugin marketplace add addyosmani/agent-skills
   ```
3. Try to install a plugin:
   ```bash
   claude plugin install agent-skills@addy-agent-skills
   ```

## Expected behavior

The plugin installs successfully (as it does outside the container).

## Actual behavior

```
Installing plugin "agent-skills@addy-agent-skills"...
✘ Failed to install plugin "agent-skills@addy-agent-skills": Failed to clone repository:
Cloning into '/home/vscode/.claude/plugins/cache/temp_github_...'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
```

## Workaround

Configure git to use HTTPS instead of SSH:

```bash
git config --global url."https://github.com/".insteadOf "git@github.com:"
```

## Suggested fix

Consider using HTTPS (`https://github.com/`) as the default clone method for public marketplace repositories, or fall back to HTTPS when SSH fails. Public repositories don't require SSH authentication, so HTTPS would work without any credentials.

## Environment

- **Claude Code version:** 2.1.110
- **OS:** Linux (Docker container / devcontainer)
- **Shell:** bash


  ## Contributing                                                                                                                     
                                                                                                                                      
  I'd be happy to submit a PR for this if the team is interested. Let me know!

addyosmani:
Real bug, and it's the same root cause we just documented a fix for: the marketplace clones via SSH, which fails in containers with no keys. #368 added the git config insteadOf HTTPS-rewrite workaround to the README, which should unblock the devcontainer case. Give that a try, and if it still fails inside Docker specifically, reopen with the exact error and I'll dig in. Thanks for reporting it.

Ic3b3rg:
Hi @addyosmani, thanks for your answer. 
I tested this in a clean container with no SSH keys/agent configured:

### Without the workaround

- `claude plugin marketplace add addyosmani/agent-skills` : **succeeds**, because it already has an automatic HTTPS fallback (`SSH not configured, cloning via HTTPS`).
[...]

PR #163, opened 2026-05-10
fix(marketplace): use relative path source to bypass install bug

git-clone-abhinav:
## Summary

`/plugin install agent-skills@addy-agent-skills` fails on Claude Code **v2.1.112** with:

```
Error: Failed to install: Failed to clone repository: Cloning into '/Users/.../.claude/plugins/cache/temp_github_...'...
remote: Repository not found.
fatal: repository 'https://github.com/:addyosmani/agent-skills.git/' not found
```

Note the malformed URL: `:addyosmani` (literal `:` prefix) and trailing `/` after `.git`. Claude Code's URL template for the `{ "source": "github", "repo": "..." }` source form is not interpolating `:owner/:repo` correctly. Reproduces regardless of marketplace add transport (github shorthand, full HTTPS URL, or local path).

## Fix

Since the plugin lives in the **same repo** as the marketplace manifest, switch the plugin entry to a relative-path source:

```diff
-      "source": {
-        "source": "github",
-        "repo": "addyosmani/agent-skills"
-      },
+      "source": "./",
```

This sidesteps the buggy github source path entirely. The relative source resolves against whatever directory Claude Code clones / reads the marketplace from, so it works for every marketplace add transport (github, https, ssh, local).

Per the [official plugin marketplace docs](https://code.claude.com/docs/en/plugin-marketplaces.md#plugin-sources), relative paths are a first-class source type.

## Why this is safe

- No change to plugin contents, location, or `plugin.json`.
- Only changes how the marketplace points to the plugin (which is itself).
- Existing CI workflow `.github/workflows/test-plugin-install.yml` already runs `claude plugin marketplace add ./` followed by `claude plugin install agent-skills@addy-agent-skills`, so the relative-source path is validated on every push.

## Test plan

- [x] CI `Test plugin installation` workflow runs against this PR
- [x] Manual: `claude plugin marketplace add https://github.com/git-clone-abhinav/agent-skills.git` (branch checked out) → `claude plugin install agent-skills@addy-agent-skills` succeeds on v2.1.112
- [ ] Maintainer verification on a clean machine

## Related

- Closed issue #32 (same root cause: install fails)
- Open issue #77 covers a separate SSH/HTTPS bug that this PR does not address.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -9,10 +9,7 @@
   "plugins": [
     {
       "name": "agent-skills",
-      "source": {
-        "source": "github",
-        "repo": "addyosmani/agent-skills"
-      },
+      "source": "./",
       "description": "Production-grade engineering skills covering every phase of software development: spec, plan, build, verify, review, and ship."
     }
   ]


git-clone-abhinav:
Co-authored with [Claude Code](https://claude.com/claude-code) (Opus 4.7).

addyosmani (approved):
LGTM. Thank you

PR #164, opened 2026-05-10
feat: add interview-me skill (during the Define phase)

addyosmani:
Adds `interview-me`, a skill that sits upstream of every other Define-phase skill. Trigger: an underspecified ask, or the user explicitly invoking the skill ("interview me", "are we sure?", "stress-test my thinking").

The gap it fills: `idea-refine` generates variations from a raw idea, `spec-driven-development` writes requirements down once you know them, and `doubt-driven-development` stress-tests a plan after you've drafted one. None of those cover the moment before when the user has said "build me X" and X is the conventional answer rather than what they actually want. This skill is the structured interview that surfaces the gap before any plan, spec, or code exists.

Distinctive mechanics:
- Confidence numbers as forcing function (Step 1 commits to 0-100%)
- Q + GUESS format: every question carries the agent's hypothesis
- One question at a time, never batched
- Checkable stop condition: "Can I predict the user's reaction to the next three questions I would ask?"
- Six-line restate template ending in explicit yes
- Names the four bad yeses ("whatever you think", "sounds good", "sure let's go", silence) with specific re-prompts

Registers the skill in:
- CLAUDE.md (Define phase, also adding idea-refine which was missing)
- README.md (skills table, count, project tree)
- skills/using-agent-skills/SKILL.md (flowchart, lifecycle sequence, quick reference table)

cc @federicobartoli @nucliweb for their thoughts :)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -15,7 +15,7 @@ docs/         → Setup guides for different tools
 
 ## Skills by Phase
 
-**Define:** spec-driven-development
+**Define:** interview-me, idea-refine, spec-driven-development
 **Plan:** planning-and-task-breakdown
 **Build:** incremental-implementation, test-driven-development, context-engineering, source-driven-development, doubt-driven-development, frontend-ui-engineering, api-and-interface-design
 **Verify:** browser-testing-with-devtools, debugging-and-error-recovery
--- a/README.md
+++ b/README.md
@@ -125,9 +125,9 @@ Skills are plain Markdown - they work with any agent that accepts system prompts
 
 ---
 
-## All 22 Skills
+## All 23 Skills
 
-The commands above are entry points. The pack includes 22 skills total — 21 lifecycle skills plus the `using-agent-skills` meta-skill. Each skill is a structured workflow with steps, verification gates, and anti-rationalization tables. You can also reference any skill directly.
+The commands above are entry points. The pack includes 23 skills total — 22 lifecycle skills plus the `using-agent-skills` meta-skill. Each skill is a structured workflow with steps, verification gates, and anti-rationalization tables. You can also reference any skill directly.
 
 ### Meta - Discover which skill applies
 
@@ -139,6 +139,7 @@ The commands above are entry points. The pack includes 22 skills total — 21 li
 
 | Skill | What It Does | Use When |
 |-------|-------------|----------|
+| [interview-me](skills/interview-me/SKILL.md) | One-question-at-a-time interview that extracts what the user actually wants instead of what they think they should want, until ~95% confidence | The ask is underspecified, or the user invokes "interview me" / "grill me" |
 | [idea-refine](skills/idea-refine/SKILL.md) | Structured divergent/convergent thinking to turn vague ideas into concrete proposals | You have a rough concept that needs exploration |
 | [spec-driven-development](skills/spec-driven-development/SKILL.md) | Write a PRD covering objectives, commands, structure, code style, testing, and boundaries before any code | Starting a new project, feature, or significant change |
 
@@ -248,7 +249,8 @@ Every skill follows a consistent anatomy:
 
 ```
 agent-skills/
-├── skills/                            # 22 skills (21 lifecycle + 1 meta)
+├── skills/                            # 23 skills (22 lifecycle + 1 meta)
+│   ├── interview-me/                  #   Define
 │   ├── idea-refine/                   #   Define
 │   ├── spec-driven-development/       #   Define
 │   ├── planning-and-task-breakdown/   #   Plan
--- a/skills/interview-me/SKILL.md
+++ b/skills/interview-me/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: interview-me
-description: Extracts what the user actually wants instead of what they think they should want. Achieves this through one-question-at-a-time interview until ~95% confidence about the underlying intent. Use when an ask is underspecified ("build me X" without "for whom" or "why now"), when the user explicitly invokes ("interview me", "are we sure?", "stress-test my thinking"), or when you catch yourself silently filling in ambiguous requirements before any plan, spec, or code exists.
+description: Extracts what the user actually wants instead of what they think they should want. Achieves this through one-question-at-a-time interview until ~95% confidence about the underlying intent. Use when an ask is underspecified ("build me X" without "for whom" or "why now"), when the user explicitly invokes ("interview me", "grill me", "are we sure?", "stress-test my thinking"), or when you catch yourself silently filling in ambiguous requirements before any plan, spec, or code exists.
 ---
 
 # Interview Me
@@ -21,7 +21,7 @@ Apply this skill when:
 - The request is conventional rather than specific ("build me X", "make it faster") and you can't unpack the convention without guessing
 - You're tempted to start with assumptions you haven't surfaced
 - The user hasn't said which value they're optimizing for when two reasonable ones are in tension (simplicity vs. flexibility, cost vs. speed)
-- The user explicitly invokes: "interview me", "before we start, are we sure?", "stress-test my thinking"
+- The user explicitly invokes: "interview me", "grill me", "before we start, are we sure?", "stress-test my thinking"
 
 **When NOT to use:**
 
@@ -135,31 +135,6 @@ The output of this skill is a **confirmed statement of intent**: the restate fro
 
 If the user wants the intent to persist (a multi-session project, a handoff to another collaborator), offer to save it to `docs/intent/[topic].md`. Only save if they confirm.
 
-## Common Rationalizations
-
-| Rationalization | Reality |
-|---|---|
-| "The ask is clear enough" | If you can't write the user's desired outcome in one sentence right now, the ask isn't clear. Run Step 1 before deciding. |
-| "Asking too many questions wastes their time" | Time wasted by 4–6 targeted questions is small. Time wasted by building the wrong thing is enormous, and the user is the one bearing that cost. |
-| "I'll figure it out as I build" | Switching costs after code exists are 10x what they are now. Discovery during implementation is rework. |
-| "They said 'whatever you think,' so I should just decide" | "Whatever you think" is delegation, not decision. Re-ask with two concrete options as a choice. |
-| "I should give them several options to pick from" | Options work when the user knows what they want and is choosing between trade-offs. They don't know what they want yet. Listing options widens the search; asking narrows it. |
-| "If I attach my guess, I'm leading them" | Leading is the point. Reacting is faster than generating from scratch. The risk is sycophancy, not leading; mitigate by being visibly willing to be wrong. |
-| "We've talked enough, I get it" | Test it: can you predict their reaction to the next three questions? If not, you don't get it yet. |
-| "The user said yes, we're done" | If the yes followed a vague restate or an open-ended "sounds good," the yes is hollow. Restate concretely and re-confirm. |
-
-## Red Flags
-
-- Three or more questions in a single message: that's batching, not interviewing
-- A question without your hypothesis attached: that's surveying, not committing
-- Accepting "whatever you think is best" as a terminal answer
-- Producing a spec, plan, or task list before the user has explicitly confirmed your restate
-- Questions framed as "what would be best practice?" instead of "what do you actually want?"
-- The user gives a sophistication-signaling answer ("scalable", "clean", "modern") and you accept it without probing whether it's what they actually want
-- Three or more rounds without your confidence visibly rising: you're asking the wrong questions, step back and reframe
-- Saving the intent doc before the user has confirmed (the doc itself implies a yes the user didn't give)
-- Skipping the "Out of scope" line in the restate (silent disagreement about non-goals is half of misalignment)
-
 ## Example
 
 A short before-and-after.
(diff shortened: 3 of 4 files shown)


federicobartoli:
Thanks for looping me in, @addyosmani. I really, really like this.

I did a quick pass against Anthropic’s Skill authoring guidance and nothing jumped out to me: the frontmatter looks valid, the description is clear and specific, and it’s comfortably within the recommended limits.

More importantly, I love the idea. It feels very architectural: don’t jump straight into solution mode, first challenge the brief a little and make sure we understand the real intent behind it.

I want to test this tomorrow in a few of my projects. Very cool.

Strong +1 from me.

addyosmani:
Thanks so much for taking a read through, @federicobartoli! Glad to hear it resonates and looking forward to any thoughts you have after trying it out. Appreciate the review!

nucliweb:
Great addition, the skill fills a real gap in the Define phase. The example (dashboard → "I don't have a list") is the strongest part: it makes the abstract process immediately concrete. The "four bad yeses" taxonomy and the 95% confidence stop condition are sharp and checkable.

One finding worth addressing before merge, plus a few minor suggestions:

Important, Flowchart overlap in `using-agent-skills/SKILL.md`

```
├── Ask is underspecified? ───────→ interview-me
├── Vague idea/need refinement? ──→ idea-refine
```

These two branches describe the same condition in most real situations. An agent parsing the flowchart top-to-bottom will match interview-me first and never reach idea-refine. The real distinction (well-explained in the skill's prose) is:

- interview-me → "don't know what I want yet"
- idea-refine → "have a rough concept, need variations"

Suggested fix:

```
├── Don't know what you want yet? ─────────→ interview-me
├── Have a rough concept, need variants? ───→ idea-refine
```

### Suggestions

**"grill me"** inconsistency, The README trigger column says the user invokes "interview me" / "grill me", but SKILL.md's When to Use only lists "interview me", "before we start, are we sure?", "stress-test my thinking". Add "grill me" to SKILL.md's explicit triggers or remove it from the README cell.

Section order, The Example section is placed after Red Flags. It's the skill's most persuasive element and would land better before Common Rationalizations, so the reader sees the payoff before hitting the rationalization table.

Interaction with Other Skills, Placed after Red Flags, which pushes Verification to feel like a footnote. Consider moving it before Common Rationalizations.

---

The Loading Constraints section (non-interactive context guard) is a smart addition worth keeping. The skill is a clean, mergeable contribution, the flowchart fix is a one-liner.

nucliweb:
## Eval results, interview-me skill (skill-creator iteration 1)

To run this eval, we used the `/skill-creator` skill from Claude Code: we took the proposed SKILL.md content from this PR, created the skill locally, and ran the skill-creator eval workflow against it to measure how the skill changes Claude's behavior compared to the baseline (no skill).

Ran 4 evals: 3 should-trigger cases + 1 control (mechanical operation that should not trigger). Each eval ran with and without the skill to measure the delta.

Results

| Eval | Prompt | with_skill | without_skill |
|---|---|---|---|
| 0 — dashboard-underspecified | "I need to build a dashboard for my team to track progress. Can you help me get started?" | 5/5 ✅ | 0/5 ❌ |
| 1 — make-it-faster-vague | "Our app feels slow. Can you help me make it faster?" | 5/5 ✅ | 0/5 ❌ |
| 2 — explicit-interview-me-invocation | "Before we start building the new onboarding flow, interview me, I want to make sure we're building the right thing." | 5/5 ✅ | 1/5 ⚠️ |
| 3 — mechanical-rename (control) | "Rename the function getUserData to fetchUserProfile across all files in the project." | 4/4 ✅ | 4/4 ✅ |

Overall pass rate: with_skill 100% vs without_skill 30%

Assertions checked (evals 0–2)

Each should-trigger eval verified: (1) HYPOTHESIS statement present before any question, (2) CONFIDENCE number stated, (3) GUESS attached to the question, (4) exactly one question asked, (5) no jumping to solutions/tools before understanding the problem.

The control eval (3) verified the inverse: no HYPOTHESIS, no CONFIDENCE, no GUESS, response addresses the task directly.

### Key findings

Evals 0 & 1 show maximum differentiation. Without the skill, Claude goes straight to React/Chart.js/Notion/Lighthouse and performance tips before understanding the problem, classic solution-first anchoring. With the skill, it commits to a hypothesis, states a confidence number, and asks exactly one question with a
guess attached.

Eval 2 reveals something useful about the skill's value. Without the skill, Claude intuits that some discovery is needed ("Great idea, a quick discovery conversation before building") but without the structural scaffold. It asks in the right direction but batches sub-questions and doesn't commit to a GUESS, so the user can't react to a hypothesis, they can only answer open-ended questions. The skill adds the forcing function that makes the interview actually work.

Eval 3 (control) passes 100% in both configurations, confirming the skill correctly scopes itself and doesn't over-trigger on mechanical operations.

### Cost

~5s and ~1,940 tokens added per run, minimal overhead for the behavioral shift produced.

### Verdict

The skill does what it says. The HYPOTHESIS + CONFIDENCE + Q + GUESS structure is the differentiator, it's not just "ask before building" (Claude already tends to do that in some cases), it's "commit to a hypothesis so the user can react rather than generate from scratch." Looks good to merge.

nucliweb:
## Eval results, interview-me skill (skill-creator iteration 2)

Re-ran the `/skill-creator` eval workflow against the skill with all changes from the review comment applied: flowchart label disambiguation in using-agent-skills, "grill me" added as an explicit trigger to SKILL.md and the frontmatter description, and section reordering (Example and Interaction with Other Skills moved before Common Rationalizations).

Added one new eval (4) to specifically validate the "grill me" trigger. 5 evals total: 4 should-trigger + 1 control. Each ran with and without the skill.

### Results

| Eval | Prompt | with_skill | without_skill |
|---|---|---|---|
| 0 — dashboard-underspecified | "I need to build a dashboard for my team to track progress. Can you help me get started?" | 5/5 ✅ | 0/5 ❌ |
| 1 — make-it-faster-vague | "Our app feels slow. Can you help me make it faster?" | 5/5 ✅ | 0/5 ❌ |
| 2 — explicit-invocation ("interview me") | "Before we start building the new onboarding flow, interview me, I want to make sure we're building the right thing." | 5/5 ✅ | 2/5 ⚠️ |
| 3 — mechanical-rename (control) | "Rename the function getUserData to fetchUserProfile across all files in the project." | 4/4 ✅ | 4/4 ✅ |
| 4 — grill-me-explicit-trigger (new) | "grill me on this before we start — I want to build a new notification system for our app" | 5/5 ✅ | 1/5 ❌ |

Overall pass rate: with_skill 100% vs without_skill 32%

Assertions checked (evals 0–2, 4)

Same five as iteration 1: (1) HYPOTHESIS statement present before any question, (2) CONFIDENCE number stated, (3) GUESS attached to the question, (4) exactly one question asked, (5) no jumping to solutions/tools before understanding the problem.

Eval 4 verified the same set with "grill me" as the trigger phrase.

The control eval (3) verified the inverse: no HYPOTHESIS, no CONFIDENCE, no GUESS, task addressed directly.

### Key findings

Eval 4 validates the "grill me" trigger. With the skill: HYPOTHESIS: ~25%, one question (Who receives these notifications — end users or internal?), GUESS attached with reasoning. Without the skill: Claude recognises it's being asked to probe before building, but produces a 10-question dump — channels, latency, delivery guarantees, deduplication, read/unread state — all at once, none with a hypothesis attached. The same asymmetry as evals 0 and 1: the problem without the skill isn't silence, it's commitment-free surveying.

Eval 2 without_skill moves from 1/5 to 2/5. Claude's natural response to an explicit "interview me" prompt now passes one question and the "begins immediately" assertions, but still misses HYPOTHESIS, CONFIDENCE, and GUESS. It gets the direction right without the skill; it gets the structure right only with it.

Eval 3 (control) passes 100% in both configurations. The new trigger didn't introduce over-triggering.

### Cost

~2.1s and ~3,706 tokens added per run vs the no-skill baseline. Slightly higher than iteration 1's ~1,940 token figure because the new eval's prompt ("grill me on this before we start...") is more elaborate and the skill processes it more thoroughly.

### Verdict

All four changes from the review apply cleanly. The skill passes 100% on an expanded eval set that now explicitly covers the "grill me" trigger. The section reordering is structural — doesn't change runtime behavior but makes the skill more inviting to read, which matters for contributor adoption. The flowchart fix in using-agent-skills is the highest-leverage change: "Don't know what you want yet?" vs "Have a rough concept, need variants?" resolves the top-to-bottom parsing ambiguity between interview-me and idea-refine. Looks good to merge.

federicobartoli:
I’ve been trying `interview-me` in a couple of local/project-like workflows today, and my first impression is that it works well for the exact gap this PR is targeting: stopping the agent before it turns an underspecified ask into a plan/spec/code too early.

Building on @nucliweb’s eval comments above, I also tried a slightly different experiment: wiring this skill into a repo-level eval harness I’ve been working on locally.

I haven’t committed the `evals/` folder yet because I’m still not fully convinced about the shape of it, but this PR felt like a good concrete case to test it against.

The experimental harness is structured roughly like this:

- `evals/tasks/*.jsonl` contains task prompts per skill
- `evals/src/runner.ts` runs each task in two conditions:
  - baseline: no plugin / no skill available
  - auto-activated: local plugin loaded, model decides whether to invoke a skill
- `evals/src/judge.ts` uses an LLM judge to score the response
- `evals/src/report.ts` summarizes activation rate, pass-rate delta, red flags, and cost
- `evals/src/skills.ts` parses each skill’s `## Verification` checklist and `## Red Flags`

For this run I used:

- Agent model: `claude-haiku-4-5-20251001`
- Judge model: `claude-sonnet-4-6`
- 7 prompts
- 2 conditions
- 3 runs
- 42 agent calls + 42 judge calls

The prompts were intentionally focused on the gap this skill is meant to catch:

- conventional artifacts that may be the wrong artifact: dashboard, mobile app, CRM integration
- vague success criteria: “make it faster”
- explicit invocation: “interview me”, “are we sure?”, “stress-test my thinking”
- sophistication-signaling language: “robust”, “scalable”, “clean architecture”, “I should probably add AI…”

Because `interview-me` is inherently interactive, I added task-level first-turn evaluation criteria instead of scoring only the full skill Verification checklist. Otherwise a one-turn eval would unfairly fail later-stage criteria like “the user confirmed the restate with an explicit yes”.

Results from this local run:

```text
Activation rate:              14.3%
Pass rate baseline:            2.4% ± 2.4
Pass rate auto-activated:     13.5% ± 1.4
Delta:                       +11.1%
Statistical separation:        z≈5.7
When interview-me fired:      83.3%
When it did not fire:          1.9%
Red flags triggered:          39/42 runs
```

My read: this is a useful signal, but still preliminary.

The most interesting thing the eval surfaced is that when `interview-me` fires, the behavior looks strong. But in most implicit cases, the model routed to adjacent skills like `idea-refine`, `performance-optimization`, or `spec-driven-development` instead. That means the skill content seems promising, while the activation/routing boundary is probably the part worth sharpening and re-testing.

That also lines up with the earlier review comment about distinguishing:

- `interview-me`: “I don’t know what I actually want yet”
- `idea-refine`: “I have a rough concept and need variants”

One caveat: this was a local preliminary run against my in-progress eval setup. I want to rerun after making sure the local harness includes the PR’s discovery/routing wording as well, so I would not over-interpret the activation number yet.

Still, this already feels useful as a regression signal: does the agent pause and interview before locking onto the user’s first proposed solution as if it were already the real requirement?

federicobartoli:
For extra context, a few representative prompts from the local eval set:

```text
“I need to build a dashboard for my team to track progress. We have a bunch of experiments running and leadership keeps asking how things are going. Can you help me get started?”

“Our app feels slow and users have started complaining. I want to make it faster before the next launch. What should we do?”

“Stress-test my thinking: I should probably add AI summaries to our support inbox because everyone expects AI features now, but I'm not totally sure this is the right move.”

“I think we need a CRM integration. Sales says account context is scattered across docs, Slack, and spreadsheets, and I want to fix it this quarter.”

I intentionally avoided toy prompts and tried to make them sound like real project asks where the proposed artifact may be a proxy for a more basic need.

**Comment by federicobartoli** [comment]
I ran a second experiment focused specifically on wording/routing.

The first eval suggested that `interview-me` behaves well when it fires, but activation was the weak point. So I ran a small wording matrix to test which discovery/frontmatter language makes the model choose `interview-me` instead of adjacent skills.

Same setup as before:

- Agent model: `claude-haiku-4-5-20251001`
- Judge model: `claude-sonnet-4-6`
- 7 prompts
- 3 runs
- auto-activated condition only
- 6 wording variants

The variants were:

- `current-local`: current local wording
- `pr-flowchart`: adds the explicit `Ask is underspecified? → interview-me` discovery branch
- `disambiguated-flowchart`: separates “don’t know what the user wants yet” from “have a rough concept, need variants”
- `solution-proxy-description`: emphasizes solution-first/proxy artifacts like dashboard, CRM, mobile app, AI feature
- `first-turn-format-description`: emphasizes the first-turn output contract: `HYPOTHESIS`, `CONFIDENCE`, one `Q`, `GUESS`
- `competitor-guards`: strengthens `interview-me` and adds “use interview-me first” guardrails to adjacent skill descriptions (`idea-refine`, `spec-driven-development`, `performance-optimization`)

Results:

```text
Variant                                      Activation      Pass    When fired   Red flags
-------------------------------------------------------------------------------------------
competitor-guards                                85.7%     84.1%         95.4%        6/21
current-local                                    14.3%     14.3%         83.3%       18/21
disambiguated-flowchart                          61.9%     57.9%         93.6%       11/21
first-turn-format-description                    71.4%     63.5%         87.8%       10/21
pr-flowchart                                     76.2%     71.4%         91.7%        9/21
solution-proxy-description                       71.4%     68.3%         95.6%        7/21

```

My read: the skill content looks strong when invoked. The larger issue is the routing boundary.

The best-performing variant was competitor-guards, which made the boundary explicit in both directions: stronger interview-me wording, plus guardrails on adjacent skills like idea-refine, spec-driven-development, and performance-optimization.

That suggests the most useful wording is around:

missing who / why / success / constraint
solution-first or proxy artifacts like dashboards, CRM integrations, mobile apps, AI features
using interview-me before idea refinement, specs, performance work, or implementation when the real intent is still unknown
A smaller change also performed well: simply adding an explicit discovery branch like Ask is underspecified? → interview-me moved activation from 14.3% to 76.2% in this local matrix.

One remaining weak spot: the “make it faster” prompt still struggled to route to interview-me. That may need especially explicit wording, e.g. “if the user says the app feels slow but does not name the user flow, metric, or success threshold, interview before performance optimization.”

Still preliminary, but this feels like a useful direction: interview-me works best when the surrounding skill system clearly says “do not refine/spec/optimize yet; first clarify what the user actually wants.

P.S. I can share the local JSON outputs and prompt set if useful; I kept the comment summarized for readability since the raw outputs are fairly large.

addyosmani:
Thanks @nucliweb and @federicobartoli - this is by far the most thoroughly reviewed PR I've shipped here, and the eval data from both of you is genuinely useful (and honestly, novel for this repo). A few separate threads worth replying to:

@nucliweb - picking up your three concrete fixes for this PR:

1. Flowchart disambiguation in using-agent-skills/SKILL.md - relabel as "Don't know what you want yet?" → interview-me vs "Have a rough concept, need variants?" → idea-refine so the top-to-bottom parse can't match both
2. Add "grill me" as an explicit trigger in interview-me/SKILL.md's When to Use list and the frontmatter description (the README mentioned it but SKILL.md didn't - that's on me)
3. Move Example and Interaction with Other Skills above Common Rationalizations

Your iteration-2 eval result - same 100% pass rate after the reorder, plus the new "grill me" eval at 5/5 - is exactly the kind of evidence that makes a structural edit safe to apply. Thanks for re-running it.

@federicobartoli - the wording matrix is super actionable. Going from 14.3% → 85.7% activation with competitor-guards (and the related observation that the skill behaves well when it fires - the issue is upstream routing, not the skill itself) is too big to fold into this PR. It touches idea-refine, spec-driven-development, and performance-optimization descriptions, plus the special-case wording for "feels slow" prompts that don't name a flow/metric/threshold. Each of those edits should be validated independently, not bundled.

So my plan:

1. Land this PR with @nucliweb's three small fixes - the skill itself is sound, both eval runs confirm that
2. Open a focused follow-up PR for the routing/competitor-guards work, validated against your wording matrix

fenixprojectteam-lgtm:
It looks really good, I hope they manage to move forward and create something incredible.

PR #175, opened 2026-05-12
docs(skills): add WHEN: trigger phrases to 20 skill descriptions

spboyer:
## Summary

Adds an explicit `WHEN: "phrase1", "phrase2", ...` clause to the description of **20 of the 22 skills**, preserving every existing `Use when...` sentence unchanged. The other 2 skills (`idea-refine`, `browser-testing-with-devtools`) are handled by #174 and were skipped here to avoid merge conflicts — happy to follow up if both PRs land.

This is **purely additive**: no skill body content was touched, no `Use when...` prose was removed, no behavior changes. Just richer machine-readable trigger metadata.

## Why this matters

The frontmatter `description` is the only thing an agent sees when deciding whether to invoke a skill. `AGENTS.md` line 109 already calls out the desired format:

> *"description: {One sentence describing when to use this skill. Include trigger phrases like `\"Deploy my app\"`, `\"Check logs\"`, etc.}"*

But the actual skills in the repo only used prose like `Use when implementing any feature...`. That works for full-sentence semantic matching, but loses signal for agents and orchestrators that route on **short, distinctive keyword tokens**. Adding a `WHEN:` clause with 4–5 quoted phrases per skill closes that gap without changing how a human or LLM reads the description.

Concretely, before this PR the skills would only fire on agents that semantically parse "Use when..." clauses. After this PR they also fire on direct intent keywords like:

- `"create a branch"` → `git-workflow-and-versioning`
- `"Core Web Vitals"` → `performance-optimization`
- `"write an ADR"` → `documentation-and-adrs`
- `"red green refactor"` → `test-driven-development`
- `"adversarial review"` → `doubt-driven-development`

…and so on across all 20 touched skills.

## What changed — pattern

Every edit follows the same shape. Example (`api-and-interface-design`):

```diff
- description: Guides stable API and interface design. Use when designing APIs, module boundaries, or any public interface. Use when creating REST or GraphQL endpoints, defining type contracts between modules, or establishing boundaries between frontend and backend.
+ description: "Guides stable API and interface design. Use when designing APIs, module boundaries, or any public interface. Use when creating REST or GraphQL endpoints, defining type contracts between modules, or establishing boundaries between frontend and backend. WHEN: \"design an API\", \"REST endpoint\", \"GraphQL schema\", \"module boundary\", \"interface contract\"."
```

Two mechanical things happen on every line:

1. The bare YAML scalar is wrapped in double quotes so the inner quoted phrases are unambiguous to parsers. Inner `"` are escaped as `\"` per YAML spec.
2. A `WHEN: "...", "...", "...", "...", "..."` clause is appended at the end. Phrases were chosen to be distinctive (avoiding overlap with adjacent skills like `code-review-and-quality` vs `code-simplification`).

## Files touched (20)

`api-and-interface-design`, `ci-cd-and-automation`, `code-review-and-quality`, `code-simplification`, `context-engineering`, `debugging-and-error-recovery`, `deprecation-and-migration`, `documentation-and-adrs`, `doubt-driven-development`, `frontend-ui-engineering`, `git-workflow-and-versioning`, `incremental-implementation`, `performance-optimization`, `planning-and-task-breakdown`, `security-and-hardening`, `shipping-and-launch`, `source-driven-development`, `spec-driven-development`, `test-driven-development`, `using-agent-skills`.

## How this was driven

The audit and pattern came from [**sensei**](https://github.com/spboyer/sensei), a Ralph-loop tool for scoring and improving skill frontmatter compliance across a catalog. Sensei scored every skill in this repo and surfaced the same gap on 21 of 22: no explicit trigger keywords. This PR applies the lower-risk *additive* variant — meaning sensei's preferred `WHEN:` markers coexist with the repo's documented `Use when...` prose rather than replacing it. Two other sensei findings were intentionally **not** applied:

- **Token trimming.** Sensei's 500-token soft limit doesn't match this repo's workflow-skill style; every skill is well under the 5k hard limit and the body content is intentional. Not in scope.
- **Classification prefixes** (`**WORKFLOW SKILL**`, etc.). Cosmetic; not worth 22 files of churn.

## Testing

This repo is documentation-only (`AGENTS.md`: *"npm test — Not applicable"*), so verification is by structural checks. Verified locally with a Python YAML parser:

- ✅ All 22 SKILL.md frontmatters parse as valid YAML
- ✅ All 22 descriptions are within the 1024-character hard limit
- ✅ All 22 still begin with a third-person action verb
- ✅ All 22 still contain a `Use when...` clause (additive guarantee)
- ✅ `git diff` shows exactly 20 files, 20 insertions, 20 deletions — no body changes

### One soft tradeoff to acknowledge

Sensei recommends descriptions stay under 60 words. After this PR, a handful (`doubt-driven-development` at 62, `incremental-implementation` at 57, `planning-and-task-breakdown` at 57) sit slightly over that soft target — the natural cost of keeping the original `Use when...` prose **and** adding `WHEN:` phrases. They are all comfortably under the 1024-character hard limit. If you'd prefer I trim the `Use when...` clauses on those few to stay strictly ≤60 words, happy to follow up.

## Relationship to #174

Independent and stackable. #174 fixes the two outlier skills (`idea-refine` description was too short; `browser-testing-with-devtools` had its MCP dependency buried). This PR upgrades the trigger metadata across the rest of the catalog. Either can merge first; no conflict either way.

### DIFF
--- a/skills/api-and-interface-design/SKILL.md
+++ b/skills/api-and-interface-design/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: api-and-interface-design
-description: Guides stable API and interface design. Use when designing APIs, module boundaries, or any public interface. Use when creating REST or GraphQL endpoints, defining type contracts between modules, or establishing boundaries between frontend and backend.
+description: "Guides stable API and interface design. Use when designing APIs, module boundaries, or any public interface. Use when creating REST or GraphQL endpoints, defining type contracts between modules, or establishing boundaries between frontend and backend. WHEN: \"design an API\", \"REST endpoint\", \"GraphQL schema\", \"module boundary\", \"interface contract\"."
 ---
 
 # API and Interface Design
--- a/skills/ci-cd-and-automation/SKILL.md
+++ b/skills/ci-cd-and-automation/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: ci-cd-and-automation
-description: Automates CI/CD pipeline setup. Use when setting up or modifying build and deployment pipelines. Use when you need to automate quality gates, configure test runners in CI, or establish deployment strategies.
+description: "Automates CI/CD pipeline setup. Use when setting up or modifying build and deployment pipelines. Use when you need to automate quality gates, configure test runners in CI, or establish deployment strategies. WHEN: \"set up CI\", \"CI/CD pipeline\", \"GitHub Actions workflow\", \"quality gates\", \"deployment pipeline\"."
 ---
 
 # CI/CD and Automation
--- a/skills/code-review-and-quality/SKILL.md
+++ b/skills/code-review-and-quality/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: code-review-and-quality
-description: Conducts multi-axis code review. Use before merging any change. Use when reviewing code written by yourself, another agent, or a human. Use when you need to assess code quality across multiple dimensions before it enters the main branch.
+description: "Conducts multi-axis code review. Use before merging any change. Use when reviewing code written by yourself, another agent, or a human. Use when you need to assess code quality across multiple dimensions before it enters the main branch. WHEN: \"review this code\", \"code review\", \"before merging\", \"review my PR\", \"assess code quality\"."
 ---
 
 # Code Review and Quality
--- a/skills/code-simplification/SKILL.md
+++ b/skills/code-simplification/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: code-simplification
-description: Simplifies code for clarity. Use when refactoring code for clarity without changing behavior. Use when code works but is harder to read, maintain, or extend than it should be. Use when reviewing code that has accumulated unnecessary complexity.
+description: "Simplifies code for clarity. Use when refactoring code for clarity without changing behavior. Use when code works but is harder to read, maintain, or extend than it should be. Use when reviewing code that has accumulated unnecessary complexity. WHEN: \"simplify this\", \"refactor for clarity\", \"clean up this code\", \"reduce complexity\", \"make this readable\"."
 ---
 
 # Code Simplification
--- a/skills/context-engineering/SKILL.md
+++ b/skills/context-engineering/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: context-engineering
-description: Optimizes agent context setup. Use when starting a new session, when agent output quality degrades, when switching between tasks, or when you need to configure rules files and context for a project.
+description: "Optimizes agent context setup. Use when starting a new session, when agent output quality degrades, when switching between tasks, or when you need to configure rules files and context for a project. WHEN: \"set up context\", \"new agent session\", \"configure agent rules\", \"CLAUDE.md\", \"project conventions for agents\"."
 ---
 
 # Context Engineering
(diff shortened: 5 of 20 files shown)


chuggies510:
Some measurement that might be useful context for this PR, since it is explicitly about growing the description field.

Codex enforces a **2% of context** ceiling on the combined `description:` of every installed skill. Past it, descriptions get silently shortened:

> Skill descriptions were shortened to fit the 2% skills context budget. Codex can still see every skill, but some descriptions are shorter. Disable unused skills or plugins to leave more room for the rest.

It is a shared pool across all installed plugins, so it is not really a per-repo budget. I measured mine (18 plugins, 122 skills, 24,009 bytes of description):

| plugin | skills | bytes | share | bytes/skill |
|---|---:|---:|---:|---:|
| **agent-skills** | **24** | **6,280** | **26.2%** | **262** |
| openai-templates | 20 | 4,925 | 20.5% | 246 |
| ponytail | 6 | 2,581 | 10.8% | 430 |
| workspace-toolkit | 22 | 2,491 | 10.4% | 113 |
| meap | 13 | 1,797 | 7.5% | 138 |
| stacks | 9 | 1,382 | 5.8% | 154 |
| *(12 others)* | 28 | 4,553 | 19.0% | — |

To be fair about it: **agent-skills is the largest total mostly because it ships the most skills**, not because any one description is egregious. At 262 bytes/skill it is 1.5x the 180-byte median, and ponytail (430), template-creator (395), and documents (313) are all denser per skill. The reason it matters is leverage, not blame — it is the one place where a proportional trim frees more room than deleting most other plugins outright.

**The part that seemed worth flagging here:** this PR is +2,330 bytes net on the description field, and #434 is +764. Together that takes agent-skills from 6,280 to 9,374, a 49% increase, and from 26.2% to 34.6% of my budget. Both are well-motivated on their own terms, and I am not arguing either is wrong: #434 took the trigger rank-1 rate from 86% to 100%, which is a real, measured win, and the eval README is right that a Tier-2 failure usually means fixing the description.

The observation is just that **the objective has no cost term for length**. Every Tier-2 miss is resolved by adding vocabulary, and nothing pushes back, so the field grows monotonically. And there is a wrinkle: rank-1 is measured on full descriptions, but past the ceiling the agent receives *truncated* ones. A rate verified at 100% untruncated does not necessarily hold for what actually reaches the model.

If that seems worth addressing, the cheapest version is probably a length-aware variant of the eval you already have, rather than any editorial rule:

- report rank-1 **per byte** alongside the raw rate, so a fix that buys 1 rank-1 for 400 bytes is visibly worse than one that buys it for 40;
- or run the existing suite against descriptions truncated to a simulated budget, which measures the thing the agent actually sees.

Either turns "is this description too long?" into something CI can answer, which is more durable than a convention.

For what it is worth I drafted a trigger-only rewrite of all 24 (6,280 → 3,844, a 39% cut) before checking it against `evals/cases/`, and it dropped eval-prompt vocabulary in 17 of 24 skills. So I am not proposing it, and I would treat any length argument that has not been run through your harness with the same suspicion. The harness is the right arbiter here; it just is not currently pointed at this dimension.

Separately, and unrelated to descriptions, I filed #440 about version metadata disagreeing across the manifests.

addyosmani:
Descriptions were reworked more recently to hit 100% routing accuracy (#434), so the WHEN: phrasing here is superseded. Closing, thanks.

PR #186, opened 2026-05-20
docs: route all lifecycle skills from meta-skill

superShen0916:
## Summary
- add code-simplification to the using-agent-skills review routing path
- add deprecation-and-migration to the shipping/lifecycle routing path
- keep the lifecycle sequence and quick reference aligned with the full skill set

## Verification
- git diff --check

## Notes
The repository already documents both skills in README.md and CLAUDE.md, but the using-agent-skills meta-skill did not route users to them. This keeps the meta-skill in sync with the available lifecycle skills.

### DIFF
--- a/skills/using-agent-skills/SKILL.md
+++ b/skills/using-agent-skills/SKILL.md
@@ -30,10 +30,12 @@ Task arrives
     │   └── Browser-based? ───────────→ browser-testing-with-devtools
     ├── Something broke? ──────────────→ debugging-and-error-recovery
     ├── Reviewing code? ───────────────→ code-review-and-quality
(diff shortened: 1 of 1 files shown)


addyosmani (approved):
LGTM.

This is a clean, surgical fix to a real gap - code-simplification and deprecation-and-migration both exist as first-class skills in the repo and are listed in the README, but the using-agent-skills meta-skill was silently ignoring them, meaning agents following the dispatch tree would never get routed to either. The PR adds them in all three places they need to appear (decision tree, lifecycle sequence, and Quick Reference table) with no collateral changes, no conflicts, and correct phase placement - code-simplification under Review and deprecation-and-migration under Ship.

PR #187, opened 2026-05-20
feat(interview-me): add reason to low-confidence numbers

danielart:
## Summary

- When `CONFIDENCE` is below ~70%, the skill now requires a brief one-line reason on the same line explaining what's still unresolved or missing
- Updated the Step 1 format block and added a rule explaining the threshold and intent
- Updated the example to show the pattern in practice (both the opening turn and the mid-interview update)
- Added a matching Red Flag and Verification checklist item to enforce the rule

## Motivation

A bare percentage like `CONFIDENCE: ~30%` is a vague signal. The user knows the agent is uncertain but has no idea *why*, so they can't help close the gap. Surfacing the reason on the same line — e.g. `~30% — missing: who it's for, what "metrics" means, and what success looks like` — makes the interview collaborative: the user can immediately address the specific gap instead of waiting for the next question. I think this can help to guide the user and save time and tokens 

## Test plan

- [ ] Read the updated Step 1 format block and verify the rule is clear
- [ ] Trace through the updated example and confirm both low-confidence lines carry a reason
- [ ] Check that the new Red Flag and Verification item are consistent with the rest of the skill's style
- [ ] No other skills reference `interview-me`'s confidence format, so no cross-skill updates needed

### DIFF
--- a/skills/interview-me/SKILL.md
+++ b/skills/interview-me/SKILL.md
@@ -43,10 +43,12 @@ Before asking anything, write down your current best read of what the user wants
 
 ```
 HYPOTHESIS: You want a way to answer "how are we doing?" in standup, and "dashboard" was the convention that came to mind.
-CONFIDENCE: ~30%
+CONFIDENCE: ~30% — missing: who it's for, what "metrics" means in context, and what success looks like
 ```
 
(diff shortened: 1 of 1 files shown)


addyosmani (approved):
LGTM. Thank you!

Issue #136, opened 2026-05-07
SKILL.md files reference non-existent scripts


ether-btc:
Several SKILL.md files contain script references in their "Usage" or "How It Works" sections that do not exist in the repository.

## Example

\`spec-driven-development/SKILL.md\` references:
\`\`\`bash
/mnt/skills/user/spec-driven-development/scripts/fetch-spec.sh [args]
\`\`\`

But \`fetch-spec.sh\` does not exist in the \`scripts/\` directory.

## Proposal

1. Remove references to non-existent scripts from SKILL.md files, OR
2. Implement the missing scripts, OR
3. Document that script examples are placeholders for contributors to implement

This ensures skills are self-contained and workflows are executable.

nfgallimore:
Need a librarian agent that checks indexes and broken refs or a husky hook; or an agent hook. 🪝

federicobartoli:
Looked into this — the example given doesn't match the current repo state and I think the issue may be based on a misunderstanding of the path convention.

1. **`spec-driven-development/SKILL.md` does not reference `fetch-spec.sh`** (or any other script). I grepped the file and the entire `skills/` tree:
   ```bash
   $ grep -rn "fetch-spec" skills/
   # no matches
   ```
2. **The only script reference in any SKILL.md across the whole repo** is in `skills/idea-refine/SKILL.md:22`, which calls `/mnt/skills/user/idea-refine/scripts/idea-refine.sh` — and that file **does exist** at `skills/idea-refine/scripts/idea-refine.sh`.
3. **`/mnt/skills/user/...` is the runtime mount path**, not a repo-local path. It's documented in `AGENTS.md`:
   ```
   bash /mnt/skills/user/{skill-name}/scripts/{script}.sh [args]
   ```
   So a reference like `/mnt/skills/user/idea-refine/scripts/idea-refine.sh` resolves correctly when the skill is mounted by the Claude runtime, even though it doesn't exist at that path on the local filesystem.

Net: there's currently no broken script reference to fix here. If you saw `fetch-spec.sh` somewhere specific (different branch, fork, older version?), could you share the exact file path and line — happy to look again. Otherwise this might be a candidate to close.

The related concern — empty `scripts/` directories — is tracked separately in #135, which has the same factual issue (no empty `scripts/` dirs exist on `upstream/main` either). See [my comment there](https://github.com/addyosmani/agent-skills/issues/135#issuecomment-4410108077) for details.

nfgallimore:
[...]

PR #201, opened 2026-05-23
feat: warn on missing local skill references

superShen0916:
## Summary
- add warning-only validation for missing local file references in SKILL.md files
- ignore external URLs, anchors, absolute/runtime paths, and fenced code examples
- allow shared root-level references/docs alongside skill-local files
- add a Node-based fixture test for missing references and ignored runtime paths

Refs #136

## Verification
- node scripts/validate-skills-test.js
- git diff --check


addyosmani:
Thanks for this. skill-lint already warns on dead cross-skill references, so this is covered. Closing.

Issue #137, opened 2026-05-07
Add automated SKILL.md linting and format standardization


ether-btc:
While SKILL.md files are generally well-structured, there are minor formatting inconsistencies across skills:

- Heading levels vary (some use \`##\`, others use \`###\` for similar sections)
- List formatting varies (some use \`-\`, others use \`*\`)
- Code block indentation is inconsistent
- Frontmatter field ordering varies

## Proposal

1. Add \`markdownlint\` configuration for automated format checking
2. Standardize heading levels across all SKILL.md files
3. Ensure consistent use of \`-\` for bullet lists
4. Document formatting conventions in \`CONTRIBUTING.md\`

This improves polish and helps AI agents parse files consistently.

addyosmani:
Mostly landed since you filed this: scripts/validate-skills.js runs in CI and checks frontmatter, naming, and required sections, with more coming (a formatting-conventions doc in #206, plus the eval framework enforcing structure). Full autoformatting of the markdown body isn't there, and I'm honestly not sure it earns its tooling versus the current checks. Keeping this open for the formatting-specific piece, but the linting half is done. Thanks for the early nudge.

PR #206, opened 2026-05-26
docs: clarify skill formatting conventions

2023Anita:
## Summary

Adds concise formatting conventions for SKILL.md files in both the contributor guide and skill anatomy documentation.

## Why

This addresses part of #137 by documenting the expected formatting style before introducing any heavier linting or automated formatting. Keeping this first step documentation-only makes the convention explicit without changing existing skills or CI behavior.

## Changes

- Document frontmatter ordering expectations
- Clarify heading, list, and fenced-code-block conventions
- Encourage compact examples and avoiding placeholder sections

## Validation

- git diff --check

Refs #137

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,6 +36,17 @@ New skills should generally follow the standard anatomy:
 
 The frontmatter fields above are required. The section anatomy is a recommended pattern: equivalent headings such as `How It Works`, `Workflow`, or `Core Process` are fine when they preserve the same intent and keep the skill easy to follow.
 
+### Formatting Conventions
+
+Keep formatting predictable so both humans and agents can scan skills consistently:
+
+- Put YAML frontmatter at the top of `SKILL.md` with `name` before `description`
+- Use ATX headings (`## Heading`) for sections
+- Use `-` for unordered lists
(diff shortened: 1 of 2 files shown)


addyosmani:
Clean, small docs improvement that merges without a fight, so it's close to a yes. It addresses part of the formatting-consistency issue and puts the conventions where contributors will actually see them. Let me confirm it doesn't overlap the skill-anatomy sections we just reworked, and if it's additive I'll take it. Tidy PR, thanks.

federicobartoli:
Nice PR, this puts the conventions right where contributors will actually look for them, and the list matches how the existing skills are already written.

Two small thoughts, take or leave. The Anthropic skill authoring guide (https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices) also suggests using consistent terminology within a skill (one term per concept, not "field" in one section and "element" in the next) and forward-slash paths in any file path examples so they work cross-platform. Would either be worth a bullet here, or is that beyond what this list is aiming for?

Also, the two lists in CONTRIBUTING.md and skill-anatomy.md overlap quite a bit. Would it make sense to keep one as the canonical copy and link to it from the other, so they can't drift apart over time?

addyosmani:
Thanks. Skill formatting conventions are documented in docs/skill-anatomy.md, so this overlaps it. Closing.

PR #117, opened 2026-04-30
fix(hooks): resolve security eval and session-start output schema

googlarz:
## Summary

Fixes two issues surfaced in #106 and #110.

### Fix #106 — eval injection vector in simplify-ignore-test.sh (Critical)

`eval "$(sed -n '/^filter_file()/,/^}/p' hooks/simplify-ignore.sh)"` extracted and eval'd a function at test-runner startup. Any future modification to `simplify-ignore.sh` that placed malicious content in or near `filter_file` would execute without review, since test runners are read with low suspicion.

Fix: restructured `simplify-ignore.sh` so it is safe to source:
- All helper functions (`hash_cmd`, `file_id`, `block_hash`, `escape_glob`, `filter_file`) come first
- A `BASH_SOURCE` guard immediately follows — sourcing stops here, so no jq check, no `INPUT=$(cat)`, no stdin blocking
- Hook execution logic (jq check, stdin read, dispatch) runs only when the script is executed directly
- `CACHE` uses `${CACHE:-...}` so callers (e.g. tests) can override via an exported variable

`simplify-ignore-test.sh` now sources the file with `. hooks/simplify-ignore.sh` — clean, no eval.

All 21 tests pass.

### Fix #110 — session-start.sh uses undocumented JSON output schema

`{"priority": "IMPORTANT", "message": "..."}` uses fields that don't exist in the Claude Code hook schema. When CC receives JSON starting with `{` but with unrecognized fields, it silently ignores them — the meta-skill content was never injected.

Claude Code docs define two valid SessionStart output shapes:
1. **Plain text stdout** — injected directly as additional context ✓
2. **`{"hookSpecificOutput": {"additionalContext": "..."}}`** — correct structured schema, but broken for plugin-registered hooks (CC bug #16538 silently drops plugin SessionStart `additionalContext`)

Fix: switch to plain text stdout (shape 1), which works reliably for both native and plugin hooks. `jq` dependency removed.

## Test plan

- [x] `bash hooks/simplify-ignore-test.sh` — 21/21 pass
- [x] `bash hooks/session-start.sh` — outputs plain text with meta-skill content
- [x] Codex review — clean pass, no findings

Fixes #106, #110

PR #211, opened 2026-05-28
Fix SessionStart hook to use documented stdout payload

### LINKED ISSUE #110: SessionStart hook silently fails: invalid JSON format + Claude Code Bug #16538 — with fix and smart loading proposal
## Problem

The `session-start.sh` hook produces **zero visible output** in Claude Code sessions, even though the script executes successfully. The meta-skill `using-agent-skills` is silently discarded on every session start.

### Layer 1: Invalid Output Format (Primary)

The hook outputs an **undocumented JSON format** that Claude Code does not recognize:

```json
{"priority": "IMPORTANT", "message": "agent-skills loaded...\n\n[FULL SKILL.MD]"}
```

**Why this fails:**
1. The official Claude Code hooks reference ([code.claude.com/docs/en/hooks.md](https://code.claude.com/docs/en/hooks.md)) defines valid SessionStart hook output as:
   - **Plain text stdout** — any non-JSON text is added directly as context
   - **`hookSpecificOutput.additionalContext`** — structured JSON with specific schema
   - **`systemMessage`** — universal top-level field
   
2. The `"priority"` / `"message"` top-level fields **do not exist** in the Claude Code hook schema. They are neither documented in the hooks reference nor in the agent-sdk hooks docs.

3. When Claude Code receives output starting with `{` but not matching the valid schema, it attempts to parse it as JSON. The unrecognized fields are silently ignored, and no context is injected.

4. This was verified by comparing with other plugins:
   - **superpowers** ([obra/superpowers](https://github.com/obra/superpowers)) uses valid `hookSpecificOutput` schema
   - **oh-my-claudecode** uses plain text with budget guards
   - **claude-mem** uses `{"continue":true,"suppressOutput":true}` (valid universal fields)

### Layer 2: Plugin Pipeline Bug (Secondary, Upstream)

Even if the format were fixed, Claude Code has a known upstream bug:

- **Bug #16538** ([anthropics/claude-code#16538](https://github.com/anthropics/claude-code/issues/16538)): Plugin-defined `SessionStart` hooks that output `hookSpecificOutput.additionalContext` have their context **silently dropped**. Native hooks defined in `~/.claude/settings.json` work correctly.
- Affected events: `SessionStart`, `PreCompact`
- Working events: `PostToolUse`, `UserPromptSubmit`

This means even a correctly-formatted `hookSpecificOutput` from the plugin's `hooks/hooks.json` would still fail.

### Verification Steps Performed

| Test | Result |
|------|--------|
| Manual execution of `session-start.sh` | ✅ Script runs, outputs JSON |
| Plugin enabled in `settings.json` | ✅ `agent-skills@addy-agent-skills: true` |
| Hook auto-merges from plugin | ✅ Confirmed via `/hooks` |
| Output visible in session context | ❌ Zero appearance |
| Native plain-text hook test | ✅ Works when defined in `settings.json` |
| Native broken-JSON hook test | ❌ Silent failure (confirms format is root cause) |

### 1. Change Output Format

Replaced the invalid `{"priority":"...","message":"..."}` format with **plain text stdout**:

```bash
#!/bin/bash
# ...

if [ -f "$META_SKILL" ]; then
  echo "agent-skills loaded. Project context detected. Most relevant skills are primed below."
  echo ""
  cat "$META_SKILL"
else
  echo "agent-skills: meta-skill not found."
fi
```

### 2. Make the Hook "Smart"

Instead of dumping the entire 8.6KB `SKILL.md` into every session, the hook now uses **tiered loading**:

- **Tier 1: Universal Core** — always loaded (Core Operating Behaviors, Failure Modes, Skill Discovery Flowchart)
- **Tier 2: Project Detection** — detects project type from repo files (`package.json`, `Cargo.toml`, etc.)
- **Tier 3: Skill Mapping** — maps project type to 3-5 most relevant skills
- **Tier 4: Primer Extraction** — loads only description + overview + "When to Use" from relevant skills
- **Tier 5: Budget Guard** — hard limit at 6000 chars (inspired by oh-my-claudecode's `SESSION_START_CONTEXT_BUDGET`)

Example: For a docs/wiki project → primed: `documentation-and-adrs`, `context-engineering`

### 3. Bypass Plugin Pipeline Bug

To avoid Bug #16538:

1. **Removed SessionStart from plugin's `hooks/hooks.json`**:
   ```json
   {"hooks": {}}
   ```
   
2. **Added native hook in `~/.claude/settings.json`**:
   ```json
   "SessionStart": [
     {
       "matcher": "",
       "hooks": [
         {
           "type": "command",
           "command": "bash $(cd ~ && pwd)/.claude/plugins/cache/addy-agent-skills/agent-skills/1.0.0/hooks/session-start.sh",
           "timeout": 10
         }
       ]
     }
   ]
   ```
   
   **Important:** Used `$(cd ~ && pwd)` instead of `${HOME}` because `${HOME}` is **not set** in Claude Code hook environments.

3. **Disabled auto-update** to prevent the fix from being overwritten:
   ```json
   "extraKnownMarketplaces": {
     "addy-agent-skills": {
       "source": {"source": "github", "repo": "addyosmani/agent-skills"},
       "autoUpdate": false
     }
   }
   ```

### Option A: Merge the Smart Hook (Recommended)

Replace `hooks/session-start.sh` with the smart version that:
1. Uses plain text output (valid Claude Code format)
2. Detects project type dynamically
3. Loads only relevant skill primers
4. Includes budget guard
5. Uses `$(cd ~ && pwd)` instead of `${HOME}`

### Option B: Minimal Fix (Quick)

If keeping the current architecture:
1. Change output from `{"priority":"...","message":"..."}` to plain text
2. Keep dumping full `SKILL.md` (accept the token cost)
3. Document that users must add the hook natively in `settings.json` to bypass Bug #16538

### Option C: Use `hookSpecificOutput.additionalContext`

```bash
jq -nc --arg ctx "$CONTENT" '
  {
    hookSpecificOutput: {
      hookEventName: "SessionStart",
      additionalContext: $ctx
    }
  }'
```

**Note:** This still requires the native-hook workaround because Bug #16538 affects all plugin-defined SessionStart hooks.

## Environment

- Claude Code versions tested: v2.1.39 – v2.1.86+ (issue confirmed across versions)
- Platform: macOS (darwin) — sed portability was also a concern
- jq installed: yes (jq-1.8.1)
- Plugin enabled: yes
- Hook executes: yes (verified manually and in debug logs)
- Output visible in session: **no** (before fix) / **yes** (after fix)

## References

- Claude Code Hooks Reference: [code.claude.com/docs/en/hooks.md](https://code.claude.com/docs/en/hooks.md)
- Claude Code Plugins Reference: [docs.anthropic.com/en/docs/claude-code/plugins-reference](https://docs.anthropic.com/en/docs/claude-code/plugins-reference)
- Bug #16538 — Plugin SessionStart hooks don't surface `additionalContext`: [github.com/anthropics/claude-code/issues/16538](https://github.com/anthropics/claude-code/issues/16538)
- Bug #256 — `CLAUDE_PLUGIN_ROOT` not set for SessionStart (fixed March 2026): [github.com/affaan-m/everything-claude-code/issues/256](https://github.com/affaan-m/everything-claude-code/issues/256)
- Related issue in planning-with-files plugin: [github.com/OthmanAdi/planning-with-files/issues/24](https://github.com/OthmanAdi/planning-with-files/issues/24)

MukundaKatta:
## Summary
The session-start hook emitted `{priority, message}` JSON, but those fields aren't in Claude Code's SessionStart hook schema. It only worked where unrecognized JSON got treated as context.

This prints the meta-skill as plain stdout (the documented SessionStart path) and drops the jq dependency. The regression test now checks stdout, and the obsolete jq-fallback steps are removed from CONTRIBUTING.

Fixes #110. Focused fix with no unrelated changes (unlike #117).

## Test
`bash hooks/session-start-test.sh` prints `session-start stdout payload OK`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,7 +52,7 @@ The frontmatter fields above are required. The section anatomy is a recommended 
 
 ## Testing Hooks
 
-The session-start hook (`hooks/session-start.sh`) injects the `using-agent-skills` meta-skill into every new Claude Code session. A regression test at `hooks/session-start-test.sh` validates the hook's JSON payload — both when `jq` is available and when it isn't.
+The session-start hook (`hooks/session-start.sh`) injects the `using-agent-skills` meta-skill into every new Claude Code session. SessionStart hooks add their stdout to the agent's context, so the hook prints the meta-skill as plain text. A regression test at `hooks/session-start-test.sh` validates the hook's stdout payload.
 
 Run it before opening any PR that touches:
 
@@ -63,21 +63,7 @@ Run it before opening any PR that touches:
 bash hooks/session-start-test.sh
 ```
 
-Expected output: `session-start JSON payload OK`. The script exits non-zero on any assertion failure.
-
-### Reproducing the no-jq fallback
-
-The hook gracefully degrades to an `INFO`-priority payload when `jq` isn't on `PATH`. To exercise that branch locally, strip `jq`'s directory from `PATH` for the test invocation:
-
-```bash
-JQ_DIR=$(dirname "$(command -v jq)")
-PATH=$(echo "$PATH" | tr ':' '\n' | grep -v "^${JQ_DIR}$" | tr '\n' ':' | sed 's/:$//') \
-  bash hooks/session-start-test.sh
-```
-
-This works cleanly when `jq` lives in its own directory (e.g. `/opt/homebrew/bin` from Homebrew, `/usr/local/bin` from a manual install). If your `jq` shares a system bin with other tools the test depends on (such as `mktemp` in `/usr/bin`), the simpler approach is to install `jq` via a separate package manager so it has its own bin directory, then re-run.
-
-The hook's `command -v jq` check fails under the stripped `PATH`, the `INFO`-priority fallback runs, and the test asserts the `jq is required` guidance message instead of the normal payload.
+Expected output: `session-start stdout payload OK`. The script exits non-zero on any assertion failure.
 
 ## Reporting Issues
 
--- a/hooks/session-start-test.sh
+++ b/hooks/session-start-test.sh
@@ -1,46 +1,25 @@
 #!/bin/bash
-# session-start-test.sh - Tests for the SessionStart hook JSON payload
+# session-start-test.sh - Tests for the SessionStart hook stdout payload
 
 set -euo pipefail
 
 tmp_payload="$(mktemp)"
 trap 'rm -f "$tmp_payload"' EXIT
 
-has_jq=0
-if command -v jq >/dev/null 2>&1; then
-  has_jq=1
-fi
+bash hooks/session-start.sh > "$tmp_payload"
 
-payload="$(bash hooks/session-start.sh)"
-printf '%s' "$payload" > "$tmp_payload"
-
-HAS_JQ="$has_jq" PAYLOAD_PATH="$tmp_payload" node <<'NODE'
+PAYLOAD_PATH="$tmp_payload" node <<'NODE'
 const fs = require('fs');
 
-const payload = JSON.parse(fs.readFileSync(process.env.PAYLOAD_PATH, 'utf8'));
-const hasJq = process.env.HAS_JQ === '1';
+const payload = fs.readFileSync(process.env.PAYLOAD_PATH, 'utf8');
 
-if (hasJq) {
-  if (payload.priority !== 'IMPORTANT') {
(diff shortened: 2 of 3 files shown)


addyosmani:
Closing this as already resolved. The SessionStart hook was reworked to emit the documented stdout payload and there's a regression test guarding it on main now, so the fix you're proposing has effectively landed by another path. Thanks for chasing it down, and sorry the timing meant it got there first from elsewhere. If you spot a case the current hook still gets wrong, reopen with a repro.

PR #216, opened 2026-06-03
add vietnamese beginner guide

Drmovie86:


federicobartoli:
Thanks for the contribution. We’re going to decline one-off translations for now because we don’t currently have a localization strategy or a way to keep translated docs in sync with the English README/docs over time.

If we decide to support translations later, we’ll likely do that with a maintained localization workflow rather than standalone translated docs.

PR #218, opened 2026-06-05
docs: update Cursor setup guide for rules and skills layout

nguyenducthaonguyen:
Reflect current Cursor project context (.cursor/rules + .cursor/skills) instead of legacy monolith patterns.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ claude --plugin-dir /path/to/agent-skills
 <details>
 <summary><b>Cursor</b></summary>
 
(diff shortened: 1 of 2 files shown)


addyosmani (approved):
The old version of this guide was quietly steering people at the legacy copy-into-rules path, so this is overdue: Cursor grew real .cursor/skills support and our docs should reflect that. I checked the mechanism is actually current before pulling it in, and the "rules stay thin, skills carry the workflow" framing lines up with how we think about context everywhere else in the pack. Good catch keeping this honest.

PR #219, opened 2026-06-06
Strengthen security skills: threat modeling, SSRF, supply chain & AI/LLM

addyosmani:
### What this does

A focused quality uplift to the repo's security assets - **+28%** to the `security-and-hardening` skill - without expanding its scope or touching unrelated skills. 

### Why

The skill was already strong on the OWASP web Top 10 from a defensive angle, but had four gaps:

1. It was a **checklist of controls with no process** - nothing told the agent to think like an attacker *first*. OWASP A04 (Insecure Design) was named in a table but never actioned.
2. **SSRF** was listed as OWASP #10 but, unlike every other item, had no pattern or example.
3. No **supply-chain** guidance beyond `npm audit` (which doesn't catch malicious/typosquatted packages).
4. **Zero coverage of AI/LLM security** - a notable miss for a repo about AI coding agents, where apps increasingly ship LLM features.

### Changes

**`skills/security-and-hardening/SKILL.md`** (349 → 448 lines)
- **Threat Model First** - a STRIDE / trust-boundary / abuse-case process step up front, turning the skill from a checklist into a process
- **SSRF prevention** - allowlist + private-IP-block pattern, with the `169.254.169.254` cloud-metadata callout
- **Supply-chain hygiene** - lockfile/`npm ci`, postinstall and typosquat risks
- **Securing AI/LLM features** - mapped to the [OWASP Top 10 for LLM Applications (2025)](https://genai.owasp.org/llm-top-10/): prompt injection, untrusted output handling, secrets-in-prompts, excessive agency, unbounded consumption - with a BAD/GOOD example
- **Secret-leak response** (rotate, don't just delete) + new entries in the review checklist, rationalizations, red flags, and verification

**`references/security-checklist.md`** (134 → 176) - mirrors the above: a Threat Modeling "start here" block, SSRF item, supply-chain checks, an AI/LLM Security section, and an OWASP LLM Top 10 quick-reference table.

**`agents/security-auditor.md`** (101 → 112) - new "AI/LLM Features" review scope, an SSRF question, and rules to start from trust boundaries (STRIDE) and check the LLM Top 10 + supply-chain risk.

### Scope note

This deliberately does **not** import the broad security-operations material (threat hunting, forensics, red teaming, MITRE ATT&CK) found in larger cybersecurity skill packs - that would conflict with this repo's builder-focused mission. It borrows the *frameworks* (STRIDE, OWASP web + LLM Top 10, supply-chain integrity) and applies them to writing secure application code.

### DIFF
--- a/agents/security-auditor.md
+++ b/agents/security-auditor.md
@@ -43,6 +43,16 @@ You are an experienced Security Engineer conducting a security review. Your role
 - Are webhook payloads verified (signature validation)?
 - Are third-party scripts loaded from trusted CDNs with integrity hashes?
 - Are OAuth flows using PKCE and state parameters?
+- Are server-side fetches of user-supplied URLs allowlisted (SSRF)?
+
+### 6. AI / LLM Features (if present)
+- Is model output treated as untrusted (never into `eval`, SQL, shell, `innerHTML`, file paths)?
+- Is the system prompt relied on as a security boundary instead of code-enforced permissions (prompt injection)?
+- Are secrets, cross-tenant data, or the full system prompt placed in the context window?
+- Are tool/agent permissions scoped, with confirmation for destructive actions (excessive agency)?
+- Are token, rate, and recursion limits set (unbounded consumption)?
+
+Map findings to the OWASP Top 10 for LLM Applications where relevant.
 
 ## Severity Classification
 
@@ -90,9 +100,10 @@ You are an experienced Security Engineer conducting a security review. Your role
 2. Every finding must include a specific, actionable recommendation
 3. Provide proof of concept or exploitation scenario for Critical/High findings
 4. Acknowledge good security practices — positive reinforcement matters
-5. Check the OWASP Top 10 as a minimum baseline
-6. Review dependencies for known CVEs
+5. Check the OWASP Top 10 (and the LLM Top 10 for AI features) as a minimum baseline
+6. Review dependencies for known CVEs and supply-chain risk (typosquats, postinstall scripts)
 7. Never suggest disabling security controls as a "fix"
+8. Start from trust boundaries — where untrusted data enters — and reason about each with STRIDE before enumerating findings
 
 ## Composition
 
--- a/references/security-checklist.md
+++ b/references/security-checklist.md
@@ -4,6 +4,7 @@ Quick reference for web application security. Use alongside the `security-and-ha
 
 ## Table of Contents
 
+- [Threat Modeling (Start Here)](#threat-modeling-start-here)
 - [Pre-Commit Checks](#pre-commit-checks)
 - [Authentication](#authentication)
 - [Authorization](#authorization)
@@ -12,8 +13,19 @@ Quick reference for web application security. Use alongside the `security-and-ha
 - [CORS Configuration](#cors-configuration)
 - [Data Protection](#data-protection)
 - [Dependency Security](#dependency-security)
+- [AI / LLM Security](#ai--llm-security)
 - [Error Handling](#error-handling)
 - [OWASP Top 10 Quick Reference](#owasp-top-10-quick-reference)
+- [OWASP Top 10 for LLMs Quick Reference](#owasp-top-10-for-llms-quick-reference)
+
+## Threat Modeling (Start Here)
+
+Before reaching for controls, spend five minutes thinking like an attacker:
+
+- [ ] Trust boundaries mapped (requests, uploads, webhooks, third-party APIs, LLM output)
+- [ ] Assets named (credentials, PII, payment data, admin actions, money movement)
+- [ ] STRIDE run per boundary (Spoofing, Tampering, Repudiation, Info disclosure, DoS, Elevation)
+- [ ] Abuse cases written next to use cases ("how would I misuse this?")
 
 ## Pre-Commit Checks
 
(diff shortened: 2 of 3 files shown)


addyosmani:
cc @federicobartoli  @nucliweb would love your thoughts if time allows :)

federicobartoli:
> cc @federicobartoli @nucliweb would love your thoughts if time allows :)

Thanks for the ping Addy! This looks like a really useful direction.
I’ll go through it properly later today and leave thoughts/comments if anything stands out.

nucliweb:
Threat-Model-First framing is the right move, it turns the skill from a checklist into a process. Two items worth tightening before merge, plus a few smaller notes.

## Worth addressing before merge

### 1. OWASP numbering is internally inconsistent

`## OWASP Top 10 Prevention` in `SKILL.md` uses pre-2021 names/numbering (Broken Authentication, Sensitive Data Exposure, XSS as its own item) and this PR adds SSRF as item **7**. In OWASP Top 10 **2021**, SSRF is **A10**. Meanwhile the `OWASP Top 10 Quick Reference` table in `references/security-checklist.md` does use the 2021 numbering `(SSRF = #10)`. Same skill ends up stating two different things.

Suggest: either renumber the prevention section to 2021, or drop the numbers and rename to `## OWASP Top 10 Prevention Patterns` so readers don't infer a ranking that isn't there. The second is the least invasive.

### 2. SSRF example is vulnerable to DNS rebinding / TOCTOU

```ts
const { address } = await lookup(url.hostname);
if (ipaddr.parse(address).range() !== 'unicast') throw …;
return url;
…
await fetch(await assertSafeUrl(req.body.webhookUrl), { redirect: 'error' });
```

Two DNS resolutions happen: the one you validate (`lookup`) and the one `fetch` does internally. An attacker with multiple A-records or short TTL can pass validation and still have `fetch` connect to an internal IP. The example is presented as the secure pattern, so it should either add a caveat ("does not defend against DNS rebinding; for high-risk surfaces, resolve once and connect by IP, or use `ssrf-req-filter` / `request-filtering-agent`") or switch to `lookup(host, { all: true })` and reject if any returned address is non-unicast.

The `range() !== 'unicast'` check itself is correct (covers IPv4/IPv6 link-local, loopback, private, unique-local via ipaddr.js). The issue is purely the TOCTOU gap.

## Minor

- **LLM08: Vector and Embedding Weaknesses** isn't covered. The skill names RAG explicitly, so a bullet on poisoned embeddings and cross-tenant leakage in shared vector stores would round out the LLM section.
- **Repo convention.** `CLAUDE.md` says each skill should have `Overview, When to Use, Process, Common Rationalizations, Red Flags, Verification`. The skill has no `## Process` header, "Threat Model First" is effectively the process. Promoting it to `## Process` aligns with the documented structure.
- **`JSON.parse` can throw** in `CommandSchema.parse(JSON.parse(await llm.replyJson(...)))`. For a "GOOD" example, `safeParse` or try/catch makes it clearer that the parse step is also handling untrusted data.

## Nits

- PR description (not the merged content) has nested brackets on the OWASP LLM link: `[[OWASP …](url)](url)` renders oddly.

Overall: LGTM once `#1` and `#2` are addressed.

addyosmani:
Thanks @nucliweb, sharp review and both of the main points were right. Pushed `d198392` addressing everything.

**1. OWASP numbering inconsistency.** Went with your least-invasive option: renamed the section to "OWASP Top 10 Prevention Patterns", dropped the per-item numbers, and added a line pointing readers to the 2021 quick-reference table for the actual ranking. The skill no longer asserts two different numberings, and SSRF is no longer implied to be "item 7".

**2. SSRF DNS rebinding / TOCTOU.** Fixed the example and kept it honest. It now resolves all records with `lookup(host, { all: true })` and rejects if any returned address is non-unicast. I also added an explicit caveat that `fetch` re-resolves (so this does not fully close DNS rebinding), with pointers to `request-filtering-agent` / `ssrf-req-filter` and the "resolve once, connect to the pinned IP" approach for high-risk surfaces. Good catch; that gap was real.

On the minor notes:

- **LLM08 (Vector and Embedding Weaknesses).** Added a bullet to the LLM section (per-tenant embedding isolation, and validating documents before indexing so poisoned content cannot steer answers), plus a matching row in the reference table. The skill names RAG, so you were right it belonged there.
- **Process header.** Promoted "Threat Model First" to `## Process: Threat Model First` so it matches the documented skill anatomy (Overview, When to Use, Process, and so on).
- **`JSON.parse` can throw.** The GOOD example now parses defensively: `JSON.parse` plus the schema check sit in a try/catch that throws a `ValidationError`, so it models handling untrusted output end to end rather than implying `JSON.parse` is safe.

Also fixed the nested-bracket OWASP LLM link in the PR description (the nit).

Thanks again; the skill is meaningfully better for this.

federicobartoli:
Tiny optional note : the LLM quick-reference table is intentionally very builder-focused, which I think makes sense for this repo.
The only thing I wondered is whether the heading “OWASP Top 10 for LLMs Quick Reference” may make readers expect all 10 entries. Since LLM04 and LLM09 are not listed, maybe either adding very short rows for completeness or renaming it to “Selected OWASP LLM risks for application builders” would avoid that ambiguity.

Not a blocker from my side; the current scope feels reasonable. @addyosmani @nucliweb 
P.S. Sorry for the delay in my reply

addyosmani:
Good call @federicobartoli. Went with the completeness option: the table now lists all ten, with LLM04 (Data and Model Poisoning) and LLM09 (Misinformation) added in order. A "Top 10" heading should hold ten rows, so a full reference reads better than trimming the title. Pushed in `cad41cb`.

PR #220, opened 2026-06-06
feat(build): add /build auto for a one-pass plan + implement

addyosmani:
### What & why

The `refine → spec → plan → build` chain is deliberately careful, but once a `SPEC.md` exists, hand-cranking every task is friction - you end up approving the *route* and every single *turn*. This adds a fast path: **approve the route once, then let it drive.**

`/build` gains a second mode:

| Command | Behavior |
|---|---|
| `/build` | Unchanged - implement the next task, then stop (one slice at a time) |
| `/build auto` | Require a spec → generate the plan if missing → **one approval** → implement every task autonomously |

(`auto`, `all`, and `fast` all trigger autonomous mode.)

### What it keeps

Autonomous mode removes the manual *stepping*, not the *verification*. Every task still runs the full TDD loop (RED → GREEN → regression → build) and gets **its own commit**, so any point is a clean rollback. It auto-pauses and asks the human only when it should:

- a test it can't make pass, or a build break without an obvious fix → `debugging-and-error-recovery`
- spec ambiguity, or a decision the spec doesn't cover
- a **high-risk/irreversible** task - auth/permission changes, destructive migrations, payments, deletions, deploys, anything touching secrets → `doubt-driven-development` + explicit sign-off

It ends with a summary of tasks completed, tests added, commits made, and anything skipped or flagged.

### Changes

- `.claude/commands/build.md` - two-mode definition
- `.gemini/commands/build.toml` - mirrored for Gemini CLI (we should probably account for Antigravity CLI more)
- `README.md`, `docs/getting-started.md` - document the new mode

### Test plan

- `/build` with no argument still implements a single task and stops.
- `/build auto` with no `SPEC.md` stops and prompts for `/spec` (doesn't invent requirements).
- `/build auto` with a spec but no `tasks/plan.md` generates the plan, asks for one approval, then runs.
- Gemini `build.toml` parses (validated) and mirrors the Claude prompt.

### DIFF
--- a/.claude/commands/build.md
+++ b/.claude/commands/build.md
@@ -29,15 +29,16 @@ Pick the next pending task from the plan. Then:
 Use this once a spec exists and you want to collapse plan + build into one run. It removes the manual stepping between tasks — **not** the verification. Every task still earns a passing test and its own commit.
 
 1. **Require a spec.** Look only for a spec at a known path: `SPEC.md` at the repo root, `docs/SPEC.md`, or a file under `spec/`. A README or arbitrary doc does **not** count. If none exists, stop and tell the user to run `/spec` first — do not invent requirements.
-2. **Plan if needed.** If there is no `tasks/plan.md`, invoke agent-skills:planning-and-task-breakdown to generate one.
-3. **Single checkpoint.** Present the full plan and wait for an unambiguous affirmative (e.g. "approve", "go", "yes"). Treat hedged responses ("looks reasonable", "I guess") as **not** approved. This is the only human gate — after approval, run autonomously.
-4. **Execute every task in dependency order.** Use each task's declared dependencies; if they aren't explicit, execute in the order the plan lists them. For each task, run the full default loop above (RED → GREEN → regression → build → commit per task → mark complete). One commit per task so any point is a clean rollback.
-5. **Stop and ask the user** (do not push through) when:
+2. **Establish a clean baseline.** Run `git status --porcelain`. If there are uncommitted changes outside the expected planning artifacts (`SPEC.md`, `docs/SPEC.md`, `spec/*`, `tasks/plan.md`, `tasks/todo.md`), stop and ask the user to commit, stash, or confirm how to handle them. Autonomous per-task commits must not absorb unrelated local work, or the clean-rollback guarantee breaks.
+3. **Plan if needed.** If there is no `tasks/plan.md`, invoke agent-skills:planning-and-task-breakdown to generate one.
+4. **Single checkpoint.** Present the full plan and wait for an unambiguous affirmative (e.g. "approve", "go", "yes"). Treat hedged responses ("looks reasonable", "I guess") as **not** approved. This is the only human gate — after approval, run autonomously. If you generated `tasks/plan.md`, commit it as a single preparatory commit now so it doesn't bleed into the first task's commit.
+5. **Execute every task in dependency order.** Use each task's declared dependencies; if they aren't explicit, execute in the order the plan lists them. For each task, run the full default loop above (RED → GREEN → regression → build → commit → mark complete). Stage only the files that task touched plus its task-status update — never `git add -A` blindly — and make one commit per task so any point is a clean rollback.
+6. **Stop and ask the user** (do not push through) when:
    - a test can't be made to pass or the build breaks without an obvious fix → follow agent-skills:debugging-and-error-recovery
    - the spec is ambiguous, or a task needs a decision the spec doesn't cover
    - a task is high-risk or irreversible — auth/permission changes, destructive data migrations, payments, deletions, deploys, anything touching secrets, **or anything you can't undo with `git revert`** → follow agent-skills:doubt-driven-development and get explicit sign-off before continuing
 
    After the user resolves a blocker, they re-invoke `/build auto` — it resumes from the next pending task.
-6. **Summarize at the end:** tasks completed, tests added, commits made, and anything skipped, flagged, or left for the user.
+7. **Summarize at the end:** tasks completed, tests added, commits made, and anything skipped, flagged, or left for the user.
 
 If any step fails, follow the agent-skills:debugging-and-error-recovery skill.
--- a/.claude/commands/build.md
+++ b/.claude/commands/build.md
@@ -1,10 +1,19 @@
 ---
-description: Implement the next task incrementally — build, test, verify, commit
+description: Implement tasks incrementally — build, test, verify, commit. Add "auto" to run the whole plan in one approved pass.
 ---
 
 Invoke the agent-skills:incremental-implementation skill alongside agent-skills:test-driven-development.
 
-Pick the next pending task from the plan. For each task:
+## Modes
+
+- **`/build`** — implement the *next* pending task, then stop (careful, one slice at a time).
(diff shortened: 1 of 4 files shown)


nucliweb:
Nice addition, the "approve the route once, then let it drive" framing is exactly right, and keeping the per-task TDD loop + commit-per-task means the autonomy doesn't compromise the rollback story. Two things worth tightening before merge, plus a few smaller notes.

## Worth addressing before merge

### 1. `fast` as an alias is misleading

`auto`, `all`, and `fast` all trigger autonomous mode, but autonomous mode isn't faster *per task*; it runs the same RED → GREEN → regression → build → commit loop. What it removes is the human stepping *between* tasks. A reader who types `/build fast` may reasonably assume "skip tests" or "skip verification."

Suggest: drop `fast` (the implicit promise doesn't match the behavior), keep `auto` as the canonical trigger, and keep `all` as a semantic alias since it genuinely communicates "the whole plan." If `fast` stays, the doc should say explicitly that it does not skip any verification step.

### 2. "SPEC.md (or equivalent)" is too loose

`/spec` writes `SPEC.md` at the project root canonically, so "or equivalent" opens the door to false positives (a stray `README.md`, `docs/notes.md`, etc.) being treated as a spec. Either tighten to `SPEC.md` at root only, or list a closed set of acceptable alternates (e.g. `SPEC.md`, `spec/*.md`, `docs/SPEC.md`). Otherwise step 1's "do not invent requirements" can be silently subverted by an agent that decides a README counts.

## Minor

- **Resume contract after a pause.** Step 5 stops on blockers but doesn't say how to resume. Worth one line: *"After the user resolves the blocker, re-invoke `/build auto`; it picks up from the next pending task."*
- **What counts as approval.** Step 3 says "explicit approval," but "explicit" doesn't define the form. A hedged "looks reasonable" could get treated as green light. Suggest: *"wait for an unambiguous affirmative (`approve`, `go`, `yes`); treat hedged responses as not approved."*
- **High-risk catch-all.** The enumerated list (auth, migrations, payments, deploys, secrets) is good, but a catch-all clause would future-proof it, something like *"...or anything you cannot undo via `git revert`."*
- **Plan-level checkpoints.** `planning-and-task-breakdown` already encourages checkpoints every 2–3 tasks in the plan itself. Autonomous mode could honor those instead of skipping them, useful for long plans. Probably a follow-up, not a blocker.

## Nits

- The `auto` section steps 1–6 read as agent-directed prose, while the PR body describes the same flow as user-directed. Already coherent inside the command; just flagging in case the README copy ends up mixed.
- Dependency-order execution (step 4) is well-supported by `planning-and-task-breakdown` (`Dependencies:` per task), so this works when the plan is generated via the skill. For hand-authored plans, a fallback line ("if dependencies aren't explicit, execute in declared order") would remove ambiguity.

Overall: LGTM with the two items above sharpened. Compatibility is clean, the high-risk gating is the right call, and the Gemini mirror is faithful.

addyosmani:
Thanks @nucliweb - genuinely useful review, and you're right on both of the main points. I decided to explore this direction after walking an audience through building an app using the skills and noticed that the per-task approval flow might feel a little "much" for some.

Pushed `faa69c8` addressing the feedback:

**1. Dropped the `fast` alias.** You nailed why it was wrong as it implies "skip tests" which is the opposite of what happens. `auto` is now canonical, `all` stays as the semantic alias, and both the Claude and Gemini commands now say explicitly that autonomous mode *isn't faster per task* - it runs the same RED → GREEN → regression → build loop and only removes the human stepping between tasks.

**2. Tightened spec detection.** Replaced "`SPEC.md` (or equivalent)" with a closed set - `SPEC.md` at the repo root, `docs/SPEC.md`, or a file under `spec/` and an explicit "a README or arbitrary doc does **not** count." That closes the loophole where step 1's "do not invent requirements" could be subverted by an agent deciding a README is a spec.

On the minor notes:

- **Resume contract** - added: "After the user resolves a blocker, they re-invoke `/build auto` — it resumes from the next pending task."
- **What counts as approval** - step 3 now requires an unambiguous affirmative (`approve` / `go` / `yes`) and explicitly treats hedged responses ("looks reasonable") as *not* approved.
- **High-risk catch-all** - added "…or anything you can't undo with `git revert`" alongside the enumerated list. Good future-proofing call.
- **Dependency fallback** - step 4 now says to use declared dependencies, and fall back to the plan's declared order when they aren't explicit. Covers hand-authored plans.

On **plan-level checkpoints**: I left this as a deliberate non-change for now. The single-approval gate is an intentional design decision for this PR ("approve the route once, then drive"), so honoring every in-plan checkpoint would reintroduce the per-phase pausing this mode is meant to remove. I think you're right that it's worth offering - but as a future opt-in (e.g. a `/build auto --checkpoints` style toggle) rather than default behavior, so I'd rather do it as a follow-up than fold it in here.

The two nits (agent- vs user-directed prose, and the dependency-order note) are handled by the above. Thanks again - this is sharper for it.

federicobartoli:
One additional safety edge I’d consider for `/build auto`: make the working-tree baseline explicit before autonomous execution.

The command’s rollback story depends on “one commit per task”, but right now the flow doesn’t say what to do if the repo already has uncommitted changes before auto mode starts. In that case, the agent could accidentally mix user work or unrelated local edits into the per-task commits, making `git revert` much less clean.

There is one subtle case: `/build auto` may intentionally generate `tasks/plan.md` before the approval checkpoint, and `/spec` / `/plan` artifacts may already exist as working documents. So maybe the rule should be:

- before autonomous implementation starts, inspect `git status --porcelain` and establish the baseline;
- if there are uncommitted changes outside expected planning artifacts (`SPEC.md`, `docs/SPEC.md`, `spec/*`, `tasks/plan.md`, `tasks/todo.md`), stop and ask the user to commit/stash/confirm how to handle them;
- if `tasks/plan.md` is generated, either commit it as a small preparatory commit after approval or treat it as the only allowed pre-task change;
- during each task, stage only the files touched for that task, plus the task-status update, before committing.

This would make the “one commit per task / clean rollback” guarantee stronger without adding another human approval gate between tasks.

addyosmani:
Great edge to catch, @federicobartoli. You're right that "one commit per task" only holds if the tree starts clean, and the flow said nothing about that. Pushed `57f5274` with your suggested shape:

- **Baseline check before execution.** Auto mode now runs `git status --porcelain` first. If anything is dirty outside the expected planning artifacts (`SPEC.md`, `docs/SPEC.md`, `spec/*`, `tasks/plan.md`, `tasks/todo.md`), it stops and asks the user to commit, stash, or confirm, so unrelated local work never lands in a per-task commit.
- **Plan as a preparatory commit.** A generated `tasks/plan.md` gets committed on its own right after approval, so it doesn't bleed into the first task.
- **Scoped staging.** Each task stages only the files it touched plus its status update. No blind `git add -A`.

All of it sits before the single approval gate or inside the existing per-task loop, so it strengthens the rollback guarantee without adding another between-task pause, which was your goal too. Mirrored across the Claude and Gemini commands. Thanks for the careful read.

samuelhenriquejava:
achei muito boa esta explicação sou novo na área estou estudando Java e tenho trabalhado com as IA  para obter explicações para desenvolver projetos muito obrigado.

PR #221, opened 2026-06-09
fix(browser-testing): correct chrome-devtools-mcp install config

nucliweb:
## Summary
- Fix package name: `@anthropic/chrome-devtools-mcp` does not exist; correct package is `chrome-devtools-mcp`
- Add `-y` to npx args to skip install confirmation prompt
- Add `--autoConnect` as recommended default, with a short explanation
- Restore the context line indicating where the config block should be placed (`.mcp.json` or Claude Code settings)

## Test plan
- [ ] Verify the config block matches a working `.mcp.json` / `claude_desktop_config.json` setup with `chrome-devtools-mcp`

### DIFF
--- a/skills/browser-testing-with-devtools/SKILL.md
+++ b/skills/browser-testing-with-devtools/SKILL.md
@@ -25,18 +25,20 @@ Use Chrome DevTools MCP to give your agent eyes into the browser. This bridges t
 
 ### Installation
 
-```bash
-# Add Chrome DevTools MCP server to your Claude Code config
-# In your project's .mcp.json or Claude Code settings:
+Add the following to your project's `.mcp.json` or Claude Code settings:
+
+```json
 {
   "mcpServers": {
     "chrome-devtools": {
       "command": "npx",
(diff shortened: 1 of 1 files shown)


federicobartoli (commented):
LGTM! My only concern is @latest: for reproducibility/security, pinning a specific chrome-devtools-mcp version would be safer? @nucliweb

nucliweb:
> LGTM! My only concern is @latest: for reproducibility/security, pinning a specific chrome-devtools-mcp version would be safer? @nucliweb

I see your point; not using `@latest` would also download the latest version if it isn’t already installed. And pinning it to a specific version would mean we’d have to keep an eye out for updates. The most I’d do in that regard is pin the major version.

What do you think, @addyosmani?

addyosmani:
Thanks for the catch and discussion folks. I am personally OK with us opting for latest here and folks can always have their agent patch it to a more compatible pinned version if there is an issue with that version. Going to go ahead and merge this one as is. Thank you!

PR #61, opened 2026-04-11
Add observability-and-monitoring skill

googlarz:
## Summary

Adds `observability-and-monitoring` — the one confirmed lifecycle gap in the collection (no existing PR, not covered by any current skill). Covers the full production visibility workflow: structured logging, four golden signals metrics, OpenTelemetry distributed tracing, SLO-based alerting, and dashboard design.

## Changes

- **New skill:** `skills/observability-and-monitoring/SKILL.md`
- **New reference:** `references/observability-checklist.md` — logging, metrics, tracing, SLOs/alerting, dashboards, pre-launch gate
- **`incremental-implementation`:** one-liner anchoring observability to slice definition ("define what you'll measure when you define the slice")
- **`using-agent-skills`:** decision tree, lifecycle (step 11, before shipping), quick reference table
- **`idea-refine`:** forward reference to `spec-driven-development`
- **`deprecation-and-migration`:** references to `incremental-implementation`, `test-driven-development`, `git-workflow-and-versioning`
- **README:** skill count 20 → 21, new entry in Ship section and directory tree

## Review feedback addressed (from #59)

| Feedback | Fix |
|---|---|
| Decision tree trigger is reactive | Reworded to "Instrumenting a service / setting up monitoring?" |
| Observability only appears at step 11 | Added anchor in `incremental-implementation` — define measurements at slice definition time |
| No `references/observability-checklist.md` | Created — consistent with security-checklist.md, testing-patterns.md, performance-checklist.md |
| Steps assume greenfield | Added Step 0: brownfield audit — inventory existing instrumentation before adding more |

## Reviews run before submission

This branch went through `/codex:adversarial-review` and `/codex:review` before the original PR #59. Both findings were fixed (metric cardinality labelling, OTel propagation, `/ship` command wiring, lifecycle ordering). This PR is the observability-only split of that work.

PR #223, opened 2026-06-09
feat: add observability-and-instrumentation skill

federicobartoli:
Fills the gap between debugging-and-error-recovery (dev-time) and
shipping-and-launch (launch-day monitoring): how to instrument code
as it's built so production behavior is visible and diagnosable.

Covers structured logging with correlation IDs, RED/USE metrics with
cardinality guidance, OpenTelemetry tracing, and symptom-based
alerting with runbooks. Follows the standard skill anatomy (Overview,
When to Use, Process, Rationalizations, Red Flags, Verification).

Registered in README tables/structure, CLAUDE.md phases, and the
using-agent-skills discovery tree (hooks/session-start-test.sh passes).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -20,7 +20,7 @@ docs/         → Setup guides for different tools
 **Build:** incremental-implementation, test-driven-development, context-engineering, source-driven-development, doubt-driven-development, frontend-ui-engineering, api-and-interface-design
 **Verify:** browser-testing-with-devtools, debugging-and-error-recovery
 **Review:** code-review-and-quality, code-simplification, security-and-hardening, performance-optimization
-**Ship:** git-workflow-and-versioning, ci-cd-and-automation, deprecation-and-migration, documentation-and-adrs, shipping-and-launch
+**Ship:** git-workflow-and-versioning, ci-cd-and-automation, deprecation-and-migration, documentation-and-adrs, observability-and-instrumentation, shipping-and-launch
 
 ## Conventions
 
--- a/README.md
+++ b/README.md
@@ -118,9 +118,9 @@ Skills are plain Markdown - they work with any agent that accepts system prompts
 
 ---
 
-## All 23 Skills
+## All 24 Skills
 
-The commands above are entry points. The pack includes 23 skills total — 22 lifecycle skills plus the `using-agent-skills` meta-skill. Each skill is a structured workflow with steps, verification gates, and anti-rationalization tables. You can also reference any skill directly.
+The commands above are entry points. The pack includes 24 skills total — 23 lifecycle skills plus the `using-agent-skills` meta-skill. Each skill is a structured workflow with steps, verification gates, and anti-rationalization tables. You can also reference any skill directly.
 
 ### Meta - Discover which skill applies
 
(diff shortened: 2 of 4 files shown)


nucliweb:
Good addition, fills a real gap between `debugging-and-error-recovery` (dev-time) and `shipping-and-launch` (launch-day). The signal-selection table with the `that/where/why` rule of thumb is clear, the cardinality guidance is specific enough to be actionable, and the symptom-based alerting section (symptom vs cause, two severities only) is one of the stronger pieces in the pack.

Two things to address before merging:

**Section heading.** `## Core Process` should be `## Process` to match the skill anatomy convention. CLAUDE.md defines the required sections as: Overview, When to Use, Process, Common Rationalizations, Red Flags, Verification. Every other skill in the pack uses `## Process`.

**Lifecycle placement.** The skill says "instrumentation is not a post-launch add-on; it's written alongside the feature, the same way tests are." Placing it at step 15 in `using-agent-skills` (immediately before `shipping-and-launch`) sends the opposite message to an agent following the sequence. Consider moving it to run alongside `test-driven-development` (step 6-7), or adding a note like "(parallel with implementation, not sequential)" so the placement doesn't contradict the skill's core message.

federicobartoli:
> Good addition, fills a real gap between `debugging-and-error-recovery` (dev-time) and `shipping-and-launch` (launch-day). The signal-selection table with the `that/where/why` rule of thumb is clear, the cardinality guidance is specific enough to be actionable, and the symptom-based alerting section (symptom vs cause, two severities only) is one of the stronger pieces in the pack.
> 
> Two things to address before merging:
> 
> **Section heading.** `## Core Process` should be `## Process` to match the skill anatomy convention. CLAUDE.md defines the required sections as: Overview, When to Use, Process, Common Rationalizations, Red Flags, Verification. Every other skill in the pack uses `## Process`.
> 
> **Lifecycle placement.** The skill says "instrumentation is not a post-launch add-on; it's written alongside the feature, the same way tests are." Placing it at step 15 in `using-agent-skills` (immediately before `shipping-and-launch`) sends the opposite message to an agent following the sequence. Consider moving it to run alongside `test-driven-development` (step 6-7), or adding a note like "(parallel with implementation, not sequential)" so the placement doesn't contradict the skill's core message.

Thanks for the review @nucliweb  <3 ! Both addressed in [5c11d59](https://github.com/addyosmani/agent-skills/pull/223/commits/5c11d590479edbb8244bd7e5313c31958e01db23).
Renamed the heading to ## Process. Small note: skill-anatomy.md does list "Core Process / The Workflow / Steps" as acceptable equivalents, and most skills in the pack actually vary ("The Process", "Process: Threat Model First"...) , but no reason to add another variant, so renamed anyway.
Good catch on the lifecycle placement, that was a real contradiction with the skill's own premise. Moved it to step 8, right next to implementation, with an explicit note that it runs parallel with 7-9 rather than as a sequential step before launch.

addyosmani:
Great work! This is honestly one of the strongest skill submissions we've seen yet, @federicobartoli.

Starting with questions (*"define working before instrumenting"*) is exactly the right move and that signal-selection table with the that/where/why split is nice. I also checked @nucliweb's notes and you handled both perfectly (the `## Process` heading and placing it parallel with the implementation).

One quick heads-up on the coordination side. We actually have an open PR (#61 by googlarz) covering the exact same observability gap (structured logging, metrics, OpenTelemetry, and SLO alerts). We can't merge two separate observability skills without hitting our duplication rule, so I need to decide whether to pick one or blend the best pieces of both into a single skill. It's zero reflection on the quality of your work here, just want to let you know why I'm holding the merge for a moment while I sort that out.

@nucliweb any thoughts between the two?

One tiny, totally optional note: your tracing example is beautifully vendor-neutral using OTel, but the metrics example leans a bit Prometheus-specific (`prom-client`). If you want, adding a quick line mentioning OTel metrics as the neutral path (with `prom-client` as just one choice of backend) would tie the two halves together nicely. Definitely not a blocker, though.

Seriously strong work on this. Let me get the overlap with #61 squared away, and I’ll get right back to you.

federicobartoli:
Thanks @addyosmani  that means a lot!
Added the OTel note in e5e8729 , fair point: OTel metrics is now framed as the default path, with prom-client as one backend choice.
Totally understand holding the merge for the #61 overlap. One factual note that might help the comparison, from reading its current diff: it doesn't appear to contain the observability SKILL.md itself: it has references/observability-checklist.md plus an unrelated skills/assistant/ package (Google Calendar OAuth scripts and an MCP server, ~1700 lines of Python) and a fork-promo note added to the README. Worth double-checking the Files tab before deciding how to blend. Either way, happy to consolidate with whatever pieces of #61 you want kept.

addyosmani:
Update on the #61 overlap: I went through its Files tab and you read it right. There's no observability SKILL.md in there, only a `references/observability-checklist.md` bundled with an unrelated `skills/assistant/` package (the Calendar OAuth and MCP code) and a fork-promo README note. A checklist isn't a competing skill, so the duplication rule doesn't bite, and #61 isn't landing in that shape anyway. I'm no longer holding this PR for the overlap, and the OTel-metrics framing in e5e8729 was the last thing on my list. Approve from me.

The one piece of #61 worth saving is that observability checklist, which maps onto how we already pair a skill with a quick-reference (security-and-hardening has security-checklist). If you're up for it, a follow-up adding `references/observability-checklist.md` as this skill's companion would fold in the useful half of #61 and give people the at-a-glance version. No need to touch it here. The only open item is the skill-count line you and #224 both edit, so whoever merges second takes the small rebase. Nice work landing this one.

nucliweb:
> @nucliweb any thoughts between the two?

Sorry, I didn't have time to go through this

PR #224, opened 2026-06-09
Add legacy-code-and-refactoring skill

federicobartoli:
All the current skills assume new or at least tested code, but the most common thing I actually point an agent at is old untested code that nobody remembers why it works. code-simplification doesn't cover it — it explicitly wants code that's already readable and tested.

So this adds the classic Feathers playbook as a skill: characterization tests before touching anything, seams and sprout to make untestable code testable, refactor and behavior change always in separate commits, strangler fig instead of rewrites.

Wired into README, CLAUDE.md and the using-agent-skills tree. session-start-test.sh passes since I touched the meta-skill.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,7 +17,7 @@ docs/         → Setup guides for different tools
 
 **Define:** interview-me, idea-refine, spec-driven-development
 **Plan:** planning-and-task-breakdown
-**Build:** incremental-implementation, test-driven-development, context-engineering, source-driven-development, doubt-driven-development, frontend-ui-engineering, api-and-interface-design
+**Build:** incremental-implementation, test-driven-development, context-engineering, source-driven-development, doubt-driven-development, frontend-ui-engineering, api-and-interface-design, legacy-code-and-refactoring
 **Verify:** browser-testing-with-devtools, debugging-and-error-recovery
 **Review:** code-review-and-quality, code-simplification, security-and-hardening, performance-optimization
 **Ship:** git-workflow-and-versioning, ci-cd-and-automation, deprecation-and-migration, documentation-and-adrs, observability-and-instrumentation, shipping-and-launch
--- a/README.md
+++ b/README.md
@@ -184,6 +184,7 @@ The commands above are entry points. The pack includes 24 skills total — 23 li
 | [doubt-driven-development](skills/doubt-driven-development/SKILL.md) | Adversarial fresh-context review of every non-trivial decision in-flight - CLAIM → EXTRACT → DOUBT → RECONCILE → STOP, with optional user-authorized cross-model escalation | Stakes are high (production, security, irreversible), working in unfamiliar code, or a confident output is cheaper to verify now than to debug later |
 | [frontend-ui-engineering](skills/frontend-ui-engineering/SKILL.md) | Component architecture, design systems, state management, responsive design, WCAG 2.1 AA accessibility | Building or modifying user-facing interfaces |
 | [api-and-interface-design](skills/api-and-interface-design/SKILL.md) | Contract-first design, Hyrum's Law, One-Version Rule, error semantics, boundary validation | Designing APIs, module boundaries, or public interfaces |
+| [legacy-code-and-refactoring](skills/legacy-code-and-refactoring/SKILL.md) | Characterization tests, seams, sprout method, strangler fig - never refactor and change behavior in the same commit | Modifying untested or unfamiliar code, or tempted to rewrite from scratch |
 
 ### Verify - Prove it works
 
@@ -287,6 +288,7 @@ agent-skills/
 │   ├── frontend-ui-engineering/       #   Build
 │   ├── test-driven-development/       #   Build
 │   ├── api-and-interface-design/      #   Build
+│   ├── legacy-code-and-refactoring/   #   Build
 │   ├── browser-testing-with-devtools/ #   Verify
 │   ├── debugging-and-error-recovery/  #   Verify
 │   ├── code-review-and-quality/       #   Review
--- a/README.md
+++ b/README.md
@@ -149,9 +149,9 @@ Skills are plain Markdown - they work with any agent that accepts system prompts
 
 ---
 
-## All 24 Skills
+## All 25 Skills
 
-The commands above are entry points. The pack includes 24 skills total — 23 lifecycle skills plus the `using-agent-skills` meta-skill. Each skill is a structured workflow with steps, verification gates, and anti-rationalization tables. You can also reference any skill directly.
+The commands above are entry points. The pack includes 25 skills total — 24 lifecycle skills plus the `using-agent-skills` meta-skill. Each skill is a structured workflow with steps, verification gates, and anti-rationalization tables. You can also reference any skill directly.
 
 ### Meta - Discover which skill applies
 
@@ -276,7 +276,7 @@ Every skill follows a consistent anatomy:
 
 ```
 agent-skills/
-├── skills/                            # 24 skills (23 lifecycle + 1 meta)
+├── skills/                            # 25 skills (24 lifecycle + 1 meta)
 │   ├── interview-me/                  #   Define
 │   ├── idea-refine/                   #   Define
 │   ├── spec-driven-development/       #   Define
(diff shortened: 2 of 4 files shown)


nucliweb:
Strong skill, well grounded in the Feathers playbook. The characterization test workflow (write a deliberately wrong assertion, run it, copy the actual value) is concrete enough for an agent to follow without interpretation. The seam example with `deps = { ... }` defaults is clean and correctly shows existing callers unchanged. The commit protocol rule ("if you had to edit a test expectation, it wasn't a refactor") is precise and easy to verify. The cross-references to `api-and-interface-design` (Hyrum's Law) and `code-simplification` (Chesterton's Fence) are well placed.

One structural fix: `## Core Process` should be `## Process` to match the skill anatomy convention. CLAUDE.md defines the required sections as: Overview, When to Use, Process, Common Rationalizations, Red Flags, Verification.

Also: the skill is not added to the numbered lifecycle sequence in `using-agent-skills`, and that's the right call. Legacy work is situationally triggered, not a sequential step in a feature lifecycle. The discovery tree and Quick Reference table are the right entry points.

federicobartoli:
> Strong skill, well grounded in the Feathers playbook. The characterization test workflow (write a deliberately wrong assertion, run it, copy the actual value) is concrete enough for an agent to follow without interpretation. The seam example with `deps = { ... }` defaults is clean and correctly shows existing callers unchanged. The commit protocol rule ("if you had to edit a test expectation, it wasn't a refactor") is precise and easy to verify. The cross-references to `api-and-interface-design` (Hyrum's Law) and `code-simplification` (Chesterton's Fence) are well placed.
> 
> One structural fix: `## Core Process` should be `## Process` to match the skill anatomy convention. CLAUDE.md defines the required sections as: Overview, When to Use, Process, Common Rationalizations, Red Flags, Verification.
> 
> Also: the skill is not added to the numbered lifecycle sequence in `using-agent-skills`, and that's the right call. Legacy work is situationally triggered, not a sequential step in a feature lifecycle. The discovery tree and Quick Reference table are the right entry points.

Thanks @nucliweb ! Renamed to ## Process in 15c2cd6.
And glad the lifecycle omission reads as intentional, because it was! 
Heads up: this and #223 both touch the skill-count lines in README and the using-agent-skills tables, so whichever merges second will need a trivial rebase (24 => 25). Happy to rebase mine once the other lands.

addyosmani:
This is nice work. Honestly, the way you split the commits between the refactor and the behavior change is my favorite part.

I did notice one real gap in step 5, though. The "run both, log diffs" approach is a classic Scientist-style shadow pattern, but it's only safe if the legacy path doesn't have side effects, right?

If someone runs both paths on code that writes to a database, charges a card, or sends an email, they're going to double those side effects in production. It’s definitely worth adding a quick warning sentence: *shadow the read paths, but for writes, either suppress the new path's side effects or compare against a dry run.* We just want to make sure nobody accidentally double-charges a customer while trying to "compare outputs."

One other small thing: the text leans pretty heavily on inline references to other skills (about ten of them, like `code-simplification`, `doubt-driven-development`, `api-and-interface-design`, etc.). 

Individually they all make sense, but stacked together mid-process, it starts to feel like a "go read six other skills first" homework assignment. 

That sort of undercuts the tight, self-contained scope we want agents to maintain. 

I'd suggest keeping just the most important, load-bearing ones inline (like `test-driven-development` for the fix and `git-workflow` for the commit split) and moving the rest to a quick "Related Skills" section at the very end.

Everything else looks totally merge-ready, and I appreciate you already flagging the #223 rebase for the skill count. Nice job!

federicobartoli:
> This is nice work. Honestly, the way you split the commits between the refactor and the behavior change is my favorite part.
> 
> I did notice one real gap in step 5, though. The "run both, log diffs" approach is a classic Scientist-style shadow pattern, but it's only safe if the legacy path doesn't have side effects, right?
> 
> If someone runs both paths on code that writes to a database, charges a card, or sends an email, they're going to double those side effects in production. It’s definitely worth adding a quick warning sentence: _shadow the read paths, but for writes, either suppress the new path's side effects or compare against a dry run._ We just want to make sure nobody accidentally double-charges a customer while trying to "compare outputs."
> 
> One other small thing: the text leans pretty heavily on inline references to other skills (about ten of them, like `code-simplification`, `doubt-driven-development`, `api-and-interface-design`, etc.).
> 
> Individually they all make sense, but stacked together mid-process, it starts to feel like a "go read six other skills first" homework assignment.
> 
> That sort of undercuts the tight, self-contained scope we want agents to maintain.
> 
> I'd suggest keeping just the most important, load-bearing ones inline (like `test-driven-development` for the fix and `git-workflow` for the commit split) and moving the rest to a quick "Related Skills" section at the very end.
> 
> Everything else looks totally merge-ready, and I appreciate you already flagging the #223 rebase for the skill count. Nice job!

Thanks Addy! Fixed in 27ee8b3 and a88b2f0.
You were right about the shadow comparison. The list step now says reads only, and the warning below it explains what to do for writes: run the new path in dry-run, or swap its DB/payment/mail client for one that records calls instead of executing them, then compare the recorded writes with the real ones. While I was there I also added a note about normalizing timestamps and generated IDs before diffing (otherwise everything looks like a mismatch) and about sampling the shadow traffic, since even pure reads double the load on the backend.
For the cross references I kept tdd and git-workflow inline and moved the other six to a Related Skills section at the end. I left the three in "When to Use" though, since those are exclusions telling you when not to use the skill rather than mid-process reading. Happy to move those too if you prefer.

addyosmani:
This is exactly the fix, and then some. The shadow-writes warning covers the gap I flagged, and the two you layered on top are the ones people usually learn the hard way: normalizing timestamps and IDs before diffing, and sampling the shadow load because even reads double backend traffic. To your open question, leave the three references in When to Use. Those are exclusions ("use code-simplification / deprecation-and-migration / debugging-and-error-recovery instead"), so they route the agent to the right skill rather than assign mid-process homework. Moving them would bury a signal that belongs exactly where a reader decides whether this skill applies at all.

The Related Skills split reads better too: TDD and git-workflow stay inline where the process leans on them, and the rest sit in the footer with a one-line why each, which is more useful than the bare inline mentions were. Nothing left on my side, so this is an approve. The only tie is the skill-count line you share with #223, so whichever lands second takes the trivial rebase. Great addition.

federicobartoli:
> This is exactly the fix, and then some. The shadow-writes warning covers the gap I flagged, and the two you layered on top are the ones people usually learn the hard way: normalizing timestamps and IDs before diffing, and sampling the shadow load because even reads double backend traffic. To your open question, leave the three references in When to Use. Those are exclusions ("use code-simplification / deprecation-and-migration / debugging-and-error-recovery instead"), so they route the agent to the right skill rather than assign mid-process homework. Moving them would bury a signal that belongs exactly where a reader decides whether this skill applies at all.
> 
> The Related Skills split reads better too: TDD and git-workflow stay inline where the process leans on them, and the rest sit in the footer with a one-line why each, which is more useful than the bare inline mentions were. Nothing left on my side, so this is an approve. The only tie is the skill-count line you share with #223, so whichever lands second takes the trivial rebase. Great addition.

Thanks Addy! 
Since #223 landed first, the rebase is on me: branch is now on current main with the skill count bumped to 25 in 830f1b1. That commit also fixes the README count line, which still said 24 on main after observability merged.

addyosmani:
I've read this one closely and it's genuinely good. The characterization-test-first backbone is the right approach for legacy code and it doesn't duplicate anything we have. The only reason it isn't merged already is that a new skill is a catalog decision rather than a quick yes, and I've been batching those. Consider it approved on merit, I just want to land new skills deliberately. It's near the front of that queue.

addyosmani (approved):
Really like this, and it fills a real gap: code-simplification is about clarity, this is about changing untested code safely (characterization tests, seams, strangler fig, never refactor and change behavior together). Two things before it merges: it's from June so the skill counts and the CLAUDE.md list have drifted and it needs a rebase, and please confirm there's an evals/cases/legacy-code-and-refactoring.json since CI requires one per skill. Approving on the content.

federicobartoli:
Rebased onto main at dc27a9c. The count-bump commit from the earlier rebase became empty against main and was dropped; the four content commits are unchanged.

Eval case added at evals/cases/legacy-code-and-refactoring.json: 5 positive triggers, 3 negative triggers with owners (code-simplification, test-driven-development, deprecation-and-migration), and two execution evals backed by a small untested Node fixture under evals/fixtures/legacy-code-and-refactoring/ (a pricing module with a hidden date dependency and a requested discount rule, so a seam is the natural first move).

One line changed in the skill itself: the description now names characterization tests and seams. The routing run showed "Add characterization tests before I touch this function" ranking 4th because the description did not contain the term, so I followed the evals README and fixed the description rather than the prompt. If you would rather keep the description as approved and drop that prompt instead, that is a one-line revert.

Counts moved to 26 in README, CLAUDE.md, using-agent-skills, the skill-gap issue template and the three setup docs that state the total. Routing run at 100% rank-1 (93/93), no collision warnings; validators, eval runner tests and hook tests all green locally.

PR #228, opened 2026-06-10
fix: quote SessionStart hook path

### LINKED ISSUE #214: SessionStart hook fails when plugin path contains a space (unquoted ${CLAUDE_PLUGIN_ROOT})
@
### Description

The `SessionStart` hook fails on startup when the plugin is installed under a path containing a space (e.g. a Windows home folder like `C:\Users\Win 10\`).

### Error

```
SessionStart:startup hook error
Failed with non-blocking status code: /c/Users/Win: /c/Users/Win: Is a directory
```

### Root cause

In `hooks/hooks.json` the command is not quoted:

```json
"command": "bash ${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh"
```

When `${CLAUDE_PLUGIN_ROOT}` expands to a path containing a space (e.g. `/c/Users/Win 10/.claude/plugins/...`), the shell word-splits it and `bash` receives `/c/Users/Win` as the script argument, producing the `Is a directory` error.

### Fix

Quote the path:

```json
"command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh\""
```

The internal `session-start.sh` already quotes its paths correctly (`SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"`), so only `hooks.json` needs the fix.

### Environment

- OS: Windows 10 (Git Bash)
- Home folder with space: `C:\Users\Win 10\`
- Claude Code v2.1.159
@

Bortlesboat:
## Summary
- quote the `CLAUDE_PLUGIN_ROOT` expansion in the SessionStart hook command
- add a regression check that locks the hook command string in `hooks/hooks.json`

## Verification
- `hooks/session-start-test.sh` passed from an LF-normalized temp checkout on Windows
- executed the SessionStart hook command from a temp path containing spaces and confirmed it returned the expected JSON payload

Closes #214.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -5,7 +5,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash ${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh"
+            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh\""
           }
         ]
       }
--- a/hooks/session-start-test.sh
+++ b/hooks/session-start-test.sh
@@ -11,6 +11,13 @@ if command -v jq >/dev/null 2>&1; then
   has_jq=1
 fi
 
+expected_command='bash "${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh"'
+actual_command="$(node -e "const fs=require('fs'); const hooks=JSON.parse(fs.readFileSync('hooks/hooks.json','utf8')); process.stdout.write(hooks.hooks.SessionStart[0].hooks[0].command);")"
+if [ "$actual_command" != "$expected_command" ]; then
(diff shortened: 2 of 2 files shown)


addyosmani:
Leaving as draft. For when you resume: quoting the SessionStart hook path is a legit fix for spaced install paths, and there's an open issue for exactly that Windows case, so this is worth finishing. Rebase against the current hook, confirm the payload test still passes, and flag me when it's ready.

Bortlesboat:
Closed as superseded by upstream commit 212e7a1 (with the follow-up formatting fix in b783377). The current command stores both candidate hook paths in a quoted SCRIPT variable and invokes bash "$SCRIPT", so it covers the original space-safe path fix while also supporting manual installations. No remaining change from this PR needs to land.

PR #232, opened 2026-06-10
feat: add Agent Resilience and Fuzzing skill

LorenzaVolponi:
🚀 Summary
This PR adds a new skill agent-resilience-and-fuzzing that guides agents through resilience and fuzzing testing for any skill, plus a helper script (fuzz-skill.sh) to automate simple fuzz tests. 🛡️

💡 Motivation
While agent-skills already covers many quality dimensions (TDD, code review, security, etc.), there is no explicit workflow for ensuring that skills themselves are resilient to unexpected or adversarial inputs. 📉

As agents get more autonomous, a malformed input or an unexpected API response can cascade into a broken state. This skill:

🛡️ Defines a concrete process for identifying and testing edge cases in any skill.
🤖 Ships a generic fuzzing helper script that can be reused across the repo.
🧩 Integrates with the existing skill anatomy by adding fuzz-cases.md, RESILIENCE.md, and resilience-related verification steps.

🧪 How to test

Run the included script against an existing skill:
./skills/agent-resilience-and-fuzzing/scripts/fuzz-skill.sh \  --skill skills/incremental-implementation/SKILL.md \  --arg "limit" \  --type number \  --fuzz-type boundary,mismatch,overflow

Verify that the script generates and runs fuzz cases, clearly showing which cases are OK vs FAILED. ✅

✅ Checklist

New skill under skills/ with kebab-case name
 SKILL.md follows docs/skill-anatomy.md
 YAML frontmatter with valid name and description
 Description starts with what the skill does, then Use when triggers
 Skill is specific, verifiable, and minimal
 No duplicate content; other skills are referenced by name
 No empty scripts/ directory; this skill ships a concrete, usable script

### DIFF
--- a/skills/agent-resilience-and-fuzzing/SKILL.md
+++ b/skills/agent-resilience-and-fuzzing/SKILL.md
@@ -1,0 +1,61 @@
+```markdown
+---
+name: agent-resilience-and-fuzzing
+description: Guides agents through resilience and fuzzing testing for any skill. Use when you have finished implementing a skill and want to ensure it handles edge cases, invalid inputs, and adversarial interactions gracefully.
+---
+
+# Agent Resilience and Fuzzing
+
+## Overview
+This skill ensures that any skill is production-ready by systematically testing its resilience to unexpected or adversarial inputs. Instead of assuming “it works on the happy path”, the agent follows a concrete fuzzing workflow to discover failures, ambiguous behaviors, and security-relevant edge cases.
+
+## When to Use
+- Use when you have finished implementing a new skill or significantly changed an existing one.
+- Use when a skill interacts with external systems (APIs, CLIs, databases) or receives untrusted input.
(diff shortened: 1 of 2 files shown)


skills/agent-resilience-and-fuzzing/SKILL.md
```diff
@@ -0,0 +1,61 @@
+```markdown
```
nucliweb, line 1: A SKILL cannot be placed within a Markdown block

addyosmani:
Closing this. The skill writes content into other skills' directories, which breaks the one-skill-per-directory convention the validator enforces, and the frontmatter doesn't parse cleanly, so it can't merge as structured. Resilience and fuzzing could be a legit topic, but it would need to be a single self-contained skill. Happy to see a focused version, and thanks for the contribution.

PR #236, opened 2026-06-10
docs: add optional references, examples, and resources directories to skill anatomy

nucliweb:
This PR proposes updating the skill documentation to explicitly specify `references/`, `examples/`, and `resources/` as optional subdirectories inside a skill directory under `skills/`.

### Changes:
- **[docs/skill-anatomy.md](file:///Users/joanleon/projects/nucliweb/GitHub/addy-agent-skills/docs/skill-anatomy.md)**: Updated the directory layout structure and naming conventions section to include local skill-specific references/examples/resources directories.
- **[AGENTS.md](file:///Users/joanleon/projects/nucliweb/GitHub/addy-agent-skills/AGENTS.md)**: Standardized the new skill creation structure with optional reference directories.
- **[CLAUDE.md](file:///Users/joanleon/projects/nucliweb/GitHub/addy-agent-skills/CLAUDE.md)**: Clarified that shared references live in the root `/references` folder while skill-specific ones live in `skills/<name>/references/`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -31,7 +31,7 @@ docs/         → Setup guides for different tools
 - YAML frontmatter with `name` and `description` fields
 - Description starts with what the skill does (third person), followed by trigger conditions ("Use when...")
 - Every skill has: Overview, When to Use, Process, Common Rationalizations, Red Flags, Verification
-- References are in `references/`, not inside skill directories
+- Shared references are in the root `references/` directory; the emerging convention for self-contained, distributable skills keeps a skill's own references inside `skills/<name>/references/`
 - Supporting files only created when content exceeds 100 lines
 
 ## Contributing
--- a/docs/skill-anatomy.md
+++ b/docs/skill-anatomy.md
@@ -154,7 +154,7 @@ When a skill ships runnable helpers under `scripts/`, each script follows these 
 - Skill files: `SKILL.md` (always uppercase)
 - Supporting files: `lowercase-hyphen-separated.md`
 - Shared references: stored in the root `references/` directory, not inside skill directories (see [Shared References](#shared-references) for why).
-- Skill-specific references: the emerging convention for self-contained, distributable skills is a `references/` directory inside the skill directory, so the skill carries its own supporting docs.
+- Skill-specific references: a single supporting doc can stay as a loose file in the skill directory (the `Supporting files` entry above); when several related docs travel with the skill, the emerging convention for self-contained, distributable skills is to group them in a `references/` directory inside the skill directory, so the skill carries its own supporting docs.
 
 ## Cross-Skill References
 
--- a/docs/skill-anatomy.md
+++ b/docs/skill-anatomy.md
@@ -11,10 +11,11 @@ skills/
   skill-name/
     SKILL.md           # Required: The skill definition
     scripts/           # Optional: Runnable helpers used by the skill workflow
+    references/        # Optional: Skill-specific reference documentation
     supporting-file.md # Optional: Reference material loaded on demand
 ```
 
-`SKILL.md` is the only required file. Add `scripts/` only when the skill actually ships runnable helpers, and omit the directory entirely for markdown-only skills.
+`SKILL.md` is the only required file. Add `scripts/` or `references/` only when the skill actually needs them, and omit them entirely for simpler skills.
 
 ## SKILL.md Format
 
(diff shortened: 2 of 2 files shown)


nucliweb:

addyosmani:
This conflicts with main now after the recent CONTRIBUTING and skill-anatomy changes, so it needs a rebase before I can even see it clean. Once it's refreshed I'll take a proper look, since the convention itself (optional references / examples / resources dirs) is reasonable. Rebase against current main and I'll pick it back up.

nucliweb:
Rebased onto current `main` (`16206b5`), conflicts resolved, now MERGEABLE.

The only conflict was in `AGENTS.md`: since this PR was opened, `main` replaced the skill-structure tree there with prose that points to `docs/skill-anatomy.md` as the single source of truth ("Do not restate that guidance here, link to it"). I resolved it in favor of `main` and dropped this PR's re-addition of the tree in `AGENTS.md`, so the optional-directories change now lives only where it belongs: the tree and Naming Conventions in `docs/skill-anatomy.md`, plus the one-line convention in `CLAUDE.md`. Net diff is +7/-3 across those two files. `node scripts/validate-skills.js` passes (24 skills, 0 errors, 0 warnings).

Worth noting this overlaps the discussion in #404 and #401 on shared vs skill-specific references: this PR documents where skill-specific references live, which is consistent with the root-`references/` = shared, skill-directory = single-skill split. Happy to align wording further if #401 lands first.

Ready for another look, @addyosmani.

addyosmani:
Thanks for the rebase, it's clean now. Two things before I take it.

Drop `examples/` and `resources/`. No skill in the repo uses either directory, and nothing in the current portability discussion asks for them, so sanctioning three optional dirs when only one is real is the speculative surface I'd rather keep out of the anatomy doc. Document skill-local `references/` and stop there. Second, the skill-local `references/` convention is itself the mechanism still being decided on #361 (federico flagged the same open question on #401), so stating it in CLAUDE.md as settled convention gets slightly ahead of that call. Frame it as the emerging convention for self-contained, distributable skills and keep the wording consistent with what #401 just landed. Trim to `references/` and align it, and I'll merge.

nucliweb:
Done, pushed as `f01aeda`.

- **Dropped `examples/` and `resources/`.** Both the tree and the prose in `skill-anatomy.md` now list only `scripts/` and `references/`, and the Naming Conventions entry documents skill-local `references/` and stops there. Net diff is down to +5/-3 across `skill-anatomy.md` and `CLAUDE.md`.
- **Framed skill-local `references/` as emerging, not settled.** Both `skill-anatomy.md` and `CLAUDE.md` now describe it as "the emerging convention for self-contained, distributable skills" rather than a fixed rule, so it does not get ahead of the #361 decision.

One heads-up: #401 is still open, not merged, so `main` still carries the pre-#401 wording ("References: stored in `references/` at the project root, not inside skill directories"). I aligned with #401's *direction* (shared in root, skill-local as the emerging convention) rather than its exact text, since there is no landed wording to match yet. If #401 merges first I am happy to do a quick pass to make the phrasing identical.

I also retitled the commit to "document skill-local references/ directory in skill anatomy" so it matches the trimmed scope.

federicobartoli:
Heads up @nucliweb , #401 just landed (that was me merging it), so this is back in conflict since both touch the References bullet in skill-anatomy.md. Sorry for the extra churn.

On the plus side, the wording you wanted to match is now on main, so the quick alignment pass you offered should be straightforward. From Addy's earlier comment it sounds like he'll take it from there once the phrasing lines up. Anything I can do to help on my side?

nucliweb:
Rebased onto current `main` and resolved the conflict in `8fd2081`.

The only conflict was the References bullet in `skill-anatomy.md` that #401 also touched. Aligned it as you suggested: the *shared references* bullet now keeps the `(see [Shared References](#shared-references) for why)` anchor #401 added, and the *skill-specific references* bullet stays to document the emerging skill-local `references/` convention. Nothing contradicts the new Shared References section, it just adds the per-skill case alongside it.

Back to MERGEABLE. Thanks for the heads up on the churn, @federicobartoli, no worries at all. @addyosmani this should be ready for a look whenever you have a moment.

federicobartoli (approved):

addyosmani (approved):
This is exactly what I asked for. `examples/` and `resources/` are gone, only `references/` stays, and framing it as the emerging convention rather than settled is the right call while #361 is still open. Keeping #401's Shared References anchor on the shared bullet so the two read as one story instead of fighting is a nice touch.

One honest nit, not blocking. Between #401's "a supporting file inside that skill's directory", the existing Naming entry ("Supporting files: `lowercase-hyphen-separated.md`"), and your new skill-local `references/` subdir, a contributor now has a few framings of where a skill's own doc goes. They don't contradict, but a half-sentence on when it's a single loose file versus a `references/` subdir (grouping several related docs) would tidy it. Take it or leave it. Approving either way, and it's docs-only so nothing downstream moves.

nucliweb:
Good call, added that distinction in `4df5698`: a single supporting doc stays a loose file, and the skill-local `references/` subdir is for when several related docs travel with the skill. Ties it back to the existing `Supporting files` entry so the three framings read as one. Thanks for the review.

PR #237, opened 2026-06-11
Add handoff skill for session continuity

Mrbaeksang:
## What

Adds a new cross-cutting **`handoff`** skill that compacts the current conversation into a resume-ready document, so another agent or a future session can continue the work without re-deriving context.

This fills a genuine gap in the pack: there is currently no skill for preserving work state across sessions / `compact` / agent handoffs. It complements `context-engineering` (which loads context *in*) by handling the *out* direction.

## Why a skill, not prose

It follows the standard anatomy from `docs/skill-anatomy.md`:

- **Overview / When to Use** — including explicit "When NOT to use" (task complete → write a commit/PR instead; everything already in durable artifacts → reference them)
- **Process** — choose an OS temp-dir destination (not the workspace), capture only conversation-local state, reference durable artifacts instead of duplicating them, redact secrets/PII, tailor to the next session's focus
- **Common Rationalizations / Red Flags** — anti-patterns like transcribing instead of compacting, committing the doc into the repo, omitting rejected approaches
- **Verification** — evidence checklist

Adapted from Matt Pocock's `handoff` skill, reworked into a model-agnostic, process-driven form (attribution noted in the skill).

## Integration

- `skills/handoff/SKILL.md` (new)
- `using-agent-skills`: discovery flowchart branch, lifecycle note, quick-reference row
- `README.md`: skill count 24 → 25, skill table row, project tree
- `CLAUDE.md`: cross-cutting phase entry
- `AGENTS.md`: intent → skill mapping

## Verification

Per `CONTRIBUTING.md`, this PR touches `skills/using-agent-skills/SKILL.md` (embedded by the session-start hook), so the hook regression test was run:

```
$ bash hooks/session-start-test.sh
session-start JSON payload OK
```

Frontmatter validated: `name` matches the directory, `description` is what-then-"Use when" and is 320 chars (< 1024).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -28,6 +28,7 @@ The agent should automatically map user intent to skills:
 - Refactoring / simplification → `code-simplification`
 - API or interface design → `api-and-interface-design`
 - UI work → `frontend-ui-engineering`
+- Running low on context / ending a session / passing work to another agent → `handoff`
 
 ### Lifecycle Mapping (Implicit Commands)
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -21,6 +21,7 @@ docs/         → Setup guides for different tools
 **Verify:** browser-testing-with-devtools, debugging-and-error-recovery
 **Review:** code-review-and-quality, code-simplification, security-and-hardening, performance-optimization
 **Ship:** git-workflow-and-versioning, ci-cd-and-automation, deprecation-and-migration, documentation-and-adrs, observability-and-instrumentation, shipping-and-launch
+**Cross-cutting:** handoff (compact a session into a resume-ready doc for the next agent)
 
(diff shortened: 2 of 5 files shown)


addyosmani:
Session continuity is top of my list, so I'm glad you took a run at it. The reason I'm not merging yet is that handoff, session-quality-gate, and the eval-learn proposals are all circling the same "carry state across sessions" problem, and I want to design that as one coherent thing rather than merge the first arrival. Your framing is a strong input to that. Let me get the shape right and I'll build on this or fold it in with credit.

addyosmani:
Thanks for this. The handoff to learning to durable-memory pipeline is being built in repository-memory (#429, from #398), so this overlaps it. Closing in favor of that, the ideas are welcome as comments there.

PR #239, opened 2026-06-11
fix(browser-testing): default to isolated profile, demote autoConnect

federicobartoli:
The recommended config attached the agent to the user's running Chrome via --autoConnect ('recommended for most users'), giving it every open window of the default profile: logged-in email, banking, GitHub sessions. Combined with a page containing injected instructions, the agent effectively holds the user's authenticated browser. The skill's own Security Boundaries section treats browser content as untrusted, but profile isolation is the layer that bounds the damage when those rules fail.

Changes:
- Recommended config now uses --isolated (temporary profile, wiped on close); documents that the server's default is already a dedicated profile separate from the user's browser
- --autoConnect described accurately (Chrome 144+, requires enabling remote debugging) and demoted to the logged-in-state use case
- New 'Profile Isolation' subsection in Security Boundaries with escalation rules: dedicated profile -> test-only profile -> real profile only with unrelated tabs closed
- Red flag added for attaching to the daily profile for localhost tests

### DIFF
--- a/skills/browser-testing-with-devtools/SKILL.md
+++ b/skills/browser-testing-with-devtools/SKILL.md
@@ -32,13 +32,15 @@ Add the following to your project's `.mcp.json` or Claude Code settings:
   "mcpServers": {
     "chrome-devtools": {
       "command": "npx",
-      "args": ["-y", "chrome-devtools-mcp@latest", "--autoConnect"]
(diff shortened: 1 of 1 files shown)


addyosmani (approved):
LGTM

PR #240, opened 2026-06-11
docs: add observability-checklist reference (follow-up to #223)

federicobartoli:
Follow-up suggested in the [#223 review](https://github.com/addyosmani/agent-skills/pull/223#issuecomment-4675865761): folds in the useful half of #61 — the observability checklist — as the skill's quick-reference companion, mirroring the security-and-hardening / security-checklist pairing.
Rewritten to match the merged skill's framing rather than copied verbatim: starts from on-call questions, RED/USE instead of per-endpoint prescriptions, cardinality rules (including #61's tenant_id example), symptom-based alerting with the page/ticket two-severity rule, a "verify the telemetry itself" section, and the pre-launch instrumentation gate (launch-day monitoring itself stays with shipping-and-launch, per the skill's NOT-for).

Registered in the README references table (4 → 5) and CLAUDE.md; linked from the skill's Verification section. hooks/session-start-test.sh passes. No skill-count changes, so no overlap with #224.


nucliweb:
LGTM. The cardinality section with explicit negative examples (no user_id, no raw URL, no error text as labels) is the kind of guardrail that prevents the most common mistake in this area.

addyosmani:
LGTM. Just sorted out the merge conflicts and will merge now.

PR #241, opened 2026-06-11
Expand api-and-interface-design with HTTP semantics (resolves #16)

### LINKED ISSUE #16: Suggestions for expanding api-and-interface-design skill
Hi Addy, hope you're well! I'm a big fan of your work — I've been using your [web-quality-skills](https://github.com/addyosmani/web-quality-skills) in my setup for a while now.

I recently adapted your [api-and-interface-design skill](https://github.com/addyosmani/agent-skills/blob/main/skills/api-and-interface-design/SKILL.md) into [my dotfiles](https://github.com/citypaul/.dotfiles) and ended up expanding it quite a bit in the process. I wanted to share what I added in case any of it might be useful for the original — totally understand if it doesn't fit your vision for the skill!

**1. Error format guidance with RFC 9457** — Rather than prescribing RFC 9457 (Problem Details) universally, the skill now helps you choose: RFC 9457 (`application/problem+json`) for public APIs with external consumers, or a simpler consistent shape for internal APIs with a single frontend. The key requirement is consistency across endpoints, not a specific format. RFC 9457 is still covered in depth — standard members (`type`, `title`, `status`, `detail`, `instance`), forward-compatible extension members, and the security considerations from §5 — but framed as the gold standard for public APIs rather than a universal mandate. Also includes guidance on when *not* to use problem details. ([See implementation](https://github.com/citypaul/.dotfiles/blob/main/claude/.claude/skills/api-design/SKILL.md#consistent-error-semantics))

**2. Idempotency** — Covers the HTTP method idempotency table, Stripe's idempotency keys pattern for POST, and making DELETE idempotent. The principle "design for at-least-once delivery" feels as fundamental as pagination for any API handling state changes. ([See implementation](https://github.com/citypaul/.dotfiles/blob/main/claude/.claude/skills/api-design/SKILL.md#idempotency))

**3. Rate Limiting** — Standard headers (`RateLimit-Limit`, `RateLimit-Remaining`, `RateLimit-Reset`) on every response, `Retry-After` with 429 responses, and the idea that rate limits are part of the API contract rather than an afterthought. ([See implementation](https://github.com/citypaul/.dotfiles/blob/main/claude/.claude/skills/api-design/SKILL.md#rate-limiting))

**4. HTTP Caching** — `Cache-Control` directives with common misconceptions (e.g., `no-cache` ≠ don't cache — it means revalidate), ETag revalidation, `Vary` header usage, and `no-store` for sensitive data. This was a gap in both our skills. ([See implementation](https://github.com/citypaul/.dotfiles/blob/main/claude/.claude/skills/api-design/SKILL.md#http-caching))

**5. API Evolution deep-dive** — Expanded "Prefer Addition Over Modification" into versioning strategies (Stripe's date-based pinning vs URL vs header), Postel's Law, Sunset/Deprecation headers (RFC 8594), enum evolution gotchas, and consumer-driven contract testing with Pact. ([See resource](https://github.com/citypaul/.dotfiles/blob/main/claude/.claude/skills/api-design/resources/api-evolution.md))

**6. API Security deep-dive** — The OWASP API Security Top 10 (2023) with TypeScript code examples, covering BOLA, mass assignment, SSRF, and others. Plus authentication pattern guidance (API keys, OAuth2+PKCE, JWT tradeoffs), browser security headers for APIs (`X-Content-Type-Options: nosniff`, `CSP: default-src 'none'`, `Referrer-Policy: no-referrer`), and TLS configuration requirements (TLS 1.2 minimum per RFC 9325/BCP 195). ([See resource](https://github.com/citypaul/.dotfiles/blob/main/claude/.claude/skills/api-design/resources/api-security.md))

**7. HTTP Fundamentals deep-dive** — Based on RFC 9205 "Building Protocols with HTTP" (BCP 56). Covers the cardinal rule (don't redefine HTTP semantics), status code discipline (don't map errors 1:1 — use general codes and put details in the body), HTTP method constraints (no semantic GET bodies, don't use OPTIONS for metadata), header design (no `X-` prefix per RFC 6648), full caching reference, browser security for APIs, and cookie/session security. ([See resource](https://github.com/citypaul/.dotfiles/blob/main/claude/.claude/skills/api-design/resources/http-fundamentals.md))

**8. Authentication & Token Security deep-dive** — Based on RFC 8725 "JWT Best Current Practices" (BCP 225) and RFC 9700 "OAuth 2.0 Security BCP" (BCP 240, January 2025). JWT coverage: algorithm allowlisting (the root cause of `alg: none` attacks), claim validation (`iss`/`sub`/`aud`/`exp`), explicit typing via `typ` header, input sanitization (`kid` as injection vector), compression oracle prevention. OAuth coverage: Authorization Code + PKCE (S256) for all client types, deprecated grants (Implicit, ROPC), exact redirect URI matching, token handling (never in query params), refresh token rotation, CSRF defense, clickjacking prevention, redirect security (303 not 307 for credential POSTs). ([See resource](https://github.com/citypaul/.dotfiles/blob/main/claude/.claude/skills/api-design/resources/auth-security.md))

**9. A few extra rationalizations** — "Retries are the client's problem", "We'll add rate limiting later", "Error messages are just for debugging".

Thanks again for open-sourcing the skills — they've been a great foundation to build on!

Mrbaeksang:
Resolves #16.

Adds the API-specific guidance @citypaul shared in #16, adapted to this repo's skill anatomy. Security/auth depth is **referenced, not duplicated**, to respect `CONTRIBUTING.md` ("Don't duplicate content between skills — reference other skills instead").

## What's added to `skills/api-and-interface-design/SKILL.md`

- **Error formats (RFC 9457)** — expands *Consistent Error Semantics*: Problem Details (`application/problem+json`) as the gold standard for public APIs, a simpler consistent shape for internal ones. The rule is consistency across endpoints, not a specific standard. Notes RFC 9457 §5 (don't leak internals).
- **API evolution** — expands *Prefer Addition Over Modification*: versioning strategies (date pinning vs URL vs header), `Deprecation`/`Sunset` headers ([RFC 8594](https://www.rfc-editor.org/rfc/rfc8594)), enum-evolution gotchas (Postel's Law), consumer-driven contract tests (Pact).
- **New `HTTP Semantics` section**
  - *Idempotency* — method table, `Idempotency-Key` for POST (Stripe pattern), idempotent DELETE
  - *Rate Limiting* — `RateLimit-*` headers, `429` + `Retry-After` as part of the contract
  - *HTTP Caching* — `Cache-Control` (`no-cache` ≠ don't cache), ETag revalidation, `Vary`, `no-store` for sensitive data
  - *Security* — points to the `security-and-hardening` skill and `references/security-checklist.md` for OWASP API Top 10 / JWT / OAuth2+PKCE; keeps only API-specific header + TLS reminders inline
- **Rationalizations** (3), **Red Flags** (4), **Verification** (5) entries for the new topics.

## Scoping decision

#16 also proposed full security, auth-token, and HTTP-fundamentals deep-dives. Inlining those would duplicate the existing `security-and-hardening` skill and `references/security-checklist.md`, so they are **referenced** instead of copied. Happy to add dedicated `references/` files for them in a follow-up if you'd prefer that over a cross-skill reference.

## Notes

- No frontmatter change; `name` unchanged. Does not touch hooks or the meta-skill, so the session-start hook test is not implicated.
- README's api row updated to list the new coverage.
- Credit to Paul Hammond ([@citypaul](https://github.com/citypaul/.dotfiles)) noted at the top of the skill and in the issue.


addyosmani:
Content-wise this is a clean win, and I verified it merges green and nudges routing in the right direction, so I want it. The one thing holding my hand is the external links: the RFC citations I'm fine with since those are the actual specs, but the link to a personal dotfiles repo is the kind of third-party mention I've been keeping out. Drop that one link (keep the thanks, lose the URL) and this goes straight in.

Mrbaeksang:
Done — dropped the dotfiles URL from the credit line, kept the thanks to Paul Hammond (@citypaul) and the #16 reference. Pushed in 939d245.

addyosmani (approved):
Strong addition and well-sourced (RFC 9457, RFC 8594, consumer-driven contract tests). It's in-scope content for the API skill and doesn't overlap the webhook section. It's from June so give it a quick rebase, but approving the content.

Mrbaeksang:
@addyosmani 
 Rebased onto the latest `main` in `ce0fed1`. All local validation passes, and the new workflow run is awaiting maintainer approval.

federicobartoli (changes requested):
Content-wise this looks great to me, and thanks for the quick rebase. One factual fix before I approve: the contract-evolution section cites RFC 8594 for both the Deprecation and Sunset headers, but RFC 8594 only defines Sunset; the Deprecation header was standardized separately as RFC 9745 (March 2025). Could you split the citation, something like "emit Deprecation ([RFC 9745]) and Sunset ([RFC 8594]) response headers"? With that one-liner fixed I'm happy to approve.

citypaul:
@federicobartoli what a great spot, nice work!

Mrbaeksang:
Fixed the citation: `Deprecation` now references RFC 9745, while `Sunset` references RFC 8594. Updated in 1d3ae52.

federicobartoli (changes requested):
Thanks for the RFC fix, that's exactly it. One more mechanical step before I can approve: #479 landed in the meantime and added a server-side idempotency section (§6, Honouring an Idempotency Key) plus entries in the same three lists you extend, so the branch now conflicts with main in Rationalizations, Red Flags and Verification. All three are append/append conflicts, keeping both sets resolves them. Two small things while you're in there: your verification item "State-changing POSTs accept an idempotency key (or are documented as unsafe to retry)" now duplicates main's "State-changing endpoints either honour an idempotency key...", so I'd drop yours; and a one-line pointer from your Idempotency subsection to §6 would help readers, since #479 was written as the server-side half of your contract-side guidance. Happy to approve as soon as it's green.

Mrbaeksang:
@federicobartoli  Resolved the #479 conflicts and merged the latest `main`. I kept both sets of entries in Rationalizations, Red Flags, and Verification, removed the duplicate `State-changing POSTs accept an idempotency key...` verification item, and linked the HTTP Semantics Idempotency subsection to §6 (`Honouring an Idempotency Key`).

The PR is mergeable again. The fresh GitHub Actions run is currently waiting for maintainer approval (`action_required`).

federicobartoli (changes requested):
Thanks for resolving the #479 conflicts. The resolution is right: both sets of entries are kept in the three lists, the duplicate verification item is gone, and the §6 pointer resolves to the heading on main. CI is green on the tip. Three small things I would fix before approving, then two optional notes.

The Security subsection points at security-and-hardening for the OWASP API Top 10 (BOLA, mass assignment), JWT algorithm allowlisting and OAuth2 + PKCE. That skill has an ownership check example that covers the idea behind BOLA, but nothing on mass assignment, JWT algorithm allowlisting or OAuth2 + PKCE, and the checklist only has JWT signature and expiry validation. It is the same problem I flagged on 
#415, a pointer to content that does not exist yet. Could the bullet keep those items as the reminder list itself and point at security-and-hardening for the general process, without saying it covers them?
"Return the standard headers" for RateLimit-Limit, RateLimit-Remaining and RateLimit-Reset: they come from the IETF httpapi draft rather than a standard, and the current draft has moved to a single RateLimit field plus RateLimit-Policy. "Commonly used" with a note that the spec is still a draft would be accurate.
The new verification item "Error responses follow one format (RFC 9457 for public APIs, or a single consistent shape)" duplicates the existing "Error responses follow a single consistent format" a few lines above. Folding the RFC mention into the existing item keeps the list tight.
Optional, take or leave: the idempotency table has no PATCH row, and this skill steers people to PATCH for updates, so a line saying PATCH is not idempotent by default (RFC 5789 §2) would close a gap an agent will hit. And Vary: Authorization is a weak example, since shared caches do not store responses to requests with Authorization unless told to (RFC 9111 §3.5) and the snippet is already private; Accept-Encoding or Accept-Language would show the point better.

One coordination note for whoever merges: 
#548 generates the README skills table from scripts/catalog.json, and that file still has this skill's old description. The two branches merge clean, but the drift check would fail on this row afterwards, so the README change needs to land in catalog.json too once #548 is in.

addyosmani (changes requested):
@federicobartoli is right, and my earlier approve was too quick, so I'm updating my position to match his. His round-3 catches are all valid:

- The Security subsection points at security-and-hardening for mass assignment, JWT algorithm allowlisting, and OAuth2 + PKCE, which that skill doesn't actually cover yet. Same dead-pointer issue Federico flagged on #415. Keep those items as the reminder list here and point at security-and-hardening for the general process only, don't say it covers them.
- The RateLimit-Limit / RateLimit-Remaining / RateLimit-Reset headers are an IETF draft (now moved to a single RateLimit field plus RateLimit-Policy), not a standard. "Commonly used, with a note the spec is still a draft" is the accurate framing.
- The new error-format verification item duplicates the existing "single consistent format" one a few lines up. Fold them.

The content is genuinely good and well-sourced once these land. Thanks for the careful read, the RFC 9745 vs 8594 catch especially, that is exactly the kind of thing a quick approve misses.

Mrbaeksang:
Addressed the latest review feedback in 09bd212.

- Fixed the security cross-reference so it no longer implies coverage that does not exist.
- Updated rate-limit guidance to reflect the current IETF draft (`RateLimit` / `RateLimit-Policy`) instead of calling the older `RateLimit-*` fields standard.
- Folded the duplicate error-format verification item into the existing check.
- Also added the optional PATCH idempotency note (RFC 5789) and improved the `Vary` example per RFC 9111.

I left the #548 catalog coordination untouched for now since that PR is still open.

Re-requesting review. Thanks.

federicobartoli (approved):
All three fixes land as I hoped, and thanks for taking the two optional notes too: the PATCH row matches RFC 5789 §2 including the If-Match advice, the Vary example now makes the point without leaning on Authorization, and the rate-limit guidance is framed as a draft with the RateLimit / RateLimit-Policy shape. One wording nit, take or leave: "Avoid using Authorization as the teaching example" reads as a note about this document rather than guidance for an API; something like "Do not rely on Vary: Authorization to keep per-user responses out of shared caches; RFC 9111 §3.5 and private/no-store govern that" would say the same thing to an agent. Approving. @addyosmani your changes-requested from the 6th is still standing, so this needs your re-approve before it can go in.

PR #244, opened 2026-06-11
feat(commands): enhance /review command with richer structure and opt…

Usama0111:
- Add effort levels (low/medium/high/ultra) for scoped reviews
- Add --fix and --comment flags for actionable output
- Include examples and output format guidance
- Reference related skills per axis for deeper dives

### DIFF
--- a/.claude/commands/review.md
+++ b/.claude/commands/review.md
@@ -4,13 +4,67 @@ description: Conduct a five-axis code review — correctness, readability, archi
 
 Invoke the agent-skills:code-review-and-quality skill.
 
-Review the current changes (staged or recent commits) across all five axes:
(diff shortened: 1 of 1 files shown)


addyosmani:
The richer structured output is a reasonable idea, but /review is one of eight commands we keep in lockstep across the tool directories, so a change to one has to land in all of them or the command-parity CI goes red. Right now this only touches the Claude version. Mirror it across the other command dirs (the validator will tell you which), and confirm it isn't just restating what code-review-and-quality already specifies.

addyosmani:
Thanks for this. The review command and code-review-and-quality already cover this ground. Closing as overlap.

PR #253, opened 2026-06-12
Add harness engineering skill

baskduf:
## Summary

- Add a new `harness-engineering` skill for durable repository rules, constraints, feedback loops, memory, evaluation, and governance.
- Update README and CLAUDE skill indexes from 24 to 25 skills.

## Why

Projects using coding agents often need durable repo-local guardrails instead of one-off prompts. This skill turns repeated agent mistakes and high-risk failure paths into project-specific instructions, checks, memory, and adoption reports while preserving the target repository's existing conventions.

## Validation

- `claude plugin validate .`
- `git diff --check`

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,7 +17,7 @@ docs/         → Setup guides for different tools
 
 **Define:** interview-me, idea-refine, spec-driven-development
 **Plan:** planning-and-task-breakdown
-**Build:** incremental-implementation, test-driven-development, context-engineering, source-driven-development, doubt-driven-development, frontend-ui-engineering, api-and-interface-design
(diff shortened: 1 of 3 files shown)


addyosmani:
Interesting, but harness engineering overlaps context-engineering more than the title suggests, since both are about setting the agent up to succeed before the work starts. Make the case in the PR for why it's a distinct skill rather than a section in context-engineering, per the pre-flight. If there's a clean boundary I'm open to it, I just want to avoid two skills teaching adjacent things.

addyosmani:
Thanks. Building agent harnesses is outside this pack's scope, which is engineering practices for using agents rather than building the runners. Closing.

PR #261, opened 2026-06-12
feat: add accessibility-engineering skill

Chukwuebuka-2003:
- New accessibility-engineering skill with 4-phase workflow (build, automated audit, manual testing, fix) and auditi-a11y.sh script
- New accessibility-auditor agent persona for WCAG 2.2 AA audits
- New /a11y command for all three platforms (Claude Code, Gemini, Antigravity)
- Updated /ship to run accessibility-auditor as a 4th parallel persona (replacing the former manual 'verify directly' TODO)
- Registered new agent in .claude-plugin/plugin.json


addyosmani:
Accessibility is a genuine gap and this is the stronger of the two a11y attempts, so I'm inclined toward it. Two things before merge: drop the committed .zip (we don't ship those, it's the thing the validator and CONTRIBUTING call out), and let me look at the /ship change on its own since it touches orchestration. Strip the zip and I'll dig into the skill itself.

Chukwuebuka-2003:
> Accessibility is a genuine gap and this is the stronger of the two a11y attempts, so I'm inclined toward it. Two things before merge: drop the committed .zip (we don't ship those, it's the thing the validator and CONTRIBUTING call out), and let me look at the /ship change on its own since it touches orchestration. Strip the zip and I'll dig into the skill itself.

Alright. I will work on resolving these

Chukwuebuka-2003:
@addyosmani I have dropped the .zip

Chukwuebuka-2003:
@addyosmani Can you review this PR?

PR #266, opened 2026-06-12
docs: fix broken relative path to testing-patterns.md

### LINKED ISSUE #265: Skill test-driven-development references a file that does not exist
Skill `test-driven-development` references a file that does not exist.

https://github.com/addyosmani/agent-skills/blob/main/skills/test-driven-development/SKILL.md

Line 347: For detailed testing patterns, examples, and anti-patterns across frameworks, see `references/testing-patterns.md`.

zeuspavilion:
Closes #265

### DIFF
--- a/skills/test-driven-development/SKILL.md
+++ b/skills/test-driven-development/SKILL.md
@@ -344,7 +344,7 @@ This separation ensures the test is written without knowledge of the fix, making
 
 ## See Also
 
-For detailed testing patterns, examples, and anti-patterns across frameworks, see `references/testing-patterns.md`.
+For detailed testing patterns, examples, and anti-patterns across frameworks, see `../../references/testing-patterns.md`.
 
 ## Common Rationalizations
 


nucliweb:
Thanks for the report! After investigating, I don't think this is a bug. Two things to unpack:

**The file exists.** `references/testing-patterns.md` is present in the repo, so the issue premise doesn't hold.

**The reference is a prose hint, not a resolvable path.** Skills are deployed in several ways: copied into `.cursor/rules/`, `.github/skills/`, pasted into system prompts, or loaded via the Claude Code plugin. In all these cases, the agent's working directory is the user's project, not the agent-skills repo. Neither `references/testing-patterns.md` nor `../../references/testing-patterns.md` resolves to an actual file in a typical deployment. The reference functions as a convention: it tells the agent or human "there's more detail in the `references/` folder of the agent-skills repo."

The proposed `../../` path would actually be worse in practice. If the skill is installed at `.cursor/rules/test-driven-development.md`, for example, `../../` points outside the project entirely.

The format `references/testing-patterns.md` is also the established convention across all skills in the project (`security-checklist.md`, `performance-checklist.md`, `accessibility-checklist.md` are all referenced the same way), so changing one in isolation would create inconsistency.

If there's appetite for making these references truly navigable, the right fix would be converting all of them to absolute GitHub URLs, so both humans and agents with web access can resolve them. But that would be a separate, coordinated change across every skill.

Closing this one for now, but happy to revisit if I've missed something.

addyosmani:
+1. Looks trivial but it's the wrong fix. It changes references/testing-patterns.md → ../../references/testing-patterns.md in one skill but all 7 skills use the bare references/x.md form, and these are inline-code pointers, not clickable links, so nothing is actually "broken." Merging it makes test-driven-development the only inconsistent skill.

PR #269, opened 2026-06-13
fix: add path fallback for manual installation in hooks.json

worldfunction4:
When users manually copy hooks to `.claude/` instead of installing via plugin or `--plugin-dir`, `${CLAUDE_PLUGIN_ROOT}` is not set and the SessionStart hook silently fails.
     This adds a fallback to `${CLAUDE_PROJECT_DIR}/.claude/hooks/` so the hook resolves in both plugin and manual install modes. If neither path exists, it silently succeeds.

Before:
 "command": "bash ${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh"
Now:
"command": "SCRIPT="${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh"; [ -f "$SCRIPT" ] || SCRIPT="${CLAUDE_PROJECT_DIR}/.claude/hooks/session-start.sh"; [ -f "$SCRIPT" ] && bash  "$SCRIPT" || true"

Resolution order:
1. `${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh` — plugin installation
2. `${CLAUDE_PROJECT_DIR}/.claude/hooks/session-start.sh` — manual copy
3.  If neither exists, silently succeed (don't block session start)

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -5,8 +5,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "SCRIPT=\"${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh\"; [ -f \"$SCRIPT\" ] || SCRIPT=\"${CLAUDE_PROJECT_DIR}/.claude/hooks/session-start.sh\"; [ -f \"$SCRIPT\" ]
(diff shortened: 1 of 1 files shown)


addyosmani:
LGTM

PR #270, opened 2026-06-13
ci: validate command descriptions stay in sync across all three tool directories

nucliweb:
## Problem

The repo ships the same slash commands in three formats for three different tools:

- `.claude/commands/` (.md) for Claude Code
- `.gemini/commands/` (.toml) for Gemini CLI
- `commands/` (.toml) for Antigravity CLI

Nothing prevented the `description` field from silently drifting between them. The three files for the same command should share an identical description, even though their prompt bodies legitimately differ (tool-specific syntax like `$ARGUMENTS`, `agent-skills:` prefixes, `GEMINI.md` vs `CLAUDE.md`).

## What this adds

`scripts/validate-commands.js` runs two checks:

1. **Parity**: every command present in one directory exists in all three (with the known name mapping `plan` in Claude vs `planning` in the TOML dirs).
2. **Description sync**: the `description` field is identical across all three equivalents for each command.

The script exits with code 1 on any failure, blocking CI. It follows the same structure as the existing `validate-skills.js`.

The `validate` job in CI now depends on both `validate-skills` and `validate-commands`, so both must pass before plugin structure validation runs.

A follow-up commit also hardens the script: `descriptionFromToml` now handles single-quoted TOML strings (previously returned `null` silently), file read errors are caught and reported as actionable CI failures instead of raw stack traces, and the summary count covers all canonical commands rather than only Claude-side ones.

## What it intentionally does not check

Prompt body differences are expected and intentional; each tool has its own syntax. Only the description field is enforced.

## Verification

```

8 commands checked, 0 error(s), PASSED
```

### DIFF
--- a/.github/workflows/test-plugin-install.yml
+++ b/.github/workflows/test-plugin-install.yml
@@ -20,9 +20,23 @@ jobs:
       - name: Validate all skills
         run: node scripts/validate-skills.js
 
+  validate-commands:
+    name: Validate command parity and description sync
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Validate commands across all tool directories
+        run: node scripts/validate-commands.js
+
   validate:
     name: Validate plugin structure
-    needs: validate-skills
+    needs: [validate-skills, validate-commands]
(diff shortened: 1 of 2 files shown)


nucliweb:
## Tests PASSED

Tests PASSED

addyosmani (approved):
passes the check on current main (8 commands, 0 errors - it even handles plan/planning and webperf), so merging it won't break CI. Landing. Thank you!

PR #36, opened 2026-04-07
feat: add context: fork to all skills and commands

nucliweb:
Closes #35

## What

Adds `context: fork` to the YAML frontmatter of ~all 20 `SKILL.md`~ files and all 7 `.claude/commands/*.md` slash commands.

## Why

Skills and commands currently inject their instructions directly into the active conversation context. With `context: fork`, each skill or command runs in an isolated subagent context instead, which brings five concrete benefits:

1. **Context isolation** — the skill runs free from prior conversation history
2. **Consistent, repeatable execution** — same clean state on every invocation
3. **Token efficiency** — execution tokens are spent in the fork, not the main session
4. **Stronger skill fidelity** — harder for the agent to rationalize skipping steps
5. **Cleaner handoff** — only the result surfaces back, not intermediate reasoning

Per the [official docs](https://code.claude.com/docs/en/slash-commands), `.claude/commands/` files ~and `SKILL.md` files~ share the same frontmatter spec and `context: fork` is valid ~in both~.

## Files changed

**Commands** (`.claude/commands/`) — 7 files  
~**Skills** (`skills/*/SKILL.md`) — 20 files~

PR #272, opened 2026-06-13
feat: Add advanced Antigravity metadata to skills

nucliweb:
This PR adds support for advanced YAML frontmatter metadata fields (kind, model, temperature, max_turns, tools) to all 24 skills, optimized specifically for Antigravity.

### Key Advantages of the Extra Fields:
1. **Safety and Least Privilege:** Restricting tools to only the required set (e.g., read-only tools for audits and reviews) ensures that agents cannot accidentally mutate code or run unauthorized shell commands during passive review stages.
2. **Context and Token Efficiency:** By limiting the available toolset on a per-skill basis, the agent avoids sending unused tool schemas back and forth, reducing token overhead.
3. **Smart Cost and Power Routing (model):** Routes lighter tasks (formatting, documentation, git commands) to gemini-3.5-flash while reserving gemini-2.5-pro for cognitively demanding tasks (debugging, security auditing, interface design).
4. **Execution Control (max_turns):** Implements hard turn limits to prevent agents from entering infinite execution loops when correcting errors.
5. **Context Isolation (kind: subagent):** Isolates heavy tasks in background subagents, keeping the parent conversation clean.

### Zero Impact on Other Agents (e.g., Claude Code, Codex, etc...):
* **Ignored Fields:** Standard parsers used by other agents only search for name and description in the YAML frontmatter. They silently ignore unrecognized fields without raising errors.
* **Progressive Disclosure:** Other agents only parse names and descriptions during the initial routing phase. The complete skill files (and their advanced metadata fields) are only loaded when a skill is active.
* **Minimal Token Overhead:** When active, the extra fields add less than 30 tokens to the prompt, representing less than 0.015% of a typical 200k context window, which has zero impact on retrieval performance.

### Implementation Checklist:
1. Updated validator script (scripts/validate-skills.js) to support parsing and enforcing constraints on these properties.
2. Updated all 24 skills with custom metadata matching their logical complexity.
3. Updated setup and anatomy documentation (docs/skill-anatomy.md and docs/antigravity-setup.md).

### DIFF
--- a/docs/antigravity-setup.md
+++ b/docs/antigravity-setup.md
@@ -104,7 +104,7 @@ You can invoke these personas directly within your session or when delegating ta
 
 Each skill is optimized for Antigravity CLI using advanced properties in its frontmatter:
 - **`kind`**: Declares whether the skill runs `inline` or spawns a background `subagent` (ideal for heavy processes to keep the parent conversation clean).
-- **`model`**: Automatically routes the skill to `gemini-2.5-flash` (saving tokens and latency) or `gemini-2.5-pro` (maximizing reasoning power).
+- **`model`**: Automatically routes the skill to `gemini-3.5-flash` (saving tokens and latency) or `gemini-2.5-pro` (maximizing reasoning power).
 - **`temperature`**: Sets model determinism (`0.0`-`0.3` for refactoring/debugging) or creativity (`0.7`-`1.0` for ideation/interviews).
 - **`max_turns`**: Halts executing agent to prevent infinite loops.
 - **`tools`**: Lists only the allowed tools (e.g. read-only tools for review/security checks), saving context window and preventing unauthorized writes.
--- a/docs/antigravity-setup.md
+++ b/docs/antigravity-setup.md
@@ -104,12 +104,12 @@ You can invoke these personas directly within your session or when delegating ta
 
 Each skill is optimized for Antigravity CLI using advanced properties in its frontmatter:
 - **`kind`**: Declares whether the skill runs `inline` or spawns a background `subagent` (ideal for heavy processes to keep the parent conversation clean).
-- **`model`**: Automatically routes the skill to `gemini-3.5-flash` (saving tokens and latency) or `gemini-2.5-pro` (maximizing reasoning power).
+- **`model`**: Automatically routes the skill to `gemini-2.5-flash` (saving tokens and latency) or `gemini-2.5-pro` (maximizing reasoning power).
 - **`temperature`**: Sets model determinism (`0.0`-`0.3` for refactoring/debugging) or creativity (`0.7`-`1.0` for ideation/interviews).
 - **`max_turns`**: Halts executing agent to prevent infinite loops.
 - **`tools`**: Lists only the allowed tools (e.g. read-only tools for review/security checks), saving context window and preventing unauthorized writes.
 
-For detailed guidelines, see [Skill Anatomy](file:///Users/joanleon/projects/nucliweb/GitHub/addy-agent-skills/docs/skill-anatomy.md).
+For detailed guidelines, see [Skill Anatomy](skill-anatomy.md).
 
 ### Project-Specific Enforcements (`AGENTS.md`)
 To enforce strict skill compliance (e.g. requiring a spec or plan before writing code), copy or link `AGENTS.md` into the root of your workspace. Antigravity CLI reads this file to align the agent's behavior and planning phase with your team's conventions.
--- a/docs/antigravity-setup.md
+++ b/docs/antigravity-setup.md
@@ -100,6 +100,17 @@ You can invoke these personas directly within your session or when delegating ta
 
 ## Configuration & Customization
 
+### Advanced Skill Metadata (YAML Frontmatter)
+
+Each skill is optimized for Antigravity CLI using advanced properties in its frontmatter:
+- **`kind`**: Declares whether the skill runs `inline` or spawns a background `subagent` (ideal for heavy processes to keep the parent conversation clean).
+- **`model`**: Automatically routes the skill to `gemini-3.5-flash` (saving tokens and latency) or `gemini-2.5-pro` (maximizing reasoning power).
+- **`temperature`**: Sets model determinism (`0.0`-`0.3` for refactoring/debugging) or creativity (`0.7`-`1.0` for ideation/interviews).
+- **`max_turns`**: Halts executing agent to prevent infinite loops.
+- **`tools`**: Lists only the allowed tools (e.g. read-only tools for review/security checks), saving context window and preventing unauthorized writes.
+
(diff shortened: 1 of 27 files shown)


nucliweb:
cc @addyosmani and @federicobartoli would love your thoughts if time allows ☺️

addyosmani:
Making very minor updates.

After reviewing: Change LGTM. This PR does adds vendor-specific frontmatter (kind/model/temperature/max_turns/tools) to every skill, which other tools (Claude/Cursor) will just ignore - totally harmless, but it's a deliberate house-style choice. Something we might just want to check we are OK with.

nucliweb:
> Making very minor updates.
> 
> After reviewing: Change LGTM. This PR does adds vendor-specific frontmatter (kind/model/temperature/max_turns/tools) to every skill, which other tools (Claude/Cursor) will just ignore - totally harmless, but it's a deliberate house-style choice. Something we might just want to check we are OK with.

Thanks @addyosmani 

What do you think @federicobartoli?

I know that’s a vendor-specific formatter, but I think it has a lot of potential; it allows for ‘surgical’ control over behaviour.

If we decide it’s best not to implement it, IMHO, we should document it so that people are aware of these options and have the opportunity to optimize how the skills work in their agents, just as I proposed in PR #36 for Claude Code.

We could have an advanced configuration document to explain the specific settings for each agent.

Having said that, I feel comfortable closing this PR, and #36, having created the document explaining it.

federicobartoli:
Thanks for looping me in @nucliweb!  I like the least-privilege and runtime-routing goals, but after checking the current specs I do not think these fields should land as top-level properties in every SKILL.md.
The Agent Skills specification reserves top-level frontmatter for name, description, license, compatibility, metadata, and experimental allowed-tools, and explicitly says vendor or client-specific properties belong under metadata. The fields added here (kind, model, temperature, max_turns, tools) are not spec fields, so the "zero impact on other agents" claim is not guaranteed: a strict spec-conformant validator or client can flag or reject them rather than silently ignoring them.
Worth noting these five fields are actually the schema for Gemini CLI subagent definitions (.gemini/agents/*.md), which are a separate resource from skills, and there kind is local or remote, not inline/subagent/workflow. So as written the frontmatter does not match the Agent Skills schema nor the Gemini subagent schema.
Two more concrete issues: Google now recommends removing temperature for all Gemini 3.x models (use thinking_level instead), so hardcoding temperature on 3.x-routed skills is already against current guidance. And the new validator only checks shape, not semantics: any non-empty string passes for model and tools, and parseInt("1.5") / parseFloat("0.2oops") slip through for max_turns / temperature.
My preference: keep the shared SKILL.md files portable and move Antigravity-specific orchestration into a separate adapter, plugin config, or under metadata, documented as an opt-in. Before adding it globally I would also want an integration test in Antigravity itself proving model selection, subagent spawning, turn limits, and tool restrictions are actually enforced.


To be clear though, I am very much aligned with your own suggestion, lets close this and #36 and fold all of it into a single "advanced per-agent configuration" doc. The detail above is not meant as a blocker on your work (which is a lot, and useful), it is exactly the material that doc should capture: the Antigravity and Claude Code fields, where they belong (under metadata or an adapter), and the current model guidance. Happy to co-write or review it with you.

nucliweb:
Thanks @federicobartoli, this is exactly the kind of review I was hoping for, and I agree with all of it. Your points are well taken:

- The fields aren't part of the Agent Skills spec, so the "zero impact" claim isn't guaranteed; vendor-specific properties belong under `metadata`, not top-level.
- The schema matches neither Agent Skills nor the Gemini CLI subagent definitions (`kind: local|remote`, not `inline|subagent|workflow`), so as written it's incorrect on both counts.
- `temperature` is already against current guidance for Gemini 3.x (`thinking_level` instead).
- The validator checks shape, not semantics, so invalid values slip through.

So my proposal: **let's close this PR and #36, and fold everything into a single "advanced per-agent configuration" doc** that keeps the shared `SKILL.md` files portable. The doc would capture:

- Which fields each agent supports (Antigravity/Gemini, Claude Code, etc.) and **where they belong** (under `metadata` or in a separate adapter, never top-level).
- Current model guidance (e.g. Gemini 3.x → `thinking_level`, not `temperature`).
- Least-privilege / runtime-routing as an **opt-in** pattern, not a global change.

One idea I'd like to explore on top of the doc: **automating it with scripts**. Instead of hand-editing per-agent metadata into each skill, a small generator could take a single source of truth and emit the agent-specific output, e.g. inject the fields under `metadata`, or produce the separate adapter files (`.gemini/agents/*.md`) where that's the right home. That keeps `SKILL.md` clean and portable while still giving people the "surgical" control I was after, and it sidesteps the drift and validation problems entirely.

I'll put together a proposal along these lines and share it for review. Thanks again @addyosmani and @federicobartoli for the careful read.

I'll hold off on closing #272 and #36 until I get an OK from both of you on this direction.

addyosmani:
Revisiting this in light of your #316, which lays out the per-agent configuration architecture. The principle in #316 is that `SKILL.md` stays portable: spec fields top-level, vendor fields under `metadata`, and runtime orchestration (model, temperature, tools, max_turns) in separate per-agent adapter files, leaving `SKILL.md` untouched.

This PR does the opposite. It puts `kind`, `model`, `temperature`, `max_turns`, and `tools` at the top level of all 24 `SKILL.md` files. Two concrete consequences:

- It couples every shared skill to one vendor's runtime model. Claude, Cursor, and the rest ignore these fields, so the canonical files carry vendor clutter that the spec reserves for `metadata`.
- It hardcodes `temperature: 0.1` on skills routed to `gemini-3.5-flash` (a 3.x model), which is the exact thing #316 says not to do (prefer `thinking_level`).

So I'd hold this and rework it to follow #316: keep the 24 `SKILL.md` files untouched and emit `.gemini/agents/*.md` adapter files instead, ideally via the generator #316 describes. If you'd rather not carry both, closing this in favor of the adapter-file approach is clean too. The intent is good; it reads better as adapters than as frontmatter on every skill.

nucliweb:
Thanks @addyosmani, fully agree. This confirms the direction I wanted, and now that you've given the OK on #316, the cleanest move is to **close #272 in favor of the adapter-file approach**.

Recap of why, so it's on the record for anyone landing here later:

- The 24 `SKILL.md` files stay **portable**: spec fields top-level, vendor config under `metadata`, and runtime orchestration (`model`, `temperature`/`thinking_level`, `tools`, `max_turns`) in per-agent adapter files, never as top-level frontmatter.
- The Antigravity/Gemini controls land as `.gemini/agents/*.md` adapters, ideally emitted by the generator described in #316, so we never hand-maintain vendor fields across 24+ skills.
- This also fixes the concrete issues raised on this PR: the schema mismatch and the hardcoded `temperature: 0.1` on 3.x-routed skills (→ `thinking_level`).

I'll fold the useful parts of this PR (the per-skill model/tool routing intent) into #316 as the generator's config map, and address your two asks there: stating explicitly that it supersedes the inline approach, and adding a copy-pasteable `thinking_level` example.

Closing this one. Thanks again to you and @federicobartoli for the careful read.

PR #276, opened 2026-06-14
fix(hooks): make simplify-ignore-test.sh executable

Dashsoap:
## What

`hooks/simplify-ignore-test.sh` is committed with file mode `100644` (non-executable), while its sibling test script `hooks/session-start-test.sh` is `100755` (executable).

This PR sets the executable bit on `simplify-ignore-test.sh` so the two test scripts match.

## Why

Both files:
- live in `hooks/`
- start with the same `#!/bin/bash` shebang
- are the project's two hook test scripts

The mode mismatch is just an oversight — `simplify-ignore-test.sh` was the only one missing the exec bit. This lets it be run directly (`./hooks/simplify-ignore-test.sh`) like its sibling, not only via `bash hooks/simplify-ignore-test.sh`.

```
mode change 100644 => 100755 hooks/simplify-ignore-test.sh
```

No content change.


addyosmani:
Thanks for the helpful PRs! This LGTM.

PR #285, opened 2026-06-14
Add reliable-agent-loops skill

vikast908:
## What

A new `reliable-agent-loops` skill on building LLM agent loops that hold up in production.

## Why

The collection already covers debugging, observability, and security, but not the agent loop itself. The three failures behind most agent-loop incidents aren't addressed anywhere in the repo: iteration that never terminates, tool errors that get swallowed, and side effects that run more than once (the classic double-charge after a retry).

## What's in it

- Bounding the loop with step and token/cost caps
- Returning tool errors to the model as structured results instead of swallowing them
- Idempotent retries (retry only retryable errors; idempotency keys on side effects)
- Gating irreversible actions behind confirmation/allowlists
- Resumable, checkpointed state so a crash doesn't replay completed side effects
- Per-run tracing

Each step has BAD/GOOD code, and the skill closes with a Verification checklist.

## Conventions followed

- Standard anatomy: Overview, When to Use, Process, Common Rationalizations, Red Flags, Verification.
- Frontmatter is `name` + `description` only; the description leads with what it does, then `Use when` triggers.
- References `security-and-hardening` (tool authz, injection through tool output) and `observability-and-instrumentation` (logging/metrics) rather than duplicating them.

Related: I've opened a second PR adding `evaluating-llm-output`, the eval counterpart for catching silent regressions. Happy to adjust naming, scope, or split anything.

### DIFF
--- a/skills/reliable-agent-loops/SKILL.md
+++ b/skills/reliable-agent-loops/SKILL.md
@@ -1,0 +1,124 @@
+---
+name: reliable-agent-loops
+description: Builds LLM agent loops that survive production. Use when implementing or modifying a model→tool→model loop, a tool-calling loop, or any multi-step autonomous workflow. Use when an agent hangs, loops forever, double-executes a side effect, blows its token budget, or loses work when a step fails.
+---
+
+# Reliable Agent Loops
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


addyosmani:
This is the anchor of a three-skill LLM-runtime cluster you've proposed, and the real question isn't quality, it's whether the pack expands into building-LLM-apps territory or stays focused on the engineering lifecycle. That's a direction call I owe you rather than leaving the three in limbo. Give me a bit to decide the scope, and if it's yes these are a strong start.

addyosmani:
Thanks for this. The pack is scoped to engineering practices for working with AI coding agents, not building LLM applications, so reliable agent loops sits outside it. Closing as out of scope.

PR #286, opened 2026-06-14
Add evaluating-llm-output skill

vikast908:
## What

A new `evaluating-llm-output` skill on building evals so LLM and agent changes don't silently regress.

## Why

The repo has `test-driven-development`, but TDD targets deterministic code. LLM output is free-form text that ordinary unit tests can't assert on, so a prompt tweak or a model bump can quietly degrade quality with nothing in CI to catch it. There's no skill covering the regression discipline for model output.

## What's in it

- A versioned golden set of representative inputs and expected outcomes
- Grading with the cheapest valid method (assertions for structured output, LLM-as-judge with a rubric for open-ended text)
- Asserting on agent behavior and tool use, not just final text
- Negative and safety cases (must-refuse, prompt injection, malformed input)
- CI gating on a pass threshold
- Tracking pass rate, cost, and latency over time, plus handling non-determinism

Each step has code, and the skill closes with a Verification checklist.

## Conventions followed

- Standard anatomy: Overview, When to Use, Process, Common Rationalizations, Red Flags, Verification.
- Frontmatter is `name` + `description` only; the description leads with what it does, then `Use when` triggers.
- References `test-driven-development` (deterministic code) and `ci-cd-and-automation` (pipeline setup) rather than duplicating them.

Related: this pairs with my other PR adding `reliable-agent-loops`. Happy to adjust naming, scope, or split anything.

### DIFF
--- a/skills/evaluating-llm-output/SKILL.md
+++ b/skills/evaluating-llm-output/SKILL.md
@@ -1,0 +1,106 @@
+---
+name: evaluating-llm-output
+description: Builds evals so LLM and agent changes can't silently regress. Use when changing a prompt, swapping or upgrading a model, or adding an LLM-powered feature. Use when you have no automated way to tell whether output quality dropped, or when "it looked fine when I tried it" is your current QA.
+---
+
+# Evaluating LLM Output
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


addyosmani:
Same note as the other two in your LLM cluster: the content isn't the question, the scope is. Three skills about operating LLM features is a real expansion of what this pack covers, and I want to make that call deliberately instead of merging them one at a time. Holding all three on the same decision, and I'll circle back rather than leave you hanging.

addyosmani:
Thanks. Same scope line: evaluating LLM output is LLM-application work rather than general engineering, so it's outside this pack. Closing.

PR #287, opened 2026-06-14
Add llm-cost-optimization skill

vikast908:
## What

A new `llm-cost-optimization` skill on cutting runtime LLM token and dollar cost without lowering output quality.

## Why

The repo has `context-engineering` (the coding agent's own session) and `performance-optimization` (web vitals and latency), but nothing on the API cost of the model calls a product makes at runtime. That cost is usually dominated by avoidable waste: a static system prompt resent uncached on every call, full history replayed each turn, whole documents stuffed into context, unbounded output, and a frontier model doing cheap work.

## What's in it

- Measuring cost per request before optimizing
- Caching the static prefix (system prompt, tool schemas, examples)
- Windowing or summarizing history instead of full replay
- Narrow retrieval instead of stuffing context
- Capping and structuring output
- Routing cheap, high-volume tasks to a smaller model
- A "don't trade away quality" guard, with a Verification checklist

## Conventions followed

- Standard anatomy (Overview, When to Use, Process, Common Rationalizations, Red Flags, Verification).
- Frontmatter is `name` + `description` only; description leads with what it does, then `Use when` triggers.
- References `context-engineering` and `performance-optimization` rather than duplicating them.

Related: this pairs with my other PRs adding `reliable-agent-loops` (#285) and `evaluating-llm-output` (#286). Happy to adjust naming, scope, or split anything.

### DIFF
--- a/skills/llm-cost-optimization/SKILL.md
+++ b/skills/llm-cost-optimization/SKILL.md
@@ -1,0 +1,116 @@
+---
+name: llm-cost-optimization
+description: Reduces token usage and LLM API cost at runtime without lowering output quality. Use when LLM spend is climbing faster than usage, when prompts or context are large, or when building a feature that calls a model in a loop or at scale. Use when cost or latency per request needs to come down.
+---
+
+# LLM Cost Optimization
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


addyosmani:
Third of your LLM-runtime trio, and I'm treating them as a set for the scope decision rather than judging each alone. The question isn't the quality, it's whether this pack grows into LLM-app operations or stays on the human-plus-agent engineering lifecycle. I owe you a clear answer on that instead of leaving three PRs in limbo. Bear with me on the direction call and I'll come back to all three together.

addyosmani:
Thanks for this. LLM cost optimization is application-building territory rather than the engineering lifecycle this pack covers. Closing as out of scope.

PR #289, opened 2026-06-14
fix(idea-refine): make idea-refine.sh executable

Dashsoap:
## What

Set the executable bit on \`skills/idea-refine/scripts/idea-refine.sh\` (100644 → 100755). Content is unchanged — this is a mode-only change.

## Why

Every other shell script in the repo is already \`100755\` — all six \`hooks/*.sh\`:

\`\`\`
100755 hooks/sdd-cache-post.sh
100755 hooks/sdd-cache-pre.sh
100755 hooks/session-start-test.sh
100755 hooks/session-start.sh
100755 hooks/simplify-ignore-test.sh
100755 hooks/simplify-ignore.sh
100644 skills/idea-refine/scripts/idea-refine.sh   <-- the odd one out
\`\`\`

\`idea-refine.sh\` has a \`#!/bin/bash\` shebang and is invoked as a runnable helper from its own \`SKILL.md\`. \`AGENTS.md\` also documents the \`scripts/\` directory as **"Required: executable scripts"**. Setting the exec bit aligns it with every other script in the repo and with the documented convention.

This mirrors the already-merged #276, which set the exec bit on \`simplify-ignore-test.sh\` for the same reason.


nucliweb:
Clean fix. The reasoning is airtight: `idea-refine.sh` has a shebang, is invoked as a runnable in `SKILL.md`, and `AGENTS.md` explicitly marks `scripts/` as *"Required: executable scripts"*. The `644` was the odd one out against the rest of the repo. Mode-only change, no content risk.

addyosmani (approved):

PR #291, opened 2026-06-15
docs: add translated README links

dowithless:
This PR adds links to translated versions of the README to help non-English-speaking users access translated documentation.

Supported languages:

- Chinese
- Japanese
- Korean
- German
- French
- Spanish
- Portuguese
- Russian

Translations are automatically synchronized and require no maintenance effort from project maintainers.

This change only affects README documentation and does not modify project code, dependencies, CI workflows, or repository settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,15 @@
 # Agent Skills
 
+<!-- Keep these links, translations synced daily. -->
+[Deutsch](https://zdoc.app/de/addyosmani/agent-skills) |
+[Español](https://zdoc.app/es/addyosmani/agent-skills) |
+[français](https://zdoc.app/fr/addyosmani/agent-skills) |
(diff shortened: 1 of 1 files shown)


nucliweb:
IMHO, if an open-source project provides documentation for non-English-speaking users, it can help attract more contributors. Over time, the community can also help maintain the translations.

It would also be a good idea to translate the contents of the `doc` folder.

That said, that’s just my opinion.

addyosmani:
Thanks @dowithless! Going to close this one for the same reason we're holding off on translated READMEs generally. These point at an external auto-translation service, and we can't commit to keeping translations accurate or vouching for a third-party service's output as the README changes, so it would risk drifting or going stale.
Appreciate you flagging it though.
Not a forever no, just not something we can maintain right now.

PR #292, opened 2026-06-16
Add web-browser-automation skill (WebCLI / real browser)

crisdosaygo:
Adds a production-grade skill for operating a real browser via the web CLI (WebCLI by DOSAYGO, https://webcli.sh). Covers the look->act->re-inspect loop, navigation, form interaction, file uploads, and human handoff for auth/CAPTCHA. Follows standard skill anatomy with Common Rationalizations and Red Flags based on real agent usage.


addyosmani:
Thanks @crisdosaygo, appreciate the thorough writeup! Going to close this one though. We keep the core pack vendor-neutral, and this wraps one specific third-party CLI (WebCLI) that we can't maintain over time, plus real-browser work is already covered by `browser-testing-with-devtools`.
It'd be a solid standalone community skill, just not a fit for the core set.
Thanks again for the effort.

PR #293, opened 2026-06-16
docs: fix step range in debugging skill (Steps 4-10 → 4-6)

shihyuho:
The Stop-the-Line Rule and Triage Checklist each have exactly 6 steps, but the text referenced "Steps 4-10" — steps that don't exist.

`A bug in Step 3 that goes unfixed makes Steps 4-10 wrong.` → `... makes Steps 4-6 wrong.`


addyosmani (approved):
TY!

PR #294, opened 2026-06-17
Harden opt-in lifecycle hooks (SSRF, cache poisoning, symlink overwrite, data loss)

bill143:
## What

Hardens the three **opt-in** lifecycle hooks (`sdd-cache-pre.sh`, `sdd-cache-post.sh`, `simplify-ignore.sh`) against security and data-integrity issues surfaced during a review. These hooks are not enabled by `hooks/hooks.json` (which only wires `session-start.sh`), so this affects users who manually enable the WebFetch cache or simplify-ignore per the docs. `session-start.sh` is unchanged (reviewed clean).

## Why / findings addressed

| Severity | Issue | Fix |
|----------|-------|-----|
| Critical | **SSRF** — `curl -L "$URL"` on an unvalidated, redirect-following URL fires automatically on every WebFetch; a 302 can reach `169.254.169.254`, `localhost`, RFC1918 | `url_is_safe()` allowlist + drop `-L` (`--proto =https --max-redirs 0`) |
| High | **Cache poisoning** — no TTL; any `304` re-serves the cached body forever as "unchanged" | TTL cap via existing `fetched_at` (`SDD_CACHE_MAX_AGE`, default 24h) |
| High | **World-readable / tamperable cache** — default perms on fetched bodies | cache dir `0700`, files `0600` via `umask 077` |
| High | **Symlink / arbitrary-file overwrite** — `cat > "$target"` on untrusted `file_path` follows symlinks; session-long TOCTOU | reject symlinks at entry + re-verify regular-file target before every write-back |
| Critical (integrity) | **Data loss** — non-atomic in-place rewrite under `set -e` can leave a half-placeholdered file after the backup was consumed | `trap`-guarded restore-from-backup around the rewrite |
| — | **Undeclared `perl` dependency** — abort point on minimal Git Bash | replaced with pure-shell trailing-newline trim (dependency removed) |
| — | **Path-identity collisions** — `C:\x` vs `/c/x` hash to different IDs → stale restore | normalize path before hashing `file_id` |

## How it's verified

- `bash -n` clean on all four hooks.
- Existing suites pass: **21/21** `simplify-ignore-test.sh`, `session-start-test.sh` OK.
- Added smoke test for the new SSRF guard: **15/15** (covers https-only, loopback, RFC1918, link-local/metadata, CGNAT, IPv6 literal, `user@host` userinfo trick, and correctly-allowed public hosts incl. just-outside-range `172.15.0.1`).

## Notes / open design decisions (flagged, not silently chosen)

- **DNS rebinding** is *not* fully closed — a public hostname resolving to a private IP still passes without resolved-IP pinning (`curl --resolve`), which adds platform variance on Git Bash. Left as a follow-up; relies on WebFetch's own egress controls for that case.
- **`--max-redirs 0`** trades cache hit-rate on redirecting origins for SSRF safety. A middle ground (`--max-redirs N --proto-redir =https` + re-validating `%{url_effective}`) is possible if hit-rate matters.
- **TTL default (24h)** is a guess; tune to how stale a cached "reading" may be.
- **`cat >` kept (not `mv`)** to preserve inode/permissions as the original intended; atomicity-of-outcome is provided by the `trap` instead.
- **`file_id` change** orphans any pre-existing cache entries from the old scheme (one-time; Stop already tolerates orphans).

Happy to split this into per-file PRs or adjust any of the above. Related findings filed as an issue for tracking.

### DIFF
--- a/hooks/sdd-cache-post.sh
+++ b/hooks/sdd-cache-post.sh
@@ -63,6 +63,30 @@ if [ -z "$CONTENT" ]; then
 fi
 dbg "extracted content bytes=${#CONTENT}"
 
+# Reject non-https URLs and internal/metadata hosts before any curl runs.
+# Mirror of the pre hook's guard. Conservative: only public https hosts pass.
+url_is_safe() {
+  local u="$1" host
+  case "$u" in
+    https://*) ;;
+    *) return 1 ;;
+  esac
+  host="${u#https://}"
+  host="${host%%/*}"; host="${host%%\?*}"; host="${host%%#*}"
+  host="${host##*@}"; host="${host%%:*}"
+  host=$(printf '%s' "$host" | tr '[:upper:]' '[:lower:]')
+  case "$host" in
+    localhost|localhost.*|*.localhost) return 1 ;;
+    127.*|0.*|10.*|169.254.*|192.168.*) return 1 ;;
+    172.1[6-9].*|172.2[0-9].*|172.3[0-1].*) return 1 ;;
+    100.6[4-9].*|100.[7-9][0-9].*|100.1[0-1][0-9].*|100.12[0-7].*) return 1 ;;
+    \[*) return 1 ;;
+    metadata|metadata.*|*.internal|*.local) return 1 ;;
+    "") return 1 ;;
+  esac
+  return 0
+}
+
 # Must match the pre hook: sha256(URL), first 32 hex chars.
 hash_key() {
   if command -v shasum >/dev/null 2>&1; then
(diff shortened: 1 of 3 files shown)


addyosmani:
This is security-sensitive by nature, so it gets a careful pass rather than a quick merge. SSRF and cache-poisoning hardening on the lifecycle hooks is worth doing, but these hooks run on every session start for every user, so I want to read the change against the existing no-jq fallback and the payload test before anything lands. Give me the time to go through it properly, and thanks for taking the hooks seriously.

federicobartoli:
Ahead of Addy's pass, I went through the current hooks and the issues this fixes reproduce for me. Both sdd cache hooks curl the raw tool_input URL with -L and no scheme or host restriction (revalidation HEAD in pre, validator HEAD in post), so the SSRF guard closes a real gap. In simplify-ignore.sh, line 154 restores backups via cat straight through the original path with no symlink check, and perl runs unguarded at lines 137 and 269 under set -euo pipefail (line 14), so a missing perl aborts the Edit/Write branch mid-rewrite.

One point I'd leave to Addy: the sdd-cache-pre.sh header documents no TTL as a deliberate choice, and the new SDD_CACHE_MAX_AGE default of 86400s reverses that, which reads as a semantics ruling rather than a review nit. Also, am I reading right that the guard checks hostname text only, so a name resolving to an internal address still passes?

addyosmani (approved):
This is the canonical #295 fix (I closed the duplicate in its favor). The SSRF guard is thorough: https-only, and it rejects loopback, RFC1918, link-local, CGNAT, and metadata hosts. Reasonable defense-in-depth for a hook that curls a model-supplied URL. Approving; worth a quick check it doesn't trip on normal cache URLs.

federicobartoli:
I ran the check @An-idd  asked for: extracted url_is_safe from the diff as-is and fed it a battery of 30 URLs. All 10 typical public docs URLs pass (mixed case, explicit ports, userinfo, CDNs included), and all 15 hostile ones are blocked (loopback, RFC1918, link-local, CGNAT, metadata hosts, bracketed IPv6, .local/.internal, plain http, ftp). The only false positives are exotic real domains whose names start like IP octets (0.gravatar.com, 127.net, 10.com): they get rejected, but the failure mode is just "no cache for that domain" and the fetch itself proceeds. Good enough for me.

Two small asks before I approve, both one-liners:

The pre hook's header still says "No TTL: if validators don't catch a change, nothing will", which the new SDD_CACHE_MAX_AGE cap now contradicts. Could you update that comment to describe the 24h default and name the env var?
On my July question about the guard checking hostname text only: I'm now satisfied the residual is small, since https-only plus zero redirects means a DNS name pointing at an internal IP only ever gets as far as a TCP/TLS handshake, because the internal host cannot present a valid certificate for that name. Could you add a comment line in url_is_safe documenting exactly that residual? The next reader will ask the same question I did.
With those two in I'll approve, and together with Addy's review that makes the two sign-offs we need. One heads-up: this overlaps #385 on the perl replacement lines; since this PR's sentinel-based strip preserves perl's exact one-newline semantics, the plan is to land this one and close [#385](https://github.com/federicobartoli/agent-skills/issues/385) as superseded after the merge.

PR #300, opened 2026-06-20
[docs] Add comparison to alternatives

addyosmani:
cc @federicobartoli @nucliweb 

This adds a comparison to Superpowers and other alternatives.

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,6 +326,12 @@ Skills bake in best practices from Google's engineering culture — including co
 
 ---
 
+## How it compares
+
+Wondering how this stacks up against [Superpowers](https://github.com/obra/superpowers) or [Matt Pocock's skills](https://github.com/mattpocock/skills)? See **[docs/comparison.md](docs/comparison.md)** for an honest, side-by-side look at how the three are shaped differently and when to reach for each — including a link to a controlled [head-to-head experiment](https://www.linkedin.com/pulse/superpowers-vs-agent-skills-faster-shipping-safer-reasoning-om-mishra-dzakf/).
+
+---
+
 ## Contributing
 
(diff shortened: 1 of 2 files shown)


nucliweb:
**Great addition**, the comparison table and honest experiment summary are exactly the kind of context developers need when evaluating these tools.

One gap worth addressing: the closing note ("you don't have to choose exclusively") is encouraging but doesn't warn about the friction that comes from running more than one framework simultaneously. Conflicts that come up in practice:

- **Name collisions**: if two frameworks define `/tdd`, `/plan`, or a similar command differently, the agent receives contradictory instructions for the same trigger.
- **Meta-skill routing**: using-agent-skills and Superpowers each define their own entry loop. With both active, the agent may start two parallel flows for the same task.
- **TDD philosophy conflicts**: Superpowers deletes prematurely written code to enforce the **RED→GREEN** line strictly; the other two are less prescriptive. Mixing them produces inconsistent behavior at that gate.
- **Commit/PR style rules**: if each framework ships its own instructions, the agent gets competing style guides.

A short paragraph at the end of "When to pick which" would make the mixing advice more actionable:

If you do combine them, load only one framework as your active router and pull in specific skills from the others à la carte. Running two meta-skills or two `/plan` commands in the same agent context will produce inconsistent behavior, not the best of both worlds.

addyosmani:
Good call, @nucliweb. You're right that "borrow freely" undersold the failure mode. Pushed `8fea8b4`: the closing now separates cherry-picking *individual* skills (fine) from running two frameworks as active routers at once (not fine, since they clash on command names like `/tdd`, routing logic, and TDD philosophy). It lands on exactly your framing: one primary router, then borrow from the others à la carte. Thanks for the sharp read.

PR #301, opened 2026-06-20
docs(contributing): guidelines to curb duplicate skill proposals

nucliweb:
## What

Adds a "Before proposing a new skill" subsection to `CONTRIBUTING.md` with a short checklist contributors run before opening a new-skill PR.

## Why

The goal is to define clear rules/guidelines that curb duplication. The open-PR backlog has accumulated clusters of near-duplicate skill proposals (several overlapping LLM/agent skills, meta-skills, and browser skills proposed independently). Most of these overlap with an existing skill or with each other, which makes triage slower.

The checklist asks contributors to:

1. Search the existing catalog (README + `skills/`).
2. Check open PRs for overlapping proposals (`gh pr list --state open`).
3. Confirm the idea fits the skill anatomy (actionable workflow, not vague advice).
4. Justify in the PR description why it isn't already covered; if it overlaps, extend the existing skill instead.

It also splits the section into "Before proposing a new skill" and "Creating the skill" so the decision step comes before the how-to.

## Scope

Documentation only. No skills, commands, or hooks touched.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,6 +4,19 @@ Thanks for your interest in contributing! This project is a collection of produc
 
 ## Adding a New Skill
 
+### Before proposing a new skill
+
+This pack already covers most of the development lifecycle, and many proposals overlap with an existing skill or another open PR. Before opening one, do these checks so reviewers aren't triaging duplicates:
+
+1. **Search the catalog.** Browse [the skill list in the README](README.md#all-24-skills) and skim `skills/` for an existing skill that covers your idea, whole or in part.
+2. **Check open PRs.** Run `gh pr list --state open` (or browse the PRs tab) and look for proposals on the same topic. Clusters of near-duplicate skills already exist; don't add to them.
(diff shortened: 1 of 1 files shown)


nucliweb:
cc @addyosmani @federicobartoli

addyosmani:
This is exactly the guardrail we need, @nucliweb. The duplication backlog is real, and a pre-flight checklist that points people at the catalog and open PRs before they propose would have headed off most of the recent near-duplicate clusters (the skill-router and slash-commands meta-skills, the two accessibility PRs, the LLM trio). It also pairs nicely with the existing "don't duplicate, reference instead" rule under *What Not to Do*: this makes it enforceable at proposal time, not just at review time. The "extend the existing skill instead of adding a new one" line is the part I care about most.

One small thing I pushed on top (`9a5e3b9`): the catalog link pointed at `README.md#all-24-skills`, and that anchor breaks the moment a new skill bumps the heading to "All 25 Skills". A doc about adding skills shouldn't carry a link that breaks when you add one, so it now points at `README.md` without the fragment.

LGTM, merging. Thanks for getting ahead of this.

PR #302, opened 2026-06-20
fix(code-review): align severity labels with the defined taxonomy

nucliweb:
## What

Aligns the severity labels used in `skills/code-review-and-quality/SKILL.md` with the taxonomy the skill itself defines.

## Why

"Step 4: Categorize Findings" defines the severity vocabulary: `Critical`, `Required` (no prefix), `Optional`/`Consider`, `Nit`, and `FYI`. Two other places in the same skill used labels outside that set:

- The Multi-Model Review prompt example said "Flag any issues as Critical, **Important**, or **Suggestion**."
- The Verification checklist referenced "All **Important** issues are resolved".

`Important` and `Suggestion` are never defined, so an agent following the skill can't map them to a real severity. This change updates both spots to use only the defined labels.

## Changes

- Multi-Model Review prompt: `Critical, Important, or Suggestion` -> `Critical, Required, Optional, or Nit`
- Verification: `All Important issues` -> `All required (non-Nit) issues`

## Scope

Documentation only, 2 lines. `node scripts/validate-skills.js` passes (24 skills, 0 errors).

### DIFF
--- a/skills/code-review-and-quality/SKILL.md
+++ b/skills/code-review-and-quality/SKILL.md
@@ -201,7 +201,7 @@ This catches issues that a single model might miss — different models have dif
 ```
 Review this code change for correctness, security, and adherence to
 our project conventions. The spec says [X]. The change should [Y].
-Flag any issues as Critical, Important, or Suggestion.
+Flag any issues as Critical, Required, Optional, or Nit.
 ```
 
 ## Dead Code Hygiene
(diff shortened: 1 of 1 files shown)


addyosmani:
Nice catch, @nucliweb. This is the kind of internal inconsistency that quietly trips an agent following the skill: "Categorize Findings" defines the vocabulary (Critical, Required, Optional/Consider, Nit, FYI), but the multi-model prompt and the verification checklist were still reaching for `Important` and `Suggestion`, which the skill never defines. An agent has no way to map those back to a real severity, so aligning both to the defined labels is exactly right.

I pushed one tiny follow-up (`38f6dde`): the verification line now reads "All Required (no-prefix) changes" to mirror the table's exact terms ("Required change", no prefix), rather than "required (non-Nit)". Same meaning, matched to the taxonomy word-for-word.

Validator's green, two-line docs change. LGTM, merging. Thanks for keeping the skills self-consistent.

PR #303, opened 2026-06-20
docs(spec): point Plan and Tasks phases to planning-and-task-breakdown

nucliweb:
## What

Adds a one-line delegation pointer to the Plan and Tasks phases of `spec-driven-development`, naming `planning-and-task-breakdown` as the canonical source for the planning mechanics.

## Why

The Plan (Phase 2) and Tasks (Phase 3) phases restated planning mechanics, dependency-graph mapping, vertical slicing, task sizing, and ordering, that the dedicated `planning-and-task-breakdown` skill already owns. That risks drift between the two skills as one evolves. The Implement phase (Phase 4) already delegates to `incremental-implementation` and `test-driven-development`; this brings phases 2 and 3 in line with that pattern.

The change is deliberately light: it keeps the inline bullets and the short task template as a quick reference, and only adds a pointer to the authoritative skill. No content is removed.

## Changes

- Phase 2: pointer to `planning-and-task-breakdown` for dependency-graph and vertical-slicing mechanics.
- Phase 3: pointer to `planning-and-task-breakdown` for task-sizing and dependency-ordering mechanics.

## Scope

Documentation only, +4 lines, 0 deletions. `node scripts/validate-skills.js` passes (24 skills, 0 errors).

### DIFF
--- a/skills/spec-driven-development/SKILL.md
+++ b/skills/spec-driven-development/SKILL.md
@@ -138,7 +138,7 @@ With the validated spec, generate a technical implementation plan:
 4. Identify what can be built in parallel vs. what must be sequential
 5. Define verification checkpoints between phases
 
-> Follow `planning-and-task-breakdown` for the dependency-graph mapping and vertical-slicing mechanics behind these steps; it is the canonical source.
+> Follow `planning-and-task-breakdown` for the dependency-graph mapping and vertical-slicing mechanics behind these steps; it is the canonical source. The bullets above are a lightweight summary; if they ever diverge, `planning-and-task-breakdown` takes precedence.
 
 The plan should be reviewable: the human should be able to read it and say "yes, that's the right approach" or "no, change X."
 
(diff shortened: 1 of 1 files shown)


federicobartoli:
Really like the direction here, delegating the planning mechanics to a single canonical skill is the right call. One source-of-truth concern though: the PR names planning-and-task-breakdown as canonical but keeps the existing Plan and Tasks bullets as normative guidance, so there are still two copies that can drift apart as either skill evolves. As written, the pointer mainly tells the reader where the other copy lives, it doesn't establish precedence.
Worth making that precedence explicit, for example: "The bullets below are a lightweight summary; if they ever diverge, planning-and-task-breakdown takes precedence." Alternatively, trim the inline bullets down to just the output contract specific to this workflow and let the canonical skill own the mechanics.
Minor consistency nit while you're in here: the two new pointers are formatted differently from each other. The Phase 2 one is a blockquote ("> Follow ...") and the Phase 3 one is a plain paragraph ("This step mirrors ..."). Picking one form for both would read more cleanly.

nucliweb:
Thanks Federico, both points were spot on.

I went with the explicit-precedence option rather than trimming the bullets, to keep the PR documentation-only and preserve the inline quick reference. Pushed in 5dc7310:

- Both pointers now state that `planning-and-task-breakdown` takes precedence if the inline summaries ever diverge, so the source of truth is unambiguous rather than just signposting where the other copy lives.
- Unified both pointers to the blockquote form (Phase 3 was a plain paragraph before), so they read consistently.

`node scripts/validate-skills.js` still passes (24 skills, 0 errors).

addyosmani (approved):

PR #304, opened 2026-06-21
feat(code-review): add structural-depth lenses to the code-review skill

addyosmani:
## What this does

**tl;dr:** We already had a pretty strong code-review skill, but I noticed there were some interesting ideas in Cursor's [thermo-nuclear](https://raw.githubusercontent.com/cursor/plugins/refs/heads/main/cursor-team-kit/skills/thermo-nuclear-code-quality-review/SKILL.md) and wanted to study if there is anything we could learn from them without conflicting with how many lenses our approach takes. This PR explores that.

Strengthens the `code-review-and-quality` skill with **structural-maintainability depth**. Our five-axis review (correctness, readability, architecture, security, performance) is broad and process-complete, but it treats *structure* passively - it asks "could this be fewer lines?" and moves on. A great reviewer goes further: they hunt for the simpler design hiding in the diff and reject refactors that just shuffle complexity around.

This PR folds that depth into the **existing** axes, sizing, severity, and verification (not a separate mode), so the skill stays one coherent review rather than two.

## The ideas added

- **Reduce, don't relocate** (Architecture axis). Ask whether a refactor *deletes* complexity or merely *moves* it. Count the concepts a reader must hold; if a "cleaner" version leaves that count unchanged, it isn't cleaner. Prefer the restructuring where whole branches/modes/layers disappear.
- **Missing-abstraction signals** (Readability axis). A new conditional bolted onto an unrelated flow, or repeated conditionals on the same shape, are design smells - not nits. Push the logic into a helper/state/dispatcher.
- **Structural Remedies** (new section). When you flag a structural problem, **name the move, not just the problem**: typed dispatcher over condition chains, collapse duplicate branches, separate orchestration from logic, reuse the canonical helper, make a type boundary explicit, delete pass-through wrappers, split large files.
- **File-size gate** (Change Sizing). We already size the *diff*; this also watches *total file size* — a small diff can push a file past a healthy boundary (~1000 lines), which should trigger "decompose first."
- **Lead with what matters** (Step 4). A prioritization rule: order findings by leverage and don't bury a real issue under cosmetic nits. One structural problem plus ten nits — the structural problem *is* the review.
- **Type-boundary cleanliness** (Architecture axis). Question gratuitous `any`/`unknown`/optional/casts and silent fallbacks that paper over an invariant.
- **Presumptive blockers** (Verification). A short default-block list (relocated-not-reduced complexity, file pushed past the size boundary, feature logic in shared modules, near-duplicate of a canonical helper, invariant-hiding fallback).
- Matching **Common Rationalizations** (+2) and **Red Flags** (+4).

## Calibration

We keep our pragmatic approval philosophy. The new structural lens is **flag-and-suggest by default** — surface the simpler design and propose it, but don't hold a genuine improvement hostage to a hypothetical rewrite. Block only when the change actively makes structure worse. This keeps the skill demanding on structure without tipping into "rewrite-or-reject."

## Scope and safety

- Purely additive: 348 → 381 lines, no deletions.
- `node scripts/validate-skills.js` passes (24 skills, 0 errors).
- Frontmatter and all existing sections unchanged.

### DIFF
--- a/skills/code-review-and-quality/SKILL.md
+++ b/skills/code-review-and-quality/SKILL.md
@@ -110,7 +110,7 @@ Small, focused changes are easier to review, faster to merge, and safer to deplo
 ~1000 lines changed  → Too large. Split it.
 ```
 
-**Watch file size, not just diff size.** A small diff can still push a file past a healthy boundary — around 1000 total lines is a common smell. When a change materially grows an already-large file, ask whether to extract helpers, subcomponents, or modules *first*, before piling more on. Decompose, then add.
+**Watch file size, not just diff size.** A small diff can still push a file past a healthy boundary — around 1000 *total* lines in a single file (distinct from the ~1000 *changed*-lines threshold above) is a common inspection signal, not a hard cap. When a change materially grows an already-large file, ask whether to extract helpers, subcomponents, or modules *first*, before piling more on. Decompose, then add.
 
 **What counts as "one change":** A single self-contained modification that addresses one thing, includes related tests, and keeps the system functional after submission. One part of a feature — not the whole feature.
 
--- a/skills/code-review-and-quality/SKILL.md
+++ b/skills/code-review-and-quality/SKILL.md
@@ -45,6 +45,8 @@ Can another engineer (or agent) understand this code without the author explaini
 - **Are abstractions earning their complexity?** (Don't generalize until the third use case)
 - Would comments help clarify non-obvious intent? (But don't comment obvious code.)
 - Are there dead code artifacts: no-op variables (`_unused`), backwards-compat shims, or `// removed` comments?
+- **Is a new conditional bolted onto an unrelated flow?** That's a design smell, not a nit — push the logic into its own helper, state, or policy instead of tangling an existing path.
+- **Do repeated conditionals on the same shape appear?** They signal a missing model or dispatcher. A "temporary" branch is usually permanent debt.
 
 ### 3. Architecture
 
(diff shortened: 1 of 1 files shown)


nucliweb:
Suggestion: tighten the "Presumptive blockers" wording

The new paragraph at the end of Verification pulls in two opposite directions within the same sentence: it opens with "treat these as **Required-by-default** unless the author justifies them" and closes with "**flag and suggest by default**, and block only when the change actively makes the structure worse." Required-by-default and flag-and-suggest-by-default are opposite stances, so a reviewer (human or agent) can't tell whether the default is to block or to propose.

Consider unifying to a single stance while keeping the signal list, e.g.:

> **Presumptive blockers:** surface and propose the simpler design for each of these; escalate to Required only when the change actively makes structure worse: a refactor that relocates complexity instead of reducing it; a change that pushes a file past the size boundary with no decomposition; feature logic added to a shared module; a near-duplicate of an existing canonical helper; a silent fallback that hides an unclear invariant.

That keeps the list intact, removes the ambiguity, and lines up cleanly with the "flag-and-suggest by default" calibration in the PR description.

addyosmani:
Good catch, @nucliweb. That paragraph was talking out of both sides of its mouth. Pushed `ba7cbe6` with your unified wording: a single stance now (propose the simpler design for each signal, escalate to Required only when the change actively makes structure worse), with the signal list intact. It lines up cleanly with the flag-and-suggest calibration in the description. Thanks for the close read.

federicobartoli:
Really like the direction here. One concrete clarity issue: 1000 now carries three meanings in the skill. It's hyperbole in the Readability axis ("1000 lines where 100 suffice"), diff size in Change Sizing ("~1000 lines changed"), and total file size in this new paragraph ("around 1000 total lines").

The latter two are adjacent, so a reader or agent could easily conflate "diff too large" with "file too large." Could we explicitly distinguish them here, for example: "Note: this refers to total file length, distinct from the ~1000 changed-lines threshold above," and clarify that total file length is an inspection signal rather than a hard boundary?

addyosmani:
Good catch, @federicobartoli. You're right that the two 1000s were sitting too close together for comfort. Pushed a4ae751 with the disambiguation inline: it now spells out total lines in a single file vs the changed-lines threshold above, and reframes it as a common inspection signal rather than a hard cap. 

The Readability-axis use is clearly hyperbole about a single hunk's local complexity, so I left it alone - let me know if you'd prefer it disambiguated too.

federicobartoli:
> Good catch, @federicobartoli. You're right that the two 1000s were sitting too close together for comfort. Pushed [a4ae751](https://github.com/addyosmani/agent-skills/commit/a4ae7518674ccfe6c5cd91353eef49046030e930) with the disambiguation inline: it now spells out total lines in a single file vs the changed-lines threshold above, and reframes it as a common inspection signal rather than a hard cap.
> 
> The Readability-axis use is clearly hyperbole about a single hunk's local complexity, so I left it alone - let me know if you'd prefer it disambiguated too.

Perfect, that reads cleanly now, the "total in a single file vs changed-lines above" split plus "inspection signal, not a hard cap" removes exactly the ambiguity I was worried about. Agree on leaving the Readability line as-is: "1000 where 100 suffice" clearly reads as a ratio, not a threshold, so there's no real collision there. LGTM from me 🚀❤️

PR #305, opened 2026-06-21
docs: explain expected commands/ warning when installed as Claude Code plugin

### LINKED ISSUE #298: Claude Code `/plugin` warns that root `commands/` (Antigravity) folder is ignored
## Summary

When `agent-skills` is installed as a **Claude Code** plugin, the `/plugin` manager shows a warning that the repo-root `commands/` folder is being ignored. This is **cosmetic — nothing is broken** (all Claude Code slash commands load fine), but it may confuse installers, so filing as an FYI + a question about whether it can be cleanly removed.

## The warning

```
❯ agent-skills (user)
   Default commands/ folder is ignored because the manifest sets "commands"
   Remove "commands" from .claude-plugin/plugin.json (or SKILL.md
   frontmatter) to auto-load the folder, or add the folder's files to the
   "commands" list if you want both
```

## Root cause

The repo serves multiple platforms, and two command directories sit side by side:

| Location | Platform | Format |
|---|---|---|
| `.claude/commands/*.md` | Claude Code (manifest `"commands": "./.claude/commands"`) | `.md` |
| `.gemini/commands/*.toml` | Gemini CLI | `.toml` |
| **root `commands/*.toml`** | **Antigravity CLI** (per `README.md` L310) | `.toml` |

Claude Code's plugin loader **auto-scans a root `commands/` folder by default**. Because `.claude-plugin/plugin.json` explicitly sets `"commands": "./.claude/commands"`, the loader correctly uses that path and ignores the root `commands/` folder — but emits the warning above about the collision.

## Impact: none

- All Claude Code commands load correctly from `.claude/commands/`.
- The root `commands/` folder is **intentional** — it's the Antigravity CLI command set and must stay. **This is not a request to delete it.**

So the warning is essentially Claude Code's loader being noisy about an intentional multi-platform layout.

## Question / possible options (maintainer's call)

1. Does the Antigravity CLI plugin manifest (root `plugin.json`) support a **configurable commands path**? If so, relocating Antigravity's commands out of the repo root (e.g. into a dedicated subdir) would stop Claude Code's default root scan from colliding with the manifest override, and the warning would disappear for Claude Code users — without affecting Antigravity.
2. If Antigravity requires `commands/` at the repo root (hardcoded convention), then this is likely just cosmetic / won't-fix. A short note in the Claude Code setup docs ("Claude Code may warn that the root `commands/` folder is ignored — this is expected; it belongs to the Antigravity CLI") would save installers a moment of confusion.

## Environment

- Installed via Claude Code plugin marketplace (`addy-agent-skills`), latest as of 2026-06-18.
- Manifest: `.claude-plugin/plugin.json` → `"commands": "./.claude/commands"`.

creazyfrog:
## Summary

- Fixes #298: Claude Code `/plugin` warns that root `commands/` (Antigravity) folder is ignored
- Adds a short Note callout under the "Using Commands" section in `docs/getting-started.md` explaining that the warning is cosmetic and expected

## Changes

- `docs/getting-started.md`: added a blockquote note explaining that the `"Default commands/ folder is ignored"` warning is expected when the plugin is installed in Claude Code, and that all Claude Code slash commands load correctly from `.claude/commands/`

## Testing

Docs-only change. Verified by re-reading the section before and after to confirm the note accurately describes the situation documented in #298 (the root `commands/` folder is intentionally for Antigravity CLI; `.claude/commands/` is the correct path for Claude Code; the warning is cosmetic).

Closes #298

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -112,6 +112,12 @@ The `.claude/commands/` directory contains slash commands for Claude Code:
 | `/review` | code-review-and-quality |
 | `/ship` | shipping-and-launch |
 
+> **Note:** When installed as a Claude Code plugin you may see a warning like
+> _"Default commands/ folder is ignored because the manifest sets 'commands'"_.
+> This is expected. The root `commands/` directory belongs to the Antigravity CLI
+> and is intentionally separate from `.claude/commands/`. All Claude Code slash
+> commands load correctly from `.claude/commands/`; the warning is cosmetic.
+
 ## Using References
 
 The `references/` directory contains supplementary checklists:


addyosmani (approved):
lgtm

PR #312, opened 2026-06-22
docs(debugging): add production incident response & blameless postmortems

HMAKT99:
## TL;DR

A single **purely additive** section added to `skills/debugging-and-error-recovery/SKILL.md` — **Production Incident Response & Blameless Postmortems**. No new directory, no README/manifest/count changes, revert-safe. `node scripts/validate-skills.js` → **0 errors, 0 warnings**.

## The gap this fills

`debugging-and-error-recovery` is a **dev-time** triage workflow (reproduce → localize → reduce → fix → guard). It explicitly lists "production incidents" in scope but has no workflow for one — and a live incident **inverts the order**: you stabilize first and diagnose second. Nothing in the pack covers declare/severity, stabilize-before-diagnose, incident roles, or the blameless postmortem that turns an incident into prevention.

## What's added

- **Severity-first response** (SEV1/2/3 → response level)
- **Stabilize before you diagnose** — the production sibling of the skill's existing Stop-the-Line Rule (rollback / flip the flag / failover, *then* root-cause)
- **Coordination** — incident driver + scribe + running log
- **Blameless postmortem** — telemetry-based timeline (by correlation ID), contributing factors not a culprit, action items with owners
- **Close the loop** — every incident yields a regression test (and an eval case for LLM-backed features) + a detection-gap fix

## Why this isn't a duplicate (per CONTRIBUTING)

It **references rather than re-teaches** adjacent skills:
- Rollback/feature-flag mechanics → `shipping-and-launch`
- Regression/eval cases → `test-driven-development`
- Timeline reconstruction relies on signals from `observability-and-instrumentation`

It's a focused edit to an existing skill (the refinement form CONTRIBUTING prefers), not a new directory.

## Validation

- [x] `node scripts/validate-skills.js` → PASSED (0 errors, 0 warnings); all required sections intact
- [x] Cross-references resolve to existing skills (no dead refs)
- [x] Diff touches only the one `SKILL.md`

### DIFF
--- a/skills/debugging-and-error-recovery/SKILL.md
+++ b/skills/debugging-and-error-recovery/SKILL.md
@@ -259,6 +259,60 @@ Add logging only when it helps. Remove it when done.
 - API error logging with request context
 - Performance metrics at key user flows
 
+## Production Incident Response & Blameless Postmortems
+
+The triage checklist above is for failures you reproduce at your desk. A **production incident** — users are hurt *right now* — inverts the order: you stabilize first and diagnose second. The instinct to find the root cause before acting is correct in development and wrong in an incident; every minute spent understanding is a minute users stay broken.
+
+```
+ DETECT ──→ DECLARE ──→ STABILIZE ──→ COORDINATE ──→ RESOLVE ──→ POSTMORTEM ──→ PREVENT
+ (alert/   (severity   (mitigate     (one driver,   (root      (blameless,    (action items
+  report)   + comms)    NOW)          one scribe)    cause)      timeline)      with owners)
+```
+
+### 1. Declare and size it
+
+Don't debate whether it's "really" an incident — declare and assign a severity. The severity sets the response, not the other way around.
+
+| Sev | Impact | Response |
+|---|---|---|
+| **SEV1** | Core flow down / data loss / security breach | All hands, page now, exec comms |
+| **SEV2** | Major feature degraded, no clean workaround | Page on-call, active mitigation |
+| **SEV3** | Minor or partial degradation, workaround exists | Business-hours fix, ticket |
+
+### 2. Stabilize before you diagnose
+
+Reach for the fastest *reversible* mitigation before hunting the cause: roll back the recent deploy, flip the feature flag off, fail over, or shed load. This is the production sibling of the **Stop-the-Line Rule** — stop the bleeding, then investigate on a system that's no longer hurting users. For rollback and feature-flag mechanics, follow the `shipping-and-launch` skill.
+
+```
+Is there a recent deploy/flag change correlated with onset?
+  → YES: roll it back / flip it off first, confirm recovery, THEN find out why
+  → NO:  mitigate the symptom (failover, rate-limit, degrade) before root-causing
+```
+
+### 3. Coordinate
+
+One **incident driver** (decisions + comms) and one **scribe** (timestamps every action and finding in a running log). Everyone else investigates. The running log is what makes the postmortem possible — write it *during*, not after.
+
+### 4. Resolve, then run a blameless postmortem
+
+Once stable and root-caused (use the Triage Checklist above on the now-safe system), write a postmortem for every SEV1/SEV2. **Blameless** means it names *contributing factors and missing safeguards*, never a person — "the deploy had no canary stage" not "Dana shipped the bug." People act in good faith on the information they had; if a single human error could take production down, the system lacked a guardrail, and that guardrail is the fix.
+
+A postmortem contains:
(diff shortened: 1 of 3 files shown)


nucliweb:
Reviewed this, and it's a solid, well-scoped addition. Sharing two notes: one confirming the sizing call, one optional.

**Size / shape is right.** The change is 67 lines added (~5K chars), purely additive, leaving the file at 367 lines, comfortably within the pack's range (`security-and-hardening` is 461, `ci-cd-and-automation` 390, `test-driven-development` 383). It's under the 100-line threshold that the contributing conventions use to decide when content should move into a supporting file, so keeping it inline is the correct call rather than spinning out a new file. `node scripts/validate-skills.js` passes clean here too (0 errors, 0 warnings), and all three cross-skill refs resolve.

**Why inline beats a `references/` file (in case it comes up).** I considered whether this belongs in `references/` instead, and I think inline is right:
- The `references/` files in this repo are quick-reference checklists and pattern catalogs. This is a *workflow* (DETECT -> DECLARE -> STABILIZE -> ... -> PREVENT), and workflows belong in skills.
- The section leans on the host skill by name, the Stop-the-Line Rule, "the Triage Checklist above", and "the Guard-Against-Recurrence test from Step 5 above". That in-context coupling is a feature; extracting it would break those back-references.

**One optional follow-up: close the discovery loop in the other direction.** The section already links outward to `shipping-and-launch`, `observability-and-instrumentation`, and `test-driven-development`, but those skills don't point back. A couple of short pointers would make incident response reachable from where people actually are when they need it:
- From `observability-and-instrumentation`: a note that the telemetry-based postmortem timeline is driven from this section.
- From `shipping-and-launch`: from the rollback / feature-flag mechanics, point to "Stabilize before you diagnose".

Not blocking. Nice work, the stabilize-before-diagnose framing and the blameless-postmortem distinction are the genuinely valuable parts.

HMAKT99:
Thanks for the careful review — and for spelling out the inline-vs-`references/` reasoning; that matches the intent exactly (it's a workflow with in-context back-references to the host skill, so extracting it would break them).

Good call on closing the discovery loop both ways. I've pushed the two back-pointers:
- `observability-and-instrumentation` (correlation IDs → the postmortem timeline is reconstructed from this telemetry)
- `shipping-and-launch` ("When to Roll Back" → "Stabilize before you diagnose")

Both are one-liners, fully additive; `validate-skills.js` still passes 0/0 and all cross-refs resolve. Appreciate the thoughtful feedback.

addyosmani:
Blameless incident response is a good addition to the debugging skill and the content reads well. Quick check before merge: make sure it doesn't overlap shipping-and-launch and observability, which already touch on-call and rollback, so the skills point at each other rather than repeat. If the boundaries are clean, I'm happy to take it.

HMAKT99:
The boundaries are clean — here's where each skill starts and stops:

- **`debugging-and-error-recovery`** (this PR): declare severity, stabilize vs diagnose, reconstruct timeline, blameless postmortem, close-the-loop test. The *process* of handling a live incident.
- **`observability-and-instrumentation`**: the telemetry that makes the timeline reconstructable. The back-pointer I added after nucliweb's review names that dependency explicitly: correlation IDs → postmortem timeline.
- **`shipping-and-launch`**: rollback mechanics and feature-flag controls. The back-pointer there reads: after you've rolled back, go to `debugging-and-error-recovery` to stabilize before diagnosing.

None of the three skills teach the same step. They call each other by name rather than repeating the content — which is what the validator checks (all cross-refs still resolve 0/0).

nucliweb (approved):
**Boundaries check out, this is ready to merge from my side.** I re-verified the three-skill split against the current diff: `debugging-and-error-recovery` owns the incident *process* (declare/severity, stabilize-before-diagnose, timeline, blameless postmortem, close-the-loop), `observability-and-instrumentation` owns the telemetry the timeline is reconstructed from, and `shipping-and-launch` owns the rollback/flag mechanics. No step is taught twice; they reference each other by name.

The two back-pointers from my earlier review are in and resolve, and the in-section anchors all land where they should: "Stop-the-Line Rule", "the Triage Checklist above", and "the Guard-Against-Recurrence test from Step 5 above" all point at the right existing headings. `validate-skills.js` stays 0/0 and CI is green.

One tiny non-blocking note: the header flow diagram lists seven stages (DETECT…PREVENT) while the body has five numbered steps, since DETECT is the trigger and POSTMORTEM/PREVENT fold into steps 4–5. Reads fine, just a count mismatch a careful reader might pause on. Not worth holding the PR for.

addyosmani:
Thanks for this. Incident response and postmortems fit observability-and-instrumentation and shipping-and-launch; closing as overlap, a targeted addition there is welcome.

PR #313, opened 2026-06-22
docs: route new-skill work through CONTRIBUTING pre-flight checks

nucliweb:
## What

Adds anti-duplication guardrails so AI agents (Claude Code, Cursor, Copilot, Gemini, OpenCode, etc.) working in this repo follow the CONTRIBUTING.md pre-flight checklist before adding a skill. CONTRIBUTING.md stays the single source of truth; the other files only point to it instead of restating the checklist.

- **CLAUDE.md**: adds a `Contributing` pointer to `CONTRIBUTING.md#before-proposing-a-new-skill` and a pre-flight line under Boundaries.
- **AGENTS.md** (the cross-agent file): replaces the stale "Creating a New Skill" section with a short, accurate pointer. The old section listed `scripts/` and per-skill `.zip` as *Required* and shipped a SKILL.md template that no longer matches the repo (it has 0 zips, scripts are used by 1 of ~30 skills, and skills are markdown-first). The canonical format now lives only in `docs/skill-anatomy.md`.
- **.claude/rules/skills-contributing.md**: a new path-scoped rule (see below).

## Why we add `.claude/rules/skills-contributing.md`

It is a Claude Code **rule** (per [Steering Claude Code](https://claude.com/blog/steering-claude-code-skills-hooks-rules-subagents-and-more)): a constraint file under `.claude/rules/` that loads contextually rather than always-on.

- **Path-scoped to `skills/**`**: the guardrail loads only when an agent is actually editing skills, so it costs no context the rest of the time. This is the blog's recommended pattern for conventions that apply selectively.
- **What it enforces**: before creating a new `skills/<name>/` directory or significantly reworking one, run the CONTRIBUTING.md pre-flight checks (search the catalog, check open PRs, justify the gap) and prefer extending an existing skill over adding a near-duplicate.
- **Why a rule and not just CLAUDE.md**: CLAUDE.md should stay short and general; the dedup guardrail is specialized and only relevant inside `skills/`. A path-scoped rule keeps the constraint strong and in-context exactly when it matters, without bloating the always-loaded root file.
- **No duplication**: the rule states the intent and links to CONTRIBUTING.md for the full steps, so the checklist lives in exactly one place.

## Coverage

- Claude Code, CLAUDE.md (always) plus the path-scoped rule (when editing skills)
- All other agents, AGENTS.md


nucliweb:
Hi @addyosmani @federicobartoli, could you take a look when you have a moment? This wires agents into the CONTRIBUTING.md pre-flight checks to prevent duplicate skills, keeping CONTRIBUTING.md as the single source of truth, and adds a path-scoped Claude Code rule that loads the dedup guardrail only when editing `skills/`. Happy to adjust the approach. Thanks!

nucliweb:
### Guardrail test: it blocks a duplicate before any PR is opened

The guardrail is instructions (pre-flight checks loaded by `.claude/rules/skills-contributing.md`), not an automated gate. To validate it, I had an agent run the flow the rule mandates for a simulated request: *"add a skill to optimize/reduce LLM API costs (token budgets, caching, model routing)."*

**Step 1: search the catalog (`skills/`) for the topic**
```bash
grep -ril -E "llm.cost|token budget|cost optim|model routing|api cost" skills/
```
```
No match in existing skills (not a merged skill yet)
```

**Step 2: check open PRs for an equivalent proposal**
```bash
gh pr list --repo addyosmani/agent-skills --state open --limit 100 \
  --json number,title -q '.[] | select(.title|test("cost|llm|token|eval";"i")) | "#\(.number)  \(.title)"'
```
```
#308  docs(skills): add the eval->learn->improve feedback loop (TDD evals + observability step 8)
#287  Add llm-cost-optimization skill        <- exact topic match
#286  Add evaluating-llm-output skill
#254  feat: add new skills, agent personas, and LLM integration checklist
#129  Add prompt-injection-defense skill (OWASP LLM01)
#71   Feature: Add confidence scoring and LLM-As-A-Judge to code-review-and-quality skill
#51   feat: introduce evaluation framework for skill testing
```

**Agent decision the rule forces (instead of opening a PR):**
> STOP: duplicate detected. `llm-cost-optimization` is already proposed in open PR #287. I will not create `skills/llm-cost-optimization/` or open a new PR. Correct actions: review/collaborate on #287, or, if my approach differs, comment **on #287** rather than duplicating. I'd only proceed if I could justify a gap #287 doesn't cover.

Without the guardrail, the agent goes straight to creating the skill and opening PR #N+1, exactly the near-duplicate cluster CONTRIBUTING.md aims to prevent. The same flow catches duplicates of *merged* skills at step 1 (e.g. "add a code-review skill" finds `code-review-and-quality/`).

addyosmani:
Good cleanup, and the direction is right, @nucliweb. I verified the moving parts. The repo has 0 zips and scripts in 1 of 24 skills, so the old AGENTS.md section that called `scripts/` and `.zip` **Required** was actively wrong. I also confirmed the `.claude/rules/` mechanism against the current Claude Code docs: `.claude/rules/*.md` auto-loads, and `paths: ["skills/**"]` is the correct field for path-scoping, so the rule fires contextually only when an agent edits `skills/**`. Routing everything through CONTRIBUTING.md plus `docs/skill-anatomy.md` as the single sources is the right shape.

One thing to fix before merge: the 102-line deletion takes out more than the stale parts. Alongside the zip / Required-scripts / claude.ai-install boilerplate (correct to remove), it also drops two subsections that are still accurate and have no home elsewhere:

- **Best Practices for Context Efficiency**: "keep SKILL.md under 500 lines," "prefer scripts over inline code." Not in `skill-anatomy.md` or CONTRIBUTING.md.
- **Script Requirements**: shebang, `set -e`, status to stderr, cleanup traps. Still relevant for the one skill that ships scripts, and any future one.

`skill-anatomy.md` currently keeps only the progressive-disclosure line, none of the above. Could you port those two into `skill-anatomy.md` (the new canonical home) as part of this PR, so the cleanup doesn't quietly lose real guidance? With that in, this is a clean win.

Optional follow-up: `.claude/rules/` loads only in Claude Code. Cross-tool reach for the other agents comes from the CLAUDE.md / AGENTS.md pointers, which is fine. A `.cursor/rules/` twin would be the natural next step if you want the same path-scoped enforcement in Cursor.

nucliweb:
Thanks @addyosmani 

Done in c8e1b46. Ported both flagged subsections into `docs/skill-anatomy.md` (the new canonical home), so the cleanup keeps the guidance it was never meant to drop:

- **Context Efficiency**: the on-demand loading model plus "keep `SKILL.md` under 500 lines" and "prefer scripts over inline code (script execution doesn't consume context, only output does)".
- **Script Requirements**: `#!/bin/bash` shebang, `set -e`, status to stderr, JSON to stdout, cleanup trap, and the `/mnt/skills/user/<skill-name>/scripts/<script>.sh` path convention.

While at it I re-audited the full 102-line AGENTS.md deletion line by line. The remaining removals are either wrong/obsolete (zip + Required-scripts Directory Structure, Creating the Zip Package) or already covered elsewhere (Naming Conventions and the SKILL.md template live in `skill-anatomy.md`; end-user installation is superseded by the `/plugin marketplace` flow documented in the README). I grep-confirmed that "under 500 lines", the `set -e` requirement, and the `/mnt/skills/...` path guidance existed nowhere else, so those two were the only orphans, exactly the ones you flagged.

On the optional follow-up: agreed a `.cursor/rules/` twin is the natural next step for path-scoped enforcement in Cursor. I'll keep it out of this PR to stay focused and open a separate one. Thanks for the careful review!

Issue #35, opened 2026-04-07
feat: use content: fork in Claude Code slash commands for isolated skill execution


nucliweb:
## Background

Skills and slash commands currently inject their instructions directly into the active conversation context. Claude Code supports a `context: fork` frontmatter field that runs them in an isolated sub-agent context instead. This applies to both `SKILL.md` files and `.claude/commands/*.md` slash command files — they share the same frontmatter specification.

> **Note:** The correct field name is `context: fork`, not `content: fork`. These are distinct: `context` controls execution isolation; there is no `content` field in the Claude Code frontmatter spec.

## Problem

When a skill or command is invoked in a long session, it competes with accumulated conversation history for context space. The skill instructions mix with unrelated prior turns, which can dilute the signal and cause the agent to drift from the intended workflow.

## Proposed Change

Add `context: fork` to the YAML frontmatter of all `SKILL.md` files and all `.claude/commands/*.md` slash commands.

**Commands** (`.claude/commands/`):

```yaml
---
description: Conduct a five-axis code review — correctness, readability, architecture, security, performance
context: fork
---
```

**Skills** (`skills/*/SKILL.md`):

```yaml
---
name: code-review-and-quality
description: Conducts multi-axis code review. Use before merging any change...
context: fork
---
```

An optional `agent:` field can also be specified to select which subagent type runs the fork (e.g. `agent: Plan` for planning skills, `agent: general-purpose` for everything else). If omitted, it defaults to `general-purpose`.

### 1. Context isolation
The skill or command runs in its own context window, free from prior conversation history. The agent focuses exclusively on the skill workflow without interference from unrelated turns.

### 2. Consistent, repeatable execution
Each invocation starts from the same clean state. Invoking `/review` twice in the same session produces equivalent results, not results that degrade as context fills.

### 3. Token efficiency in the main session
Execution tokens are spent in the forked context, not in the main conversation. The main session's context window stays available for other work.

### 4. Stronger skill fidelity
Skills are step-by-step workflows, not reference material. Running them in a focused, isolated context makes it harder for the agent to shortcut or skip steps because there is no accumulated "I already did this" history to rationalize against.

### 5. Cleaner skill handoff
When a skill completes, only the result surfaces back to the main conversation — not the entire intermediate reasoning. This keeps the main session readable.

## Affected Files

**Commands** (`.claude/commands/`):
- `build.md`
- `code-simplify.md`
- `plan.md`
- `review.md`
- `ship.md`
[...]

PR #316, opened 2026-06-22
docs: add advanced per-agent configuration guide

nucliweb:
Adds a draft reference doc consolidating the per-agent configuration discussion from #272 and #36.

The goal is a single place that documents how to keep shared `SKILL.md` files portable (spec frontmatter only) while applying the vendor-specific runtime controls each agent offers as an opt-in:

- **Principles** for where vendor fields belong (spec fields top-level, vendor metadata under `metadata`, runtime orchestration in per-agent adapter files).
- **Claude Code:** `context: fork` and `allowed-tools` in `.claude/commands/`.
- **Gemini CLI / Antigravity:** subagent definitions in `.gemini/agents/*.md`, model routing, tool restrictions, turn limits, plus Gemini 3.x `thinking_level` guidance.
- **Automating with scripts:** generating per-agent output from a single source of truth, with semantic validation.

The doc is explicitly marked as a draft, and the script-based automation is a proposal pending agreement before any implementation.

### DIFF
--- a/docs/advanced-per-agent-configuration.md
+++ b/docs/advanced-per-agent-configuration.md
@@ -26,7 +26,8 @@ This is the canonical reference for per-agent configuration, and it supersedes t
 Claude Code reads `name` and `description` for discovery. Custom commands and skills share the same frontmatter, so the `context` and `allowed-tools` fields below work in both `.claude/commands/*.md` and `.claude/skills/*/SKILL.md`.
 
 - **`context: fork`** runs a command or skill in an isolated subagent context instead of the active conversation. This gives context isolation, repeatable execution from a clean state, execution tokens spent in the fork rather than the main session, and a cleaner handoff where only the result surfaces back. An optional **`agent`** field selects which subagent type runs the fork (for example `agent: Plan`); it defaults to `general-purpose`.
-- **`allowed-tools`** restricts the tool set available during a command, supporting least-privilege for passive stages such as review or audit.
+- **`allowed-tools`** pre-approves tools; it does not restrict them. Listed tools run without a permission prompt for the turn that invokes the command or skill, and the grant clears on your next message. Nothing is removed from the tool pool.
+- **`disallowed-tools`** is the field that actually restricts: it removes the listed tools from the available pool while the skill is active, which is what least-privilege needs for passive stages such as review or audit. The restriction also clears on your next message. Note it is a Claude Code field and not one of the specification's six, so it belongs in a local copy or an adapter file; in a published `SKILL.md` it fails packaging with an unexpected-key error.
 
 These are Claude Code conventions. Keep the shared, published `SKILL.md` limited to spec fields; apply these runtime controls in your own local command or skill copies instead.
 
--- a/docs/advanced-per-agent-configuration.md
+++ b/docs/advanced-per-agent-configuration.md
@@ -1,6 +1,6 @@
 # Advanced per-agent configuration
 
-Skills in this repository are intentionally portable. Every `SKILL.md` keeps its frontmatter limited to the fields the Agent Skills specification defines, so the same file works across Claude Code, Cursor, Gemini CLI, Antigravity, and any other spec-conformant client.
+Skills in this repository are intentionally portable. Every `SKILL.md` keeps its frontmatter limited to the fields the [Agent Skills specification](https://agentskills.io) defines, so the same file works across Claude Code, Cursor, Gemini CLI, Antigravity, and any other spec-conformant client.
 
 Several agents also support extra runtime controls: model routing, tool restrictions, turn limits, execution isolation. These controls are valuable, but they are vendor-specific and do not belong in the shared, portable frontmatter. This document explains which controls each agent supports, where they belong, and how to apply them as an opt-in without breaking portability.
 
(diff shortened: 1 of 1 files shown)


addyosmani:
This is the right architecture, @nucliweb, and a cleaner answer than the inline approach in #272. Keeping `SKILL.md` portable (spec fields top-level, vendor config under `metadata` or in per-agent adapter files) is exactly how the shared skills should stay tool-agnostic. The generator idea is the part that makes it sustainable: a per-skill config map that emits the adapter files and leaves `SKILL.md` untouched, so we never hand-maintain vendor fields across 24+ skills.

Two things to take it out of draft:

1. Say explicitly that this supersedes the inline-frontmatter approach, so there's no ambiguity for contributors (or for #272). As written, a reader could still think top-level `model`/`temperature` on a skill is fine.
2. The "prefer `thinking_level` over hardcoded `temperature` for 3.x models" guidance is a good catch. Worth a one-line example so it's copy-pasteable.

Once it graduates, I'd make this the canonical reference.

nucliweb:
Addressed both points in `8b34e6f` and took it out of draft:

- **Supersedes the inline approach (explicit).** Added a statement up front: this is the canonical reference, and `kind`/`model`/`temperature`/`max_turns`/`tools`/`context` must not appear at the top level of a skill, they belong under `metadata` or in a per-agent adapter file. No ambiguity for contributors.
- **`thinking_level` example.** Added a copy-pasteable `.gemini/agents/*.md` snippet using `thinking_level: high` with a `# not: temperature: 0.1` note.
- Updated Status: #272 and #36 are now closed in favor of this doc.

Ready for review. Happy to co-write the generator follow-up whenever you want to take that on.

nucliweb:
Pushed `7948a1a` addressing review follow-ups and the four refinements from a fresh pass over the doc:

1. **Consolidated #35.** Added it to the Status list alongside #272 and #36 (now closed in favor of this doc), so no open issue on the fork topic is left unlinked.
2. **Documented the `agent` field.** `context: fork` now notes its optional `agent` companion that selects the subagent type (e.g. `agent: Plan`, default `general-purpose`).
3. **Clarified the commands/skills framing.** Since custom commands and skills share the same frontmatter, the doc now states `context` and `allowed-tools` work in both `.claude/commands/*.md` and `.claude/skills/*/SKILL.md`, and keeps the "shared SKILL.md stays spec-only" as a portability recommendation rather than a technical limit.
4. **Linked the spec.** First mention of the Agent Skills specification now points to agentskills.io, backing the normative "must not appear at the top level" claim.

Verified the `context: fork`/`agent` and `allowed-tools` fields against the current Claude Code docs before writing. Docs-only change, CI green. Ready for another look.

nucliweb:

addyosmani:
Good direction and well-written, but it commits us to an architecture I'm still deciding on, specifically whether per-agent config lives in the skill files or in adapter files. I don't want to publish the doc until that call is made, or we end up documenting one answer while an open PR argues the other. Let me settle the config-location question and this slots right in.

nucliweb:
No rush on this, @addyosmani, just checking in. This is still parked on the config-location call from your last note, whether per-agent config lives in the skill files or in adapter files, and the doc is ready to align the moment that is settled. If it helps move it, I am happy to sketch both framings side by side so you can see which reads cleaner, or to co-write the generator follow-up we discussed. Whatever is most useful, no pressure on timing.

addyosmani (approved):
Yes. Keeping SKILL.md frontmatter to the spec fields and pushing vendor controls under metadata or an adapter is exactly the portability line I want, and it lines up with the Agent Plugins work in #463. Approving.

federicobartoli (changes requested):
I went through this against the primary sources: the Agent Skills spec field list matches exactly, the Gemini 3 guidance (temperature at default, thinking_level instead) checks out against Google's docs, and the Claude Code claims are all confirmed by the official docs, including the general-purpose default for agent. Well grounded.

One correction before I approve, since this positions itself as the canonical reference: the allowed-tools line says it "restricts the tool set available", but per the Claude Code docs it is a permission pre-approval (listed tools run without per-use prompts, and the grant clears on the next message); it does not remove anything from the pool. The field that actually restricts is disallowed-tools. Could you rephrase that bullet as pre-approval and mention disallowed-tools for the least-privilege case? With that one line fixed I'll approve!

nucliweb:
@federicobartoli, you're right, and the docs confirm it word for word. Fixed in `485cd8e`.

The frontmatter reference gives `allowed-tools` as "Tools Claude can use without asking permission during the turn that invokes this skill. The grant clears when you send your next message", and the section it links to is titled "Pre-approve tools for a skill". `disallowed-tools` is the restricting one: "Tools removed from Claude's available pool while this skill is active."

So the bullet is now two:

- **`allowed-tools`** pre-approves tools; it does not restrict them. Listed tools run without a permission prompt for the turn that invokes the command or skill, and the grant clears on your next message. Nothing is removed from the tool pool.
- **`disallowed-tools`** is the field that actually restricts: it removes the listed tools from the available pool while the skill is active, which is what least-privilege needs for passive stages such as review or audit.

Chasing that down surfaced one more thing worth having in a doc about portability: `disallowed-tools` is **not** one of the specification's six fields. The published list is `allowed-tools, compatibility, description, license, metadata, name`, so a `disallowed-tools` key in a shared `SKILL.md` fails packaging with an unexpected-key error rather than being ignored. I added that caveat, because it means the least-privilege recommendation has to live in a local copy or an adapter file, which is exactly the line this doc is drawing.

Separately, I rebased onto current `main` (`f63ec56`). The branch was 169 commits behind and predated the validator scripts, so CI had nothing current to run against it. All five validators pass locally now.

nucliweb:
@federicobartoli gentle nudge on this one whenever you have a window.

The `allowed-tools` correction you asked for landed in `485cd8e` on 27 Aug: the bullet is now split into `allowed-tools` as pre-approval (nothing leaves the tool pool, and the grant clears on the next message) and `disallowed-tools` as the field that actually restricts, which is what least privilege needs. Chasing it down also surfaced that `disallowed-tools` is not one of the specification's six top-level fields, so it fails packaging rather than being ignored, and that caveat is in the doc now too.

I also rebased onto current `main` (`f63ec56`), since the branch was 169 commits behind and predated the validator scripts. CI has something real to run against it now: all four checks green, `MERGEABLE`.

No urgency at all. Flagging it only because your 8 Aug review is still the standing verdict, so the PR reads as blocked from the outside even though the one line is resolved.

federicobartoli:
LGTM!

federicobartoli (approved):

PR #340, opened 2026-07-03
Add Autohand Code setup notes

igorcosta:
Hi, thanks for maintaining this skills pack.

I noticed the Quick Start already covers several coding-agent harnesses. This adds a small Autohand Code section that follows the same collapsible style and points users to the verified Autohand skill locations:

- user-level: `~/.autohand/skills/`
- project-level: `.autohand/skills/`
- catalog installs via `autohand --skill-install`

I kept this README-only and avoided claiming a direct registry install for this repo, since I only verified the generic Autohand catalog install flow from the current Autohand Code CLI docs.

Tests were not run because this is documentation-only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,32 @@ Use agent definitions from `agents/` as Copilot personas and skill content in `.
 </details>
 
 <details>
+<summary><b>Autohand Code</b></summary>
+
+Autohand Code discovers `SKILL.md` files from user and project skill directories.
+
+**User-level install:**
+
+```bash
+git clone https://github.com/addyosmani/agent-skills.git
+mkdir -p ~/.autohand/skills
+cp -R agent-skills/skills/* ~/.autohand/skills/
+```
+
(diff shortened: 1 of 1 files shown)


addyosmani:
Leaning against this one. We're keeping the core pack vendor-neutral, and a setup guide for a specific third-party product is a promotion-and-maintenance commitment I'm cautious about taking on. If Autohand reads the skills standard natively the way our other integrations do, the pack already works there without a bespoke guide. Happy to be corrected if it needs special wiring, but as setup-notes-for-a-product it's probably a pass.

addyosmani:
Thanks, and I appreciate that you matched the existing collapsible style. I'm going to pass on this one though: the pack stays vendor-neutral, and a setup section for a specific commercial coding tool is a maintenance commitment to something we don't control. If Autohand reads the skills standard the way our native integrations do, it already works without a bespoke section. Closing, but no knock on the tool.

PR #345, opened 2026-07-05
docs(deprecation): add database schema migration patterns (expand/contract)

HMAKT99:
## What this adds

A `### Database Schema Migrations (Expand/Contract)` section under the existing `## Migration Patterns` in `skills/deprecation-and-migration/SKILL.md`.

## The gap it fills

The skill covers system and API migration (Strangler, Adapter, Feature Flag) but has no guidance for schema evolution — the failure mode most likely to cause downtime or data loss during a deploy. This section adds the expand→migrate→contract pattern with backward/forward compatibility guidance across rolling deploys, plus rationalization rows and red flags specific to schema changes.

## Why here

Schema migration is a specialised case of the Strangler pattern the skill already uses as its primary frame: old shape and new shape live in parallel until all consumers move, then the old shape is removed. Putting it under `Migration Patterns` reuses that vocabulary and keeps the expand/contract steps visually parallel to what's already there.

## Non-duplication check

`incremental-implementation` covers slice-by-slice feature delivery, not schema evolution. `ci-cd-and-automation` covers pipeline configuration, not schema state. This section cross-links `incremental-implementation` for the build-slice framing but teaches nothing that skill teaches.

## Checklist

- [x] Touches only `skills/deprecation-and-migration/SKILL.md`
- [x] Rationalization rows, red flags, and verification checkboxes added in host skill's style

### DIFF
--- a/skills/deprecation-and-migration/SKILL.md
+++ b/skills/deprecation-and-migration/SKILL.md
@@ -161,6 +161,34 @@ function getTaskService(userId: string): TaskService {
 }
 ```
 
+### Database Schema Migrations (Expand/Contract)
+
+A schema change is the riskiest migration because the data is the one thing you cannot roll back by reverting a deploy. The failure mode is coupling the schema change to the code change: rename a column in the same release that starts using the new name, and during the rollout window — when old and new code run at once — one of them is querying a column that doesn't exist. The fix is to **never change a column in place**. Migrate in additive phases so old and new code are both valid at every step.
+
+```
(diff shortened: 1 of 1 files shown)


addyosmani (approved):
Glad to see expand/contract spelled out with a real rename walkthrough, because "never change a column in place" is one of those rules everyone nods at and then forgets the moment a deadline shows up. Calling out that each phase stays independently reversible is the part that actually saves someone at 2am. Good addition.

PR #346, opened 2026-07-05
docs(code-review): add dependency upgrade workflow to Dependency Discipline

HMAKT99:
## What this adds

An upgrade workflow appended to the existing `## Dependency Discipline` section in `skills/code-review-and-quality/SKILL.md`. The current section is an *adopt* gate (five questions to ask before adding a dependency). This extension adds the missing *upgrade* path.

## The gap it fills

Once a dependency is in the project, the skill has no guidance on when or how to upgrade it — breaking-change triage, staged upgrade sequencing, lockfile discipline, and transitive-dependency risk. Teams regularly defer upgrades until they're painful; a concrete process reduces that drift.

## Non-duplication check

`security-and-hardening` already owns `npm audit` results and supply-chain hygiene. This section explicitly does **not** re-teach those topics — it cross-links the security skill instead and focuses only on the upgrade workflow itself (changelog triage, staged one-at-a-time upgrades, lockfile verification, before/after test gates).

## Checklist

- [x] Fully additive — extends an existing section, no edits to existing prose
- [x] Touches only `skills/code-review-and-quality/SKILL.md`
- [x] Rationalization row, red flags, and verification checkbox added in host skill's style

### DIFF
--- a/skills/code-review-and-quality/SKILL.md
+++ b/skills/code-review-and-quality/SKILL.md
@@ -289,6 +289,16 @@ Part of code review is dependency review:
 
 **Rule:** Prefer standard library and existing utilities over new dependencies. Every dependency is a liability.
 
+**Upgrading an existing dependency** is a code change like any other, and the riskiest upgrades are the ones merged in bulk with a message like "bump deps." Review them with the same discipline:
+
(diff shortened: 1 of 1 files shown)


addyosmani (approved):
The one-dependency-per-change point earns its keep on its own, since bulk bump PRs are precisely where a build breaks quietly and nobody can tell which package did it. Keeping the security verdict over in the hardening skill rather than re-arguing it here keeps the boundary clean too. Happy to take this.

PR #347, opened 2026-07-05
docs(security): add data privacy & compliance section

HMAKT99:
## What this adds

A `## Data Privacy & Compliance` section in `skills/security-and-hardening/SKILL.md`, covering PII classification, data minimisation, retention/deletion policy, consent, cross-border transfer, and breach response.

## The gap it fills

`security-and-hardening` covers authentication, injection prevention, supply-chain hygiene, and secrets management well. It has no guidance on data privacy or compliance — a separate discipline that's increasingly a first-class engineering concern (GDPR, CCPA, HIPAA). Engineers making decisions about what to log, how long to store it, and when to delete it have nowhere in the skill pack to look.

## Why here

Privacy is a dimension of hardening: minimising what data you hold reduces your attack surface. The PII classification table mirrors the threat-model framing the skill already uses, so it fits naturally.

## Non-duplication check

`observability-and-instrumentation` covers what to log for debugging, not what to *avoid* logging for compliance. The privacy section cross-links it to flag the PII-in-logs failure mode without duplicating the instrumentation guidance.

## Checklist

- [x] Touches only `skills/security-and-hardening/SKILL.md`
- [x] Rationalization rows, red flags, and verification checkboxes added in host skill's style

### DIFF
--- a/skills/security-and-hardening/SKILL.md
+++ b/skills/security-and-hardening/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: security-and-hardening
-description: Hardens code against vulnerabilities. Use when handling user input, authentication, data storage, or external integrations. Use when building any feature that accepts untrusted data, manages user sessions, or interacts with third-party services. Use when collecting, storing, retaining, deleting, or sharing personal data, or addressing privacy/compliance (GDPR/CCPA).
+description: Hardens code against vulnerabilities. Use when handling user input, authentication, data storage, or external integrations. Use when building any feature that accepts untrusted data, manages user sessions, or interacts with third-party services. Use when personal data or privacy compliance (GDPR, CCPA) is involved.
 ---
 
 # Security and Hardening
--- a/skills/security-and-hardening/SKILL.md
+++ b/skills/security-and-hardening/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: security-and-hardening
-description: Hardens code against vulnerabilities. Use when handling user input, authentication, data storage, or external integrations. Use when building any feature that accepts untrusted data, manages user sessions, or interacts with third-party services.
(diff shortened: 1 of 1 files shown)


addyosmani:
The content here is excellent. The privacy-is-not-security framing and the PII classification table are exactly right and I want them in. One fix first: broadening the security description with "data / storing / sharing" language pulled its routing rank down two points in the evals, because those generic words now over-trigger security for other skills' prompts. Trim the frontmatter addition back to the essentials and leave the body as-is, then it's a clean merge.

HMAKT99:
Glad the content landed right. Fixed the frontmatter — replaced the long verb list (`collecting, storing, retaining, deleting, sharing`) with a tight trigger: `Use when personal data or privacy compliance (GDPR, CCPA) is involved.` Same intent, no generic words that bleed into unrelated routing. Body is untouched. Validator passes 0/0.

nucliweb:
**Re-ran the eval suite on a local test-merge; the routing regression is cleared.** Merged into current `main` and ran `node scripts/run-evals.js`:

| | checks | errors | trigger rank-1 |
|---|---|---|---|
| `main` baseline | 124 | 0 | 86% (65/76) |
| `main` + this PR | 124 | 0 | 86% (65/76) |

Identical to baseline. The trimmed trigger in `a93fb23` does the job: the description change is additive-only, so the sole risk was `security-and-hardening` over-triggering on other skills' prompts, which is exactly the negative-prompt check the runner enforces, and it passes with no violations. The two-point drop from the original verb list is gone.

Content and boundaries hold: privacy is scoped as a dimension of hardening (minimize data = shrink attack surface), it cross-links `observability-and-instrumentation` for the don't-log-PII side rather than repeating it, and both referenced skills resolve. Validator 0/0, CI green. Ready from my side.

addyosmani:
Thanks. Data privacy and compliance fit security-and-hardening rather than a separate section; closing here, a focused addition to that skill would be welcome.

PR #349, opened 2026-07-06
docs: clarify plugin install login requirement

### LINKED ISSUE #343: /plugin install everything-claude-code@everything-claude-code
Hi,

I am trying to install the subject plugin direct at my termaminal but I got the below message 

Not logged in · Please run /login

Image

ketpatil77:
Closes #343`n`nClarifies that `/plugin install` relies on an authenticated Claude Code session and adds a local-path fallback for users who hit `Not logged in` during setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,8 @@ Prefer a native integration? Pick your tool below.
 /plugin install agent-skills@addy-agent-skills
 ```
 
+> **Seeing `Not logged in`?** `/plugin install` uses your current Claude Code session. Run `/login` first, then retry the install command.
+
 > **SSH errors?** The marketplace clones repos via SSH. If you don't have SSH keys set up on GitHub, either [add your SSH key](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account) or use the full HTTPS URL to force HTTPS cloning during the marketplace-add step:
 > ```bash
 > /plugin marketplace add https://github.com/addyosmani/agent-skills.git


abhisheksharma2411 (commented):
Still applies, and I think there's a better argument for it than the description makes — worth saying because the issue it cites was closed as out-of-scope, which makes this look stale when it isn't.

**The gap is real and still open.** `grep -n "Not logged in\|/login\|logged in" README.md docs/getting-started.md` returns nothing on current `main`. Nothing in the repo tells a user what `Not logged in` means during install.

**#343 being closed doesn't undercut it.** It was closed 10 Jul because the reporter was installing `everything-claude-code`, a different marketplace entry — @addyosmani's own reply was *"the actual error, 'Not logged in, run /login', is Claude Code telling you to authenticate first."* So the diagnosis was confirmed by the maintainer; only the reporter's plugin was out of scope.

**The precedent is directly beneath this diff.** The note lands immediately above:

> **SSH errors?** The marketplace clones repos via SSH. If you don't have SSH keys set up on GitHub…

That's also not an `agent-skills` defect — it's a GitHub prerequisite — and it's documented anyway, because users hit it during this exact install step. A Claude Code session prerequisite at the same step is the same shape of note, in the same place, for the same reason. The two read as a pair.

Two small things:

- **The description contains a literal `` `n`n ``** — `Closes #343\`n\`nClarifies that…`. Looks like a PowerShell escape that didn't expand. The closing reference did register (`closingIssuesReferences` resolves to #343), so nothing is broken, but it'll carry into the squash-merge message. Worth a quick edit.
- **Wording nit, take it or leave it:** "uses your current Claude Code session" is slightly indirect for someone staring at an error. "`/plugin install` requires an authenticated Claude Code session" states the prerequisite rather than the mechanism. Either is fine.

Mergeable as-is (`mergeable=true`), two months on. No conflict with the surrounding section.

(Not a maintainer, so this is a reviewer note rather than an approval.)

PR #359, opened 2026-07-07
Improve Playwright locator examples

debs-obrien:
## Summary
- Replace CSS/text selectors in the Playwright E2E example with accessibility-first locators
- Scope task assertions/actions to the task list item
- Use role and label based locators to better reflect Playwright best practices

## Testing
- Not run; documentation-only change

### DIFF
--- a/references/testing-patterns.md
+++ b/references/testing-patterns.md
@@ -202,23 +202,22 @@ import { test, expect } from '@playwright/test';
 test('user can create and complete a task', async ({ page }) => {
   // Navigate and authenticate
   await page.goto('/');
-  await page.fill('[name="email"]', 'test@example.com');
-  await page.fill('[name="password"]', 'testpass123');
-  await page.click('button:has-text("Log in")');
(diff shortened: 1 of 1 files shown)


nucliweb:
Hola @debs-obrien 👋

Thank you for the contribution ☺️

federicobartoli:
Optional tiny docs-shaping thought: for the task row, would it be a little closer to the Playwright list examples to write this as:

```ts
const task = page.getByRole('listitem').filter({ hasText: 'Buy groceries' });
```

That keeps the scoped-row pattern while avoiding any reliance on the `listitem`'s accessible name. The rest looks great to me: this is a clear improvement toward user-facing locators and scoped actions.

LGTM either way.

addyosmani:
LGTM, and hard to argue Playwright locator style with you of all people, @debs-obrien. Role and label based locators are exactly what we should be modeling in the reference (they test what users actually perceive), and scoping the checkbox and assertion to the list item is a nice touch that quietly kills a whole class of strict-mode flake.
Thanks for making the example something agents can safely imitate. Merging.

PR #195, opened 2026-05-22
feat: replace plain-text confirmations with question tool in 10 skills

prullanferragut:
## Summary

- Replaces 20 plain-text `yes/no/refine?` prompts across 10 skill files with instructions to use OpenCode's `question` tool, which renders structured option lists with keyboard navigation
- Replaces the legacy `AskUserQuestion` reference in `idea-refine` with `question`
- 8 of the 19 originally listed skills had no genuine pause-and-ask points and are unchanged

## Skills Updated

`debugging-and-error-recovery`, `code-review-and-quality`, `incremental-implementation`, `using-agent-skills`, `spec-driven-development`, `doubt-driven-development`, `ubiquitous-language`, `context-engineering`, `idea-refine`, `interview-me`

## Approach

Skills remain platform-agnostic — no JSON or tool-call syntax was added to any SKILL.md. Each change adds natural prose instructing the agent to use the `question` tool with named options at the relevant decision point. Code block examples showing agent output (e.g., `Yes / no / refine?` in `interview-me`) are left intact.

## Notes

- `docs/workflow-tracker-todo-screenshot.md` is a placeholder for a screenshot taken during this session — it can be reverted independently
- Spec: `docs/specs/spec-question-tooling.md`
- Plan: `docs/plans/plan-question-tooling.md`

Issue #352, opened 2026-07-06
Evals: graduate Tier 3 to trusted, then ratchet the deterministic gates


addyosmani:
Follow-ups from #342 review (thanks @federicobartoli and @nucliweb), bundled per nucliweb's suggestion since they're all the same shape: observational today, gate tomorrow.

**Graduate Tier 3 from provisional to trusted**
- Author real fixtures under `evals/fixtures/` for each behavioral eval (`files[]` is wired in the runner; content doesn't exist yet), so evals measure performing the skill, not describing it.
- Remove `trust_level: "provisional"` per eval as its fixtures land. Until then Tier 3 results are sanity checks, not evidence.
- Add pressure-test scenarios (time pressure, sunk cost, authority) for discipline skills once trusted.

**Ratchet the deterministic tier**
- `--min-rank1 <pct>` CI flag once the baseline stabilizes (current: 85%), so routing quality can't silently regress; consider requiring one `top_k: 1` signature prompt per skill.
- Promote the per-case minimums check (3 positive / 2 negative / 1 behavioral) from warning to error.
- Promote missing-eval-file coverage from warning to error, and update the CONTRIBUTING note that currently says "warning for now".

Target: after the in-flight skill PRs (#261, #253, #271, etc.) land or close, so we don't break them mid-air.

nucliweb:
One sequencing note on the **Graduate Tier 3** block: it assumes the behavioral runner works end-to-end with real fixtures, but two latent bugs in `run-evals.js` (flagged in #342) will surface the moment the first real fixture is authored, so they're prerequisites for that block rather than separate work:

1. **Grader hits `ARG_MAX`.** `graderPrompt` embeds the full executor trace (`stream-json`, up to the 64MB `maxBuffer`) and passes it as an argv entry to `execFileSync('claude', ['-p', graderPrompt])`. Any non-trivial trace exceeds the OS argument limit (`E2BIG`) and throws before grading. Needs to move to stdin (`{ input: graderPrompt }`).

2. **Executor sets no tool-permission mode.** The executor `claude -p` runs without a permission flag or `--allowedTools`, so in headless mode file edits and command runs can be denied, meaning the agent narrates the skill instead of performing it, which is exactly what the fixtures + trace-grading are meant to measure.

[...]

Issue #351, opened 2026-07-06
Description vocabulary gaps: frontend-ui-engineering, performance-optimization


addyosmani:
Surfaced by the eval framework in #342 (fair trigger prompts could not rank these skills because the descriptions lack the vocabulary users actually say):

- `frontend-ui-engineering`: description never says "responsive", "accessible", "WCAG", or "page". A user asking "make this page responsive and WCAG compliant" may not route here.
- `performance-optimization`: description never mentions queries, N+1, or databases. "Optimize this N+1 query" may not route here.

Fix: enrich both descriptions (keeping the what-then-Use-when format), then:
1. restore the stricter eval prompts that were softened in #342 to stay fair,
2. add the deferred `owner: frontend-ui-engineering` to the api-and-interface-design negative ("Make this landing page responsive on mobile"),
3. expect the trigger rank-1 rate to rise above the 85% baseline.

The evals will verify the fix mechanically.

PR #360, opened 2026-07-08
feat(interview-me): add opt-in collaborative mode

### LINKED ISSUE #251: Proposal: add a collaborative mode to `interview-me`
## Background

The current `interview-me` skill is extraction-shaped: the agent's job is to surface what the user actually wants, with hypotheses and guesses serving the interview rather than the user's decision space. That's the right default. The skill explicitly warns against listing options ("listing widens the search; asking narrows it") and pushes the agent to converge by probing.

What is missing is a mode for cases where the developer wants the agent as an **interlocutor**, not just as an interviewer. The 95% confidence then arises from consensus between user and agent, not from the agent pulling intent out of the user. In practice this happens often: the developer has shaped a question, the agent has accumulated repo context, and the most useful next move is for the agent to **propose**, with risks attached, and let the conversation converge through that.

This proposal adds that as an opt-in mode inside `interview-me`, not as a separate skill. Keeping it in `interview-me` avoids duplicating the Process and reuses the 95% stop condition.

## Anti-goals

- Not a default. Extraction-first must stay the default; the anchoring risk is real and the current skill is explicit about it.
- Not a way to skip the interview. The PROPOSAL step lives inside the interview, after at least one round of user signal, never on turn 1.
- Not a license to list options. The mode keeps the "no widening" rule: 2 to 3 options max, always with the agent's lean stated.

### Activation (three entries, none automatic)

1. **Explicit**: `interview-me --collaborative`, or natural-language synonyms ("give me your take", "propose something", "what would you do here?", "let's think together").
2. **Promoted from the normal flow**: after Step 3, if the agent has a defensible opinion, it offers to enter the mode: *"I have a take on this. Want to hear it before I restate, or just restate?"* The user opts in.
3. **Recovery from delegation**: when the user says *"whatever you think"*, instead of the current re-ask with two options (Step 5), the collaborative mode lets the agent propose with a lean and ask for consensus.

### New Step 3.5: Propose, with lean and risk

Inserted between Step 3 (listen) and Step 4 (restate). Only appears when the mode is active.

Format for a single proposal:

```
PROPOSAL: <one sentence — what the agent thinks is the right move>
WHY:      <one line — what trade-off this resolves, what info it draws on>
RISK:     <one line — the assumption that could be wrong, or where context might be missing>
LEAN:     <strong / soft>
ASKING:   Does this match what you'd land on, or does it miss something?
```

Format for 2 to 3 options:

```
Two paths I'd consider:
  A. <option> — solves X, but constrained by Y
  B. <option> — solves Z, but costs W
I lean toward A because <reason>. Where do you land?
```

Rules:
- Max 3 options.
- LEAN is mandatory. "Either could work" is delegation in disguise.
- One PROPOSAL per turn. No stacking.
- Differential confidence: separate what the agent knows (constraints, repo facts) from what is judgment.

### Change in Step 4: the restate is negotiated, not extracted

In collaborative mode, each line of the restate is tagged with origin:

```
- Outcome:      <line>             ← from your answer
- Success:      <line>             ← from your answer
- Constraint:   <line>             ← my proposal in 3.5, you accepted
- Out of scope: <line>             ← my push, you confirmed
```

This prevents the agent from laundering its own proposals as the user's intent and leaves an honest trace of the conversation.

### Stop condition: anti-anchoring sub-check

The 95% rule stays. An additional sub-check is applied at the stop:

> *"Can I distinguish what the user actually wanted from what they accepted because I proposed it well?"*

If not, anchoring is suspected. The agent goes back to Step 3 with: *"To check I haven't pulled you, if I hadn't proposed X, what would you have said?"*

Triggered earlier if the user accepted more than one PROPOSAL with no pushback or refinement.

### New Red Flags (collaborative-mode-specific)

- Proposing on turn 1, before any user signal
- Proposing without a LEAN, or saying "either works"
- More than 3 options in a single PROPOSAL
- Accepting the user's yes to a PROPOSAL without the anti-anchoring check
- Proposing in a domain where the user has clear expertise advantage (the agent should defer, not propose)
- Restate that does not mark the origin of each line

### New Common Rationalizations

| Rationalization | Reality |
|---|---|
| "I have context, I should just propose" | Propose with rationale AND a risk. A naked recommendation hides assumptions. |
| "If I state a lean, I bias them" | A lean is what makes the proposal useful. The user can push back on a lean; they cannot push back on "either is fine". |
| "More options give them choice" | 2 to 3 max. More is widening, which is the failure mode the skill exists to avoid. |
| "They said yes to my proposal, we are done" | Run the anti-anchoring check. Would they have said the same thing if you had not proposed it? |

### Frontmatter description (delta)

Append to the existing description:

> *"Supports a collaborative mode (`interview-me --collaborative`, or via 'give me your take' / 'propose something') where the agent contributes proposals with explicit leans and risks, reaching the 95% threshold through consensus rather than pure extraction."*

## Interaction with neighboring skills

- **`grill-me`**: challenges an existing plan. Collaborative `interview-me` challenges the solution space **before** a plan exists. Clean boundary: plan written → grill-me; intent not fixed → interview-me (collaborative).
- **`idea-refine`**: in this mode, light divergence happens inside the interview. Deep divergence still belongs in idea-refine downstream.

## Questions for the maintainer

1. Does this fit `interview-me`'s scope, or does it dilute the extraction discipline that makes the skill work today?
2. Is implicit promotion (the agent offering to enter the mode after Step 3) acceptable, or should activation be explicit-only to keep the skill predictable?
3. Any concern about overlap with `grill-me` once the mode exists?

Happy to send the PR with the SKILL.md change once direction is agreed. The change is contained: new Step 3.5, deltas in Step 4 and stop condition, additions to Red Flags / Common Rationalizations / Verification, and a one-line addition to the frontmatter description.

kevglynn:
## Summary

Implements the collaborative mode proposed in #251 inside the existing `interview-me` skill. The skill's default remains extraction-first — one question at a time, guess attached, no option-listing. What's new is an opt-in mode for when the user wants the agent as an interlocutor rather than an interviewer: the agent may propose, with an explicit lean and risk attached, and the ~95% confidence threshold is reached through consensus instead of pure extraction.

Concretely, following the design in #251:

- **New "Collaborative Mode (opt-in)" section** — three entry points (explicit invocation, offered-and-accepted after Step 3, recovery from "whatever you think" delegation), plus the anti-goals: never the default, never turn 1, never more than 2–3 options.
- **New Step 3.5** — the `PROPOSAL / WHY / RISK / LEAN / ASKING` format, the 2–3-option variant with a mandatory lean, one proposal per turn, and the expertise-deference rule.
- **Step 4 delta** — in collaborative mode the restate is origin-tagged per line (user's answer vs. accepted proposal vs. confirmed push), so the agent can't launder its proposals as user intent.
- **Stop-condition sub-check** — the anti-anchoring question ("can I distinguish what they wanted from what they accepted because I proposed it well?"), triggered early when more than one PROPOSAL is accepted without pushback.
- **Four new rationalizations, six new red flags, four new verification items** — all collaborative-mode-scoped, appended without touching the existing entries.
- **Frontmatter description** gains the collaborative trigger vocabulary ("give me your take", "propose something", "what would you do here?", "let's think together"). Description is 779 chars, under the 1024 limit.
- **Eval case** (`evals/cases/interview-me.json`) gains two positive triggers and one behavioral eval asserting the mode's contract (no turn-1 proposal, lean+risk present, origin-tagged restate, anti-anchoring check). One trigger uses organic phrasing ("I keep flip-flopping on what I actually want here — weigh in with your own proposal and the risks you see"); the other exercises the explicit catchphrase invocations.

Two deliberate deviations from the issue text:

1. It references a `grill-me` skill for the boundary discussion; this repo's equivalent is `doubt-driven-development`, so the boundary line was added there ("plan written → doubt-driven review; intent not fixed → collaborative interview").
2. The design's entry point 1 included an explicit `interview-me --collaborative` flag invocation; this implementation keeps only the natural-language synonyms, since skills in this repo are invoked by intent matching, not flags.

## Non-duplication check

- No open PR implements #251. The only open PR touching `skills/interview-me/SKILL.md` is #195 (replaces plain-text confirmations with the question tool across 10 skills) — different scope; worst case is a small merge conflict in Step 5.
- Searched open PRs for "interview" and "collaborative": no other hits.
- No existing skill covers this: `idea-refine` is downstream divergence, `doubt-driven-development` is post-plan review. The mode is added inside `interview-me` precisely to avoid a near-duplicate skill, per CONTRIBUTING's "prefer a focused edit over a new directory."

## Test plan

- `node scripts/validate-skills.js` → 24 skills checked — 0 errors, 0 warnings — PASSED
- `node scripts/run-evals.js` → 122 checks passed — 0 errors, 0 warnings; trigger rank-1 rate 85% (63/74) — both new positive triggers rank `interview-me` first (rate unchanged from the 85% baseline, 61/72 before). The organic-phrasing trigger ranking first is the meaningful signal here; the catchphrase trigger confirms the explicit invocations route as intended. The behavioral eval is `provisional` (no fixtures yet), so it documents the mode's contract rather than evidencing it.
- `SKILL.md` is 311 lines, under the 500-line cap; frontmatter valid.

Closes #251.

### DIFF
--- a/evals/cases/interview-me.json
+++ b/evals/cases/interview-me.json
@@ -48,6 +48,7 @@
     },
     {
       "id": 2,
+      "kind": "dialogue",
       "prompt": "I need a caching layer for our API but I'm torn on the approach. Interview me, and give me your take as we go — I want your proposals, not just questions.",
       "expected_output": "An interview that enters collaborative mode on the user's explicit invitation: proposals carry an explicit lean and risk, the restate tags each line's origin, and the agent checks it hasn't anchored the user",
       "expectations": [
@@ -55,8 +56,7 @@
         "Each proposal states a lean and a risk, with at most 3 options in a single turn",
         "The final restate marks which lines came from the user's answers and which from accepted proposals",
         "Before stopping, the agent runs an anti-anchoring check on accepted proposals"
-      ],
-      "trust_level": "provisional"
+      ]
     }
   ]
 }
--- a/evals/cases/interview-me.json
+++ b/evals/cases/interview-me.json
@@ -15,7 +15,7 @@
         "top_k": 3
       },
       {
-        "prompt": "I've been going back and forth on this feature — give me your take and let's think together before anything gets built",
+        "prompt": "I keep flip-flopping on what I actually want here — weigh in with your own proposal and the risks you see before we settle on anything",
         "top_k": 3
       },
       {
--- a/evals/cases/interview-me.json
+++ b/evals/cases/interview-me.json
@@ -13,6 +13,14 @@
       {
         "prompt": "Ask me one question at a time until you understand the feature",
         "top_k": 3
+      },
+      {
+        "prompt": "I've been going back and forth on this feature — give me your take and let's think together before anything gets built",
+        "top_k": 3
+      },
+      {
+        "prompt": "What would you do here? Propose something for this requirement and tell me where you'd push back",
+        "top_k": 3
       }
     ],
     "negative": [
@@ -37,6 +45,18 @@
         "The agent does not propose solutions before understanding the need",
         "The interview surfaces the underlying goal behind the stated ask"
       ]
+    },
+    {
+      "id": 2,
+      "prompt": "I need a caching layer for our API but I'm torn on the approach. Interview me, and give me your take as we go — I want your proposals, not just questions.",
+      "expected_output": "An interview that enters collaborative mode on the user's explicit invitation: proposals carry an explicit lean and risk, the restate tags each line's origin, and the agent checks it hasn't anchored the user",
+      "expectations": [
+        "The agent does not propose on turn 1; at least one round of user signal comes first",
+        "Each proposal states a lean and a risk, with at most 3 options in a single turn",
+        "The final restate marks which lines came from the user's answers and which from accepted proposals",
+        "Before stopping, the agent runs an anti-anchoring check on accepted proposals"
+      ],
+      "trust_level": "provisional"
     }
   ]
 }
--- a/skills/interview-me/SKILL.md
+++ b/skills/interview-me/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: interview-me
-description: Extracts what the user actually wants instead of what they think they should want. Achieves this through one-question-at-a-time interview until ~95% confidence about the underlying intent. Use when an ask is underspecified ("build me X" without "for whom" or "why now"), when the user explicitly invokes ("interview me", "grill me", "are we sure?", "stress-test my thinking"), or when you catch yourself silently filling in ambiguous requirements before any plan, spec, or code exists.
+description: Extracts what the user actually wants instead of what they think they should want. Achieves this through one-question-at-a-time interview until ~95% confidence about the underlying intent. Use when an ask is underspecified ("build me X" without "for whom" or "why now"), when the user explicitly invokes ("interview me", "grill me", "are we sure?", "stress-test my thinking"), or when you catch yourself silently filling in ambiguous requirements before any plan, spec, or code exists. Supports an opt-in collaborative mode — invoked with "give me your take", "propose something", "what would you do here?", or "let's think together" — where the agent contributes proposals with explicit leans and risks, reaching the ~95% threshold through consensus rather than pure extraction.
 ---
 
 # Interview Me
@@ -34,6 +34,24 @@ Apply this skill when:
 ## Loading Constraints
 
 This skill needs a live, responsive user. **Do not invoke in non-interactive contexts** like CI pipelines, scheduled runs, `/loop`, or autonomous-loop. If you're in one of those and the ask is underspecified, flag that as a blocker for the user instead of guessing.
+
+## Collaborative Mode (opt-in)
+
+By default this skill is extraction-shaped: your hypotheses and guesses serve the interview, not your own preferences. Collaborative mode is for when the user wants you as an **interlocutor**, not just an interviewer — they've shaped a question, you've accumulated repo context, and the most useful next move is a proposal with risks attached. The ~95% confidence then arises from consensus between user and agent, not from extraction alone.
+
+The mode is never automatic. Three ways in:
+
+1. **Explicit**: the user invokes it ("give me your take", "propose something", "what would you do here?", "let's think together").
+2. **Promoted from the normal flow**: after Step 3, if you have a defensible opinion, offer it: *"I have a take on this. Want to hear it before I restate, or just restate?"* The user opts in.
+3. **Recovery from delegation**: when the user says "whatever you think" in Step 5, propose with a stated lean and ask for consensus instead of only re-asking with two options.
+
+What the mode is **not**:
+
+- Not the default. Extraction-first stays the baseline; the anchoring risk of proposing early is real.
+- Not a way to skip the interview. Proposals live inside the interview, after at least one round of user signal — never on turn 1.
+- Not a license to list options. The "no widening" rule holds: 2–3 options max, always with your lean stated.
+
+When active, the mode adds Step 3.5, changes the shape of the restate in Step 4, and adds an anti-anchoring sub-check to the stop condition. Everything else is unchanged.
 
 ## The Process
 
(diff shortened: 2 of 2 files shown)


nucliweb (commented):
Thanks for picking this up, @kevglynn, and for flagging it for me. I read the diff line by line against the #251 design; this is a faithful, near-verbatim implementation of the spec. Step 3.5, the origin-tagged restate, the anti-anchoring sub-check, the four rationalizations, six red flags, and four verification items all match the design as written. Both repo adaptations are correct calls.

### On the open questions

- **Q2 (implicit promotion):** keep entry point 2. It is the most useful part of the mode, and it is safely gated: the agent never enters unilaterally, and the "or just restate" fallback keeps extraction the default. No change wanted here.
- **Q3 (grill-me):** confirmed. There is no `grill-me` in this repo (the one users may know is from a different skill set), so anchoring the boundary on `doubt-driven-development` is the right move.
- **Q1 (dilution):** agreed. No reachable path skips extraction, and the anti-anchoring check sends the agent back to Step 3, so if anything the stop condition is stricter. That was the intent.

### Two adjustments before merge

1. **Call out the dropped `interview-me --collaborative` invocation.** The design listed the explicit flag as entry point 1; the PR drops it in favor of NL synonyms only. I think that is the right choice for this repo (skills here are not flag-invoked), but it is a real deviation from the spec and is not in your "one deliberate deviation" note. Worth adding it there so the record is clean.

2. **De-circularize one positive eval prompt.** Both new positive prompts embed the exact catchphrases just added to the description ("give me your take" + "let's think together"; "what would you do here?" + "propose something"). The evals README warns against copying the description because it games the trigger check: the TF-IDF ranks `interview-me` first largely because those tokens were just added to its description, so "both new prompts rank first" is partly tautological rather than independent evidence. Suggest rephrasing at least one to express the collaborative intent without the catchphrases (e.g. "I keep flip-flopping on this, weigh in with what you would pick and why") so it exercises organic routing.

### Note, not a change request

The behavioral eval is `provisional` with no fixtures, so CI only checks lexical routing, not the mode's actual contract (no turn-1 proposal, lean+risk, origin tags, anti-anchoring). That is inherent to tier 3 until fixtures land (#352), just worth not citing that eval as evidence the mode behaves correctly yet.

Net: technically mergeable, faithful to the design. The two adjustments above are small. Nice work.

kevglynn:
Both adjustments are in.

1. **Dropped flag documented.** The PR description now lists two deliberate deviations — the `doubt-driven-development` boundary and the dropped `interview-me --collaborative` flag (kept NL synonyms only, since skills here route by intent matching, not flags). You're right that it belonged in the record; it was a silent call at implementation time.

2. **First trigger de-circularized** (4e12016). Now reads: *"I keep flip-flopping on what I actually want here — weigh in with your own proposal and the risks you see before we settle on anything."* One data point from getting it to route: a fully organic phrasing without any description vocabulary ("I keep flip-flopping on this design decision — weigh in with what you'd pick…") ranked 5th, behind `api-and-interface-design`. The version that landed shares only "what I actually want" (pre-existing description vocabulary, not the new catchphrases) plus the naturally-overlapping "proposal"/"risks", and ranks first. That gap is arguably a milder instance of the #351 pattern — collaborative asks phrased without any interview vocabulary don't route here — but widening the description further felt like the wrong trade against collision risk, so I left it. Flagging in case you read it differently.

Second trigger intentionally keeps the catchphrases: it now serves as the check that the explicit invocations route as intended, while the first provides the organic evidence. Rank-1 rate holds at 85% (63/74).

On the behavioral eval: agreed — it's `provisional` with no fixtures, so it documents the mode's contract rather than evidencing it. The PR description now says exactly that. Happy to write fixtures for it once the #352 prerequisites land.

addyosmani:
Really like this. Collaborative mode fills the gap where the user wants an interlocutor rather than an interrogator, and keeping it opt-in with explicit entry phrases means it can't hijack the default extraction behavior. The origin-tagged restate and the anti-anchoring check are the thoughtful touches. This delivers something I've had on the roadmap, so I want to read the new section end to end and then land it.

addyosmani:
Did the end-to-end read I promised, and it holds up well. The three entry paths (explicit, promote-from-flow, recover-from-delegation), the differential-confidence rule, and the don't-propose-in-the-user's-domain guardrail are the thoughtful bits, and keeping it strictly opt-in with proposals never on turn 1 is what stops it wrecking the default extraction behavior. Structurally clean, CI's green.

One coordination thing, not a flaw in this PR: your eval2 is provisional with no fixture, which is fine today but collides with #381 (the eval-graduation PR) the moment it lands, since that flips provisional evals to a hard CI error. interview-me is genuinely awkward to fixture since it's a dialogue, so let's either add a light fixture here or settle the conversation-skill exemption in #381 first, then merge these in the right order. Content-wise I'm ready to take it.

addyosmani:
Quick unblock for the eval collision I flagged: I've proposed a `kind: "dialogue"` exemption on #381 (tested, suite stays green) so conversation skills don't need a code fixture. Once that lands, your eval2 just becomes `kind: "dialogue"` and drops the provisional flag, nothing to fixture. So content's ready and this is good to go pending that merge ordering, nothing more needed from you here.

kevglynn:
Adopted the `kind: "dialogue"` exemption ahead of time in `fe390b8` — eval 2 is now `kind: "dialogue"` with the provisional flag dropped. The current runner ignores the `kind` field (full suite still green locally), so this PR and #381 can land in either order now; no merge sequencing needed. I left eval 1 alone since #381 already converts it on main, and touching it here would just create a conflict.

kevglynn:
Rebased onto current `main` to clear the merge conflict from #381. Resolution: kept upstream's eval 1 as `kind: \"dialogue\"` (no provisional), and our eval 2 is the same shape. Collaborative triggers and the skill body are unchanged. Locally: validate-skills and run-evals both green, no remaining conflicts with `main`.

kevglynn:
Hey @nucliweb — following up since this has been sitting since the Jul 18 rebase. Both adjustments you flagged (the deviation note, the de-circularized eval prompt) are in, described in the comments above. Let me know if there's anything else you want changed, or if it's good to merge.

PR #181, opened 2026-05-18
fix: rename /review to /code-review and delegate persona framework to skill

googlarz:
Closes #95. Closes #172.

## Problem

Two related issues that both stem from the same design gap:

**#95 — Command collision:** agent-skills' `/review` command shadowed Claude Code's built-in `/review`. Because Claude Code resolves plugin before built-in, the built-in becomes unreachable after installing this plugin. First-run experience for Claude Code users was broken.

**#172 — Routing ambiguity + duplication:** `agents/code-reviewer.md` inlined its own copy of the five-axis framework from `skills/code-review-and-quality/SKILL.md`. Two sources of truth caused wording drift and left agents confused about which entry point to use for a "review this" intent.

## Solution

**Command rename:** `.claude/commands/review.md` → `code-review.md` (and Gemini equivalent). The skill slugs in this repo are already long-form (`code-review-and-quality`) — only the slash command layer had a collision. `/code-review` is unambiguous and consistent.

**Persona delegation:** Replaced 37 lines of copy-pasted framework in `agents/code-reviewer.md` with a 4-line delegation directive pointing at the skill. Added a fallback for contexts where the skill file isn't loaded. Output format, rules, and composition sections are unchanged — those are legitimately persona-specific.

## Files changed

| File | Change |
|------|--------|
| `.claude/commands/review.md` | Renamed → `code-review.md` |
| `.gemini/commands/review.toml` | Renamed → `code-review.toml` |
| `agents/code-reviewer.md` | Remove inlined 5-axis framework; delegate to skill |
| `README.md` | `/review` → `/code-review` in commands table and diagram |
| `CLAUDE.md` | `/review` → `/code-review` in slash commands list |
| `agents/README.md` | `/review` → `/code-review` in decision matrix and examples |
| `docs/gemini-cli-setup.md` | `/review` → `/code-review` in commands table |
| `references/orchestration-patterns.md` | All `/review` references updated |

## Test plan

- [ ] `/code-review` is discoverable in Claude Code after plugin install
- [ ] Claude Code's built-in `/review` is no longer shadowed
- [ ] `agents/code-reviewer.md` no longer duplicates the five-axis framework
- [ ] `grep -rn '/review' . --include='*.md' --include='*.toml' | grep -v 'security-review\|code-review'` returns empty

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #362, opened 2026-07-08
ci: label PRs that add a skill duplicating an open PR

### LINKED ISSUE #315: CI: label PRs that add a skill duplicating an open PR
## Context

PR #313 adds an instruction-level guardrail (`.claude/rules/skills-contributing.md` + CONTRIBUTING.md pre-flight checks) that steers agents to check the catalog and open PRs before adding a skill. The test in #313 shows it catches duplicates, but it relies on the agent actually following the instructions.

This issue tracks a deterministic, low-noise CI complement: flag when a PR adds a new skill that collides with another **open** PR.

## Proposal

A GitHub Actions workflow that:

1. Determines the new skill directories this PR adds (files `skills/<name>/SKILL.md` with status `added`).
2. For each other open PR, checks (via the API) whether it also adds `skills/<name>/SKILL.md` with the same `<name>`.
3. On a name collision:
   - adds the `duplicate` label to the PR, and
   - writes the detail and a link to the colliding PR in the job summary.
   When the collision is resolved, the label is removed.

## Why a label (instead of a comment)

- Works on fork PRs, where a plain `pull_request` comment cannot be posted (read-only token).
- Idempotent: adding/removing a label is clean and doesn't accumulate comments.
- Filterable: `is:pr is:open label:duplicate`.
- The `duplicate` label exists by default in GitHub repos, so it likely needs no setup.

## Security

The labeler must run on `pull_request_target` to have `pull-requests: write`. This is safe **only because it never checks out or runs the PR head code**, it only reads PR metadata via the API and toggles a label. Do not add `actions/checkout` of the fork code to this workflow.

Minimal permissions: `pull-requests: write`, `contents: read`.

## Scope and constraints (decided in #313)

- **Name-collision only**, not semantic similarity. Same-topic/different-name (e.g. `llm-cost-optimization` vs `reduce-token-spend`) stays the job of the human/agent pre-flight checks; CI fuzzy matching would be too noisy.
- **Not** matching against already-merged skills, editing an existing `skills/<name>/SKILL.md` is a legitimate change, not a duplicate.
- **Non-blocking:** the label is a signal for reviewers, not a hard merge gate.

## Out of scope

- Hooks (too intrusive for an open-source contribution flow, per #313 discussion).
- Inline PR comments (possible later via a `pull_request` + `workflow_run` two-stage pattern; not needed for v1).

## Notes

Intentionally kept out of #313 to keep that PR docs-only; this is a separate `ci:` concern with its own review surface (token permissions, fork behavior).

ayobamiseun:
## Summary

Implements #315: a deterministic CI complement to the CONTRIBUTING pre-flight checks (#313). A new standalone workflow flags when a PR adds a `skills/<name>/SKILL.md` that another **open** PR also adds:

- On a name collision it adds the `duplicate` label and writes the colliding PR links + skill names to the job summary.
- When the collision clears (skill removed from the diff, or the colliding PR closes and this PR is pushed again), the label is removed.
- Non-blocking: the label is a reviewer signal, filterable via `is:pr is:open label:duplicate` — not a merge gate.

Closes #315

## Scope (as decided in #315 / #313)

- **Name-collision only** — no fuzzy/semantic matching. Same-topic/different-name proposals remain the job of the human/agent pre-flight checks.
- **Added files only** (`status: added`) — editing an existing `skills/<name>/SKILL.md` is a legitimate change and is never flagged.
- Not matched against already-merged skills.

## Security

Runs on `pull_request_target` so it can label fork PRs (a plain `pull_request` run has a read-only token there). This is safe because the job **never checks out or executes PR head code** — it only reads PR metadata via the API (`actions/github-script`) and toggles a label. A comment at the top of the workflow warns future editors not to add a checkout. Permissions are pinned to the minimum: `pull-requests: write`, `contents: read`. Per-PR `concurrency` group with `cancel-in-progress` avoids label races on rapid pushes.

## Evidence: dry-run against the live open-PR list

I ran the exact detection logic (same regex, same `status == "added"` filter) locally via `gh api` against all **89 currently open PRs**:

- 28 open PRs add at least one new `skills/<name>/SKILL.md`.
- Exactly **one true name collision** found: `skills/assistant/SKILL.md` is added by #181, #117, #62, and #61 — each of these would get the `duplicate` label with the others linked in its job summary.
- **No false positives**: same-topic-but-different-name clusters (e.g. #287 `llm-cost-optimization` vs #286 `evaluating-llm-output`, or the several accessibility proposals #261/#254) are correctly *not* flagged, matching the name-collision-only scope.

<details>
<summary>Dry-run output</summary>

```
Open PRs: 89

#335 adds: lightweight-design-analysis, software-design-principles
#331 adds: session-quality-gate
#326 adds: project-workflow
#311 adds: skill-builder-specialized-agents
#287 adds: llm-cost-optimization
#286 adds: evaluating-llm-output
#285 adds: reliable-agent-loops
#261 adds: accessibility-engineering
#254 adds: accessibility-testing-and-remediation, database-migrations-and-schema-design, llm-api-integration
#253 adds: harness-engineering
#250 adds: skill-router
#242 adds: slash-commands
#237 adds: handoff
#232 adds: agent-resilience-and-fuzzing
#224 adds: legacy-code-and-refactoring
#197 adds: workflow-tracker
#193 adds: vertical-slicing
#192 adds: ubiquitous-language
#181 adds: assistant
#171 adds: skill-design-patterns
#165 adds: browser-testing-with-real-sessions
#149 adds: ast-resilient-remediation, chaos-simulation-and-mocking, finops-mothballing-and-optimization, inter-agent-protocol-verification, semantic-paradigm-auditor
#129 adds: prompt-injection-defense
#117 adds: assistant
#115 adds: design-sprint
#62 adds: assistant
#61 adds: assistant, observability-and-monitoring
#52 adds: delegation-provenance

=== COLLISIONS (would be labeled 'duplicate') ===
skill "assistant": #181 <-> #117 <-> #62 <-> #61
```

</details>

Note: a `pull_request_target` workflow always runs the **base repo's** version, so it can't be exercised from a fork before merge — the dry-run above stands in for that. API cost is modest (one `listFiles` per open PR, only when the triggering PR actually adds a skill).

## Known limitations (by design / follow-ups)

- Event-driven: only the triggering PR is (un)labeled. The *older* colliding PR gets its label on its own next push. Likewise, if a colliding PR closes, the label on the survivor clears on its next `synchronize`. Re-evaluating all open PRs on `closed` events would fix this and can be a small follow-up if wanted.
- Relies on the default `duplicate` label existing (it does in GitHub repos by default, per the issue).

### DIFF
--- a/.github/workflows/duplicate-skill-label.yml
+++ b/.github/workflows/duplicate-skill-label.yml
@@ -1,0 +1,100 @@
+name: Duplicate Skill Labeler
+
+# Flags PRs that add a skills/<name>/SKILL.md already added by another open PR
+# (name collisions only — see issue #315 and the scope decided in #313).
+#
+# Runs on pull_request_target so it can label fork PRs. This is safe ONLY
+# because the job never checks out or executes PR head code: it reads PR
+# metadata via the API and toggles a label. Do not add a checkout of the
+# fork code to this workflow.
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+
+permissions:
+  contents: read
+  pull-requests: write
+
+concurrency:
+  group: duplicate-skill-label-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  check-collisions:
+    name: Check for skill name collisions
+    runs-on: ubuntu-latest
+    steps:
+      - name: Compare added skills against other open PRs
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const LABEL = 'duplicate';
+            const SKILL_RE = /^skills\/([^/]+)\/SKILL\.md$/;
+            const { owner, repo } = context.repo;
+            const prNumber = context.payload.pull_request.number;
+
+            const addedSkills = async (pull_number) => {
+              const files = await github.paginate(github.rest.pulls.listFiles, {
+                owner, repo, pull_number, per_page: 100,
+              });
+              return files
+                .filter((f) => f.status === 'added')
+                .map((f) => SKILL_RE.exec(f.filename)?.[1])
+                .filter(Boolean);
+            };
+
+            const ours = new Set(await addedSkills(prNumber));
+            core.info(`PR #${prNumber} adds ${ours.size} skill(s): ${[...ours].join(', ') || '(none)'}`);
+
+            const collisions = [];
+            if (ours.size > 0) {
+              const openPRs = await github.paginate(github.rest.pulls.list, {
+                owner, repo, state: 'open', per_page: 100,
+              });
+              for (const pr of openPRs) {
+                if (pr.number === prNumber) continue;
+                const overlap = (await addedSkills(pr.number)).filter((name) => ours.has(name));
+                if (overlap.length > 0) {
+                  collisions.push({ number: pr.number, title: pr.title, url: pr.html_url, skills: overlap });
+                }
+              }
+            }
+
+            const currentLabels = await github.paginate(github.rest.issues.listLabelsOnIssue, {
+              owner, repo, issue_number: prNumber, per_page: 100,
+            });
+            const hasLabel = currentLabels.some((l) => l.name === LABEL);
+
+            if (collisions.length > 0) {
+              if (!hasLabel) {
+                await github.rest.issues.addLabels({
+                  owner, repo, issue_number: prNumber, labels: [LABEL],
+                });
+              }
+              core.summary.addHeading('Duplicate skill detected', 3);
+              core.summary.addRaw(
+                `This PR adds a \`skills/<name>/SKILL.md\` that another open PR also adds. ` +
+                `Per the [contributing pre-flight checks](https://github.com/${owner}/${repo}/blob/main/CONTRIBUTING.md#before-proposing-a-new-skill), ` +
+                `coordinate with the colliding PR instead of duplicating it.\n\n`
+              );
+              core.summary.addTable([
+                [{ data: 'Colliding PR', header: true }, { data: 'Skill(s)', header: true }],
+                ...collisions.map((c) => [
+                  `<a href="${c.url}">#${c.number} — ${c.title}</a>`,
+                  c.skills.map((s) => `<code>${s}</code>`).join(', '),
+                ]),
+              ]);
+              await core.summary.write();
+              core.notice(
+                `Labeled #${prNumber} '${LABEL}': collides with ` +
+                collisions.map((c) => `#${c.number} (${c.skills.join(', ')})`).join(', ')
+              );
+            } else if (hasLabel) {
+              await github.rest.issues.removeLabel({
+                owner, repo, issue_number: prNumber, name: LABEL,
+              }).catch((e) => { if (e.status !== 404) throw e; });
+              core.notice(`No remaining skill name collision — removed '${LABEL}' label from #${prNumber}.`);
+            } else {
+              core.info('No skill name collisions with other open PRs.');
+            }
--- a/.github/workflows/duplicate-skill-label.yml
+++ b/.github/workflows/duplicate-skill-label.yml
@@ -68,9 +68,21 @@ jobs:
 
             if (collisions.length > 0) {
               if (!hasLabel) {
-                await github.rest.issues.addLabels({
-                  owner, repo, issue_number: prNumber, labels: [LABEL],
-                });
+                // Self-heal if the default 'duplicate' label was deleted:
+                // create it (ignoring already-exists races), then retry once.
+                try {
+                  await github.rest.issues.addLabels({
+                    owner, repo, issue_number: prNumber, labels: [LABEL],
+                  });
+                } catch (e) {
+                  await github.rest.issues.createLabel({
+                    owner, repo, name: LABEL, color: 'cfd3d7',
+                    description: 'This issue or pull request already exists',
+                  }).catch((err) => { if (err.status !== 422) throw err; });
+                  await github.rest.issues.addLabels({
+                    owner, repo, issue_number: prNumber, labels: [LABEL],
+                  });
+                }
               }
               core.summary.addHeading('Duplicate skill detected', 3);
               core.summary.addRaw(
@@ -78,11 +90,12 @@ jobs:
                 `Per the [contributing pre-flight checks](https://github.com/${owner}/${repo}/blob/main/CONTRIBUTING.md#before-proposing-a-new-skill), ` +
                 `coordinate with the colliding PR instead of duplicating it.\n\n`
               );
+              const esc = (s) => s.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;').replace(/"/g, '&quot;');
               core.summary.addTable([
                 [{ data: 'Colliding PR', header: true }, { data: 'Skill(s)', header: true }],
                 ...collisions.map((c) => [
-                  `<a href="${c.url}">#${c.number} — ${c.title}</a>`,
-                  c.skills.map((s) => `<code>${s}</code>`).join(', '),
+                  `<a href="${c.url}">#${c.number} — ${esc(c.title)}</a>`,
+                  c.skills.map((s) => `<code>${esc(s)}</code>`).join(', '),
                 ]),
               ]);
               await core.summary.write();


nucliweb:
Really solid PR. The `pull_request_target` handling is the part that usually goes wrong and here it is right: no checkout or execution of PR head code, minimal permissions (`contents: read`, `pull-requests: write`), no untrusted data interpolated into `${{ }}`, per-PR concurrency, and a header comment warning editors not to add a checkout. The live dry-run against the 89 open PRs (one true `assistant` collision, zero false positives) is exactly the right way to validate a workflow that can't be exercised from a fork before merge.

Two things I validated while reviewing, plus a maintainability suggestion.

**Detection scope is correct.** The `status === 'added'` filter plus the anchored `^skills/([^/]+)/SKILL\.md$` regex means editing an existing skill (`modified`) is never flagged, and matching is on the exact one-level skill name, not fuzzy or nested paths. So this only catches two *open* PRs racing to add the same brand-new skill name, not edits and not the already-merged catalog. That is the useful signal and it behaves as intended.

**Suggestion: extract the detection logic into a committed, testable script.** You already ran the exact logic locally via `gh api` for the dry-run, so formalizing that into a script both the workflow and a local test invoke would turn the manual dry-run into a repeatable test, which is where the regex/intersection bugs would surface. One caveat if you go this route: keep the injection-safe property `github-script` gives you for free. Whatever the language, untrusted PR fields (title, head ref) must arrive via `env:` and be quoted, never inlined as `${{ ... }}` in a shell body. A middle path that avoids touching the security-sensitive glue: keep `github-script` for the API/label I/O but pull the pure detection function into a small module with unit tests. Non-blocking either way.

**Minor hardening (optional, pre-merge):**
- `issues.addLabels` has no guard if the `duplicate` label is ever deleted (it exists today, verified). Unlike `removeLabel`, there is no `.catch`, so a missing label would fail the step. A `createLabel`-if-missing (catching 422) would make it self-healing.
- The colliding PR title is interpolated raw into the summary HTML link text. GitHub sanitizes step summaries so impact is cosmetic, but escaping `c.title` would be tidier.

None of this blocks; the workflow is mergeable as-is. The label-asymmetry and `closed`-event follow-ups you already documented are the right calls.

ayobamiseun:
Thanks for the careful review — both hardenings applied in 160dcad:

- **Missing-label self-heal:** `addLabels` is now wrapped in a try/catch; on failure it creates the `duplicate` label (default color `cfd3d7` + GitHub's stock description, ignoring an already-exists 422 race) and retries once. Defensive regardless of whether the API auto-creates on add.
- **Title escaping:** PR titles and skill names are now HTML-escaped (`& < > "`) before going into the job-summary table.

On extracting the detection logic into a committed, testable script: I considered it and decided to keep the workflow self-contained, at least in this PR. Loading a module (or invoking a script) requires adding a checkout step to a `pull_request_target` workflow — even a safe base-ref checkout weakens the current "no checkout at all, nothing to get wrong" property, and creates a spot where a future edit flipping the ref to the PR head becomes a vulnerability rather than a no-op. The pure logic being guarded is ~15 lines (one anchored regex + a set intersection), which felt like the wrong trade for that surface. If maintainers prefer the tested-module route, I'm happy to do it as a follow-up alongside the `closed`-event re-evaluation, where the checkout question has to be answered anyway.

addyosmani:
This is carefully done, and I appreciate the explicit note about never checking out PR head code, since that's the exact footgun with pull_request_target. It also lines up with the anti-duplication work we've been doing. The only reason I'm not merging on the spot is that any workflow with that trigger gets my own eyes on the permissions before it lands. Give me a beat to review it properly.

ayobamiseun:
Sounds good, take your time. To make the security pass quick, the full surface is: `permissions` pinned to `contents: read` + `pull-requests: write`; no checkout step of any kind; no `${{ }}` interpolation of PR-controlled data anywhere in the workflow (untrusted fields like PR titles only flow through the github-script API client, and are HTML-escaped before landing in the job summary as of 160dcad). Happy to adjust anything you flag.

addyosmani (approved):
Want this. It automates the exact duplicate-skill problem I keep catching by hand, and it's done safely: pull_request_target with no checkout of fork code, just metadata plus a label. Approving.

abhisheksharma2411:
This is still open at 7 weeks and `mergeable_state: clean`, which surprised me — #315 has a comment from @addyosmani saying "This is live now: #362 implements exactly what you described", so I'd assumed it had landed. It hasn't: `main` has only `test-plugin-install.yml` in `.github/workflows/`, so the labeler isn't running on anything yet.

Flagging that because it may just have been lost rather than deliberately held. Same shape as #444, which sat approved for three weeks with zero check runs.

## A gap worth knowing about, with me as the worked example

I read the script rather than the description, and the scope limit is explicit and deliberate:

```js
const SKILL_RE = /^skills\/([^/]+)\/SKILL\.md$/;
...
.filter((f) => f.status === 'added')
```

Added skill directories, name collisions only — exactly as the comment says and as #313 scoped it. No complaint about that.

**But the duplicate I personally caused in this repo would have been invisible to it.** #505 and #428 both modified `scripts/lib/skill-lint.js`, and #428 was a month older and already reviewed. Neither added a skill directory, so `addedSkills()` returns empty for both and the labeler correctly does nothing. CONTRIBUTING's actual instruction is broader than the check — *"Check open PRs... don't add to near-duplicate clusters"* — and the clusters that bite hardest here aren't new skills, they're several people converging on the same validator file.

From today's open PRs the same-file convergence is easy to see: #428, #444 and #498 all touch `scripts/`, and #444 and #428 both add a `skill-lint` test file at different paths.

## Offer

The extension is small and belongs in your script rather than a second workflow — a second labeler competing with this one would be its own near-duplicate cluster, which would be a poor joke. Roughly: alongside `addedSkills`, a `modifiedPaths` set, and a collision when the intersection is non-empty, reported distinctly from a skill-name collision so the label still means something specific.

Happy to push that to your branch, or to open it as a follow-up once this merges — your call, and this is your PR. I'd rather ask than start writing.

For what it's worth I'd merge this as-is first: the skill-name axis is the higher-frequency case, and the modified-file axis is strictly additive.

abhisheksharma2411 (commented):
This is well-built, and the security posture is right — the comment explaining *why* `pull_request_target` is safe here (never checks out head code) is the part people usually leave out, and it's the part that stops someone adding a `checkout` step in six months. Scoping to `status: 'added'` so edits to an existing skill are never flagged is the correct call too.

I checked the pinning before raising it: the repo's other workflows use `actions/checkout@v6` and `actions/setup-node@v6`, so `github-script@v7` matches the house convention. Not a finding.

Three things, one of which I think is worth addressing before merge.

**1. Only one PR of a colliding pair ever gets the label.** Every write in the script targets `issue_number: prNumber` — the PR that triggered the run. So if #400 opens first and #410 opens later adding the same skill, #410's run sees #400 and labels **#410**. #400 is never labeled, because its own run happened before #410 existed.

That undercuts the stated purpose. The label is meant to be a filterable reviewer signal via `is:pr is:open label:duplicate`, but that filter returns the later half of each pair only — a reviewer sees #410 and has to open it to discover #400 exists. The pair is the useful unit, and half of it is missing from the query.

Labeling both sides when a collision is found would fix it, at the cost of the workflow writing to a PR other than the one that triggered it. Given `pull-requests: write` is already held that's not a new permission, but it is a scope change worth being deliberate about.

**2. Stale labels, which you've flagged, bite harder because of #1.** The body notes the label clears only when the PR is pushed again. Combined with the above: when #410 is merged, #400 was never labeled so nothing is stale there — but when a *third* PR collides and then closes, the survivor keeps `duplicate` indefinitely with nothing to collide with. A reviewer filtering on the label then can't distinguish live collisions from resolved ones, which is the same failure the label exists to prevent.

A `schedule:` sweep that re-evaluates open labeled PRs, or handling `pull_request_target: closed` to re-run collision checks on the PRs that referenced it, would close it. Either is a follow-up, not a blocker.

**3. The API cost is O(open PRs) per event, and this repo is at 60.** Each run does one `listFiles` for the triggering PR, then `pulls.list`, then one paginated `listFiles` **per other open PR** — so ~62 calls minimum right now, more with pagination on large PRs. That's per `opened`/`synchronize`/`reopened`, so a contributor pushing five times to one PR costs ~300 calls, and `GITHUB_TOKEN` is limited to roughly 1,000 requests/hour per repository shared with every other workflow.

The early exit when `ours.size === 0` is what saves this — most PRs here don't add a skill, so they cost 1 call and stop. Worth stating that explicitly in the header comment, because it's the load-bearing property and it isn't obvious. If it ever becomes a problem, filtering the open-PR list by `pr.changed_files` or by a `skill` label before fanning out would cut it further.

None of these are correctness bugs in what's written — the script does what it says. #1 is the one I'd want resolved, since it's the difference between the label being a reliable filter and being a hint.

PR #363, opened 2026-07-08
fix: enable slash commands for GitHub Copilot CLI

pedrormonteiro:
## Problem

After installing `agent-skills` as a GitHub Copilot CLI plugin, the 8 lifecycle slash commands (`/spec`, `/plan`, `/build`, etc.) listed in the README do not work. Users see no commands registered when installing `agent-skills` using Copilot's marketplace.

## Root Cause

The `plugin.json` file was missing a `"commands"` field. Without it, GitHub Copilot CLI never discovers the command files. The `.claude/commands/*.md` files already existed with the correct content and format — they just weren't being pointed to.The commands were only wired up for Claude Code (via `.claude/commands/` auto-discovery) and Antigravity CLI (via `commands/*.toml` filesystem convention). Copilot CLI requires an explicit `"commands"` field in `plugin.json`.

## Fix

Add `"commands": ".claude/commands"` to `plugin.json`. This tells Copilot CLI to load the existing `.claude/commands/*.md` files as slash commands.

## Testing

Tested locally by modifying the installed plugin at `~/.copilot/installed-plugins/` and reloading the session. All 8 commands resolved correctly.

### DIFF
--- a/docs/copilot-setup.md
+++ b/docs/copilot-setup.md
@@ -2,6 +2,30 @@
 
 ## Setup
 
+### Slash Commands (GitHub Copilot CLI)
+
+After installing the plugin, all 8 lifecycle slash commands are available natively in the GitHub Copilot CLI:
+
+| Command | What it does |
+| --- | --- |
+| `/spec` | Write a structured spec before coding |
+| `/plan` | Break work into small, ordered tasks |
+| `/build` | Implement the next task incrementally (add `auto` for full plan in one pass) |
+| `/test` | TDD workflow — red, green, refactor |
+| `/review` | Five-axis code review before merge |
+| `/webperf` | Web performance audit |
+| `/code-simplify` | Simplify code for clarity |
+| `/ship` | Pre-launch checklist and go/no-go |
+
+Install the plugin if you haven't already:
+
+```bash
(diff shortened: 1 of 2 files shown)


pedrormonteiro:
@addyosmani (sorry for tagging you directly), I know this change might not be useful for everyone, but as someone who uses agent-skills and references it on a weekly-basis to my colleagues, who have to use Copilot, I believe this would be a good addition. Basically, when adding agent-skills through the skill marketplace, the slash commands do not work, but this PR fixes that issue and ensures we have a single/easier way to install.

addyosmani:
The docs half is good. The part I need to sit with is adding a commands key to the root plugin.json, since that file is the Antigravity manifest and I want to confirm Copilot actually reads commands from there and that we aren't perturbing Antigravity in the process. Can you point me at where the Copilot CLI picks up commands? If it's the same manifest, great, I just want to verify before it ships.

addyosmani:
Thanks. I'm going to close this one. The Copilot slash-command docs need to match how the commands actually register (see #421), and I'd rather start fresh than patch this version. A new focused PR to add or recover that documentation is welcome.

Issue #329, opened 2026-06-26
Clarify why shared references are not colocated under each skill directory


noahacgn:
Hi, thanks for publishing this skill pack — it is a very useful reference implementation.

I have a question about the current repository layout.

From the Agent Skills specification, a skill appears to be intended as a self-contained directory:

```text
skill-name/
├── SKILL.md
├── scripts/
├── references/
├── assets/
└── ...
```

The spec also says that file references inside `SKILL.md` should use paths relative to the skill root, for example:

```markdown
See references/REFERENCE.md for details.
```

In this repository, however, the reference documents are currently stored in a top-level shared directory:

```text
agent-skills/
├── skills/
│   ├── test-driven-development/
│   │   └── SKILL.md
│   └── ...
├── references/
│   ├── definition-of-done.md
│   ├── testing-patterns.md
│   ├── security-checklist.md
│   └── ...
└── ...
```

Some `SKILL.md` files refer to files such as `references/security-checklist.md` or `references/definition-of-done.md`, but those files are not located under the individual skill directory.

My concern is portability: if a client or user installs/copies only an individual skill directory, or installs only the `skills/` path, the referenced files may not be available. This seems especially relevant because the README includes examples such as installing with `--path skills`.

Is the top-level `references/` directory an intentional pack-level design choice? If so, would it make sense to document that these references are shared pack-level resources and that users/clients should install the full repository rather than individual skill directories?

Alternatively, should the references that a skill depends on be colocated under that skill’s own `references/` directory, to better match the Agent Skills spec and keep each skill self-contained?

Possible options:

1. Move or copy reference files under the relevant `skills/<skill-name>/references/` directories.
2. Keep the shared top-level `references/` directory, but document that this repository is a skill pack with shared resources and may not be fully portable when individual skills are copied.
3. Update the installation guidance so clients that need references install the full repository/package, not just `skills/`.

I may be missing some project-specific context, so I wanted to ask before assuming this is a bug. Thanks!

addyosmani:
Fair question, and the short answer is deliberate de-duplication. Several skills point at the same checklists (security, performance, accessibility, definition-of-done), so colocating would mean either copying them into every skill that uses them or picking one skill to "own" each, and both drift over time. Keeping them in references/ gives one source of truth. The tradeoff is the per-skill npx install gap in #361, which we're working. Good enough question that I may fold this rationale straight into the docs.

kevglynn:
[...]

PR #358, opened 2026-07-07
fix(evals): cover description vocabulary gaps

ZhiyaoWen999:
## Summary

- Enrich `frontend-ui-engineering` and `performance-optimization` descriptions with the user vocabulary called out in #351.
- Restore stricter trigger coverage for responsive/WCAG page work and N+1/database query optimization.
- Add the deferred `frontend-ui-engineering` owner to the `api-and-interface-design` responsive landing page negative.

## Why

The deterministic trigger evals surfaced that fair user prompts could miss these skills because the frontmatter descriptions did not include vocabulary users actually say. This keeps the descriptions in the repo's what-then-Use-when format while making those routing cues explicit.

Fixes #351.

## Validation

- `node scripts/validate-skills.js`
- `node scripts/run-evals.js` -> 124 checks passed, 0 errors, 0 warnings; trigger rank-1 rate 86% (65/76)

PR #366, opened 2026-07-08
fix(skills): point shared references at GitHub URLs for npx installs (#361)

### LINKED ISSUE #361: npx packs only skills/

evanclan:
## Summary
Per-skill `npx skills add … --skill <name>` installs only that skill directory, so relative links like `` `references/definition-of-done.md` `` resolve to nothing. This switches the 17 shared-checklist mentions across 11 skills to canonical GitHub blob URLs, and documents the convention so future skills stay portable under both whole-repo and per-skill installs.

Fixes #361. Implements the approach suggested by @nucliweb on that issue (and addresses the portability concern raised in #329 without duplicating checklists into each skill).

## Problem
`planning-and-task-breakdown` (and 10 other skills) point at repo-root checklists via a bare relative path:

```markdown
See `references/definition-of-done.md`.
```

That path only exists when the full repo is present (Claude Code marketplace plugin, whole-repo clone). Under `npx skills add addyosmani/agent-skills --skill planning-and-task-breakdown`, the sibling `references/` folder is never copied, so the link is dead. The skill still runs — these are See Also deep-dives — but the checklist is unreachable.

## Solution
- Replace every `` `references/<file>.md` `` mention in `skills/*/SKILL.md` with a markdown link to `https://github.com/addyosmani/agent-skills/blob/main/references/<file>.md` (11 skills, 17 links).
- Note the per-skill install gap in the README Quick Start next to the `--skill` examples.
- Codify the URL convention in `CONTRIBUTING.md` and `docs/skill-anatomy.md` so new skills don't reintroduce relative paths.

Whole-repo installs keep working: the URLs resolve either way, and the checklists remain unduplicated at the repo root.

## Out of scope
- Did not colocate or copy `references/` under each skill (would fight the no-duplication rule in CONTRIBUTING).
- Did not change the third-party `npx skills` installer.
- Did not touch open PRs that already cover adjacent work (#201 local-reference audit, #358 description vocabulary, draft #228 SessionStart quoting).

## Test plan
- [x] `node scripts/validate-skills.js` — 24 skills, 0 errors, 0 warnings
- [x] `bash hooks/session-start-test.sh` — `session-start JSON payload OK` (meta-skill content changed)
- [x] `rg '\`references/' skills/ --glob '**/SKILL.md'` — no remaining relative checklist refs
- [ ] CI to confirm the rest

## Related
- #361 — original report (npx packs only `skills/`)
- #329 — layout/portability question that motivated documenting the pack-level `references/` design
- Comment on #361 by @nucliweb — proposed the URL approach this PR implements

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,6 +57,7 @@ The frontmatter fields above are required. The section anatomy is a recommended 
 - Don't create supporting files unless content exceeds 100 lines
 - Don't create an empty `scripts/` directory just to match another skill — add `scripts/` only when the skill includes runnable helpers
 - Don't put reference material inside skill directories — use `references/` instead
+- When a skill links to a shared checklist in `references/`, use the canonical GitHub URL (`https://github.com/addyosmani/agent-skills/blob/main/references/<file>.md`), not a bare relative path — per-skill `npx` installs do not include the repo-root `references/` folder ([#361](https://github.com/addyosmani/agent-skills/issues/361))
 
 ## Modifying Existing Skills
 
--- a/README.md
+++ b/README.md
@@ -57,6 +57,8 @@ npx skills add addyosmani/agent-skills --skill interview-me              # requi
 npx skills add addyosmani/agent-skills --skill test-driven-development   # red-green-refactor, enforced
 ```
 
+> **Per-skill installs and shared checklists:** `npx skills add … --skill <name>` copies only that skill directory. Shared checklists live in the repo-root [`references/`](references/) folder and are not included. Skills link to those checklists via their GitHub URLs so the links still resolve; for offline/local copies of the checklists, install the full pack or clone the repo.
+
 Prefer a native integration? Pick your tool below.
 
 <details>
--- a/docs/skill-anatomy.md
+++ b/docs/skill-anatomy.md
@@ -143,6 +143,7 @@ When a skill ships runnable helpers under `scripts/`, each script follows these 
 - Skill files: `SKILL.md` (always uppercase)
 - Supporting files: `lowercase-hyphen-separated.md`
 - References: stored in `references/` at the project root, not inside skill directories
+- Links from a skill to a shared checklist should use the canonical GitHub URL (`https://github.com/addyosmani/agent-skills/blob/main/references/<file>.md`), not a bare relative path — per-skill installs do not include the repo-root `references/` folder
 
 ## Cross-Skill References
 
--- a/skills/code-review-and-quality/SKILL.md
+++ b/skills/code-review-and-quality/SKILL.md
@@ -338,8 +338,8 @@ Part of code review is dependency review:
 ```
 ## See Also
 
-- For detailed security review guidance, see `references/security-checklist.md`
-- For performance review checks, see `references/performance-checklist.md`
+- For detailed security review guidance, see [references/security-checklist.md](https://github.com/addyosmani/agent-skills/blob/main/references/security-checklist.md)
+- For performance review checks, see [references/performance-checklist.md](https://github.com/addyosmani/agent-skills/blob/main/references/performance-checklist.md)
 
 ## Common Rationalizations
 
--- a/skills/doubt-driven-development/SKILL.md
+++ b/skills/doubt-driven-development/SKILL.md
@@ -43,7 +43,7 @@ If you doubt every keystroke, you ship nothing. The skill applies only to non-tr
 
 This skill is designed for the **main-session orchestrator**, where Step 3 (DOUBT, detailed below) can spawn a fresh-context reviewer.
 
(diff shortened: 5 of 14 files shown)


nucliweb (commented):
Thanks for picking this up, @evanclan, and for implementing the URL approach cleanly. I verified the conversion against the base: 17 bare relative references in `skills/` on `main`, 17 canonical URLs on this branch, zero relative checklist refs left in `skills/`. The README / CONTRIBUTING / skill-anatomy notes are a good addition (documenting the convention in skill-anatomy too is more than I suggested on #361). Nice work.

### Main thing before merge: the conversion is incomplete outside `skills/`

Your own rationale (per-skill installs do not include the repo-root `references/` folder) applies to other extractable artifacts too, and four bare relative references were left untouched:

- `agents/web-performance-auditor.md:175` -> `references/performance-checklist.md`
- `.claude/commands/ship.md:24` -> `references/orchestration-patterns.md`
- `.gemini/commands/ship.toml:23` -> `references/orchestration-patterns.md`
- `commands/ship.toml:23` -> `references/orchestration-patterns.md`

The audit in the test plan is scoped to `skills/ --glob **/SKILL.md`, which is exactly why these slipped through. Agents and commands can also be used without the repo-root `references/` present, so their links are equally dead there, and the repo is now inconsistent (skills use URLs, agents/commands use relative paths). Either convert these four for completeness and consistency, or make a conscious call that agents/commands are whole-repo-only and say so. I would lean toward converting them so the new convention holds everywhere.

### Coordination note with #201

After this PR, skills have no local `references/` links left. #201 (warn on missing local skill references) would need to account for that: a validator that checks local reference files would either have nothing to check or, worse, flag all of these URL links as unresolved. Worth a heads-up on #201 so whoever implements it validates URL liveness (or is at least aware references are now URLs) rather than colliding with this change.

### Minor

- #329 asks specifically to clarify *why* references are not colocated under each skill. This PR documents the URL convention (the mechanism) but not fully the no-duplication rationale. If you intend to close #329, a sentence on that reasoning would round it out.
- The "Out of scope" section calls #201/#358/#228 open PRs; they are issues. Cosmetic.
- The links use `blob/main`, which renders HTML. If the goal is for an agent to fetch the checklist content, `raw.githubusercontent.com/.../main/references/<file>.md` returns raw markdown. For human clicks `blob` is fine (it is what I suggested on #361), so this is optional.

Net: faithful, well-documented implementation, mergeable in essence. The agents/commands completeness gap is the one I would close before merge so the convention is consistent across the repo.

addyosmani:
The underlying problem is real: per-skill installs don't carry the repo-root references, so relative links break. My hesitation is that the fix hardcodes blob/main URLs across a dozen skills, which trades a broken-link problem for a rename-fragility one and leans on an external CLI convention I've been careful about mentioning. I'd rather solve this without that coupling if there's a lighter touch. Let me think on it before we bake absolute URLs into every skill.

nucliweb:
Heads up @evanclan on how this now sits against `main`. This PR predates two changes that landed in the same region of `skill-anatomy.md`, so it will need a rebase, but the good news is there is no real conflict of intent, just repositioning.

What changed under it:
- **#401 (merged)** added a `## Shared References` section explaining *why* shared checklists live in the repo-root `references/` and tracking the portability gap as #361, plus a `(see [Shared References](#shared-references) for why)` anchor on the References bullet.
- **#236 (open, mergeable, likely landing first)** rewrites that single References bullet into two: a *Shared references* bullet and a *Skill-specific references* bullet documenting the emerging skill-local `references/` convention.

The key point: your change and #236 are complementary halves of the same #361 fix, not competing answers. #236 covers material used by a single skill (colocate it inside the skill so a per-skill install carries it). Yours covers material shared across skills that cannot be colocated (link it by canonical GitHub URL so the link still resolves). Together they close the gap for both cases.

Concretely, once you rebase onto current `main`: the bullet you anchor after (`- References: stored in...`) will have become `- Shared references:...`, so your canonical-URL bullet slots naturally right after it, and it would read best tied into the new Shared References section rather than standing alone. Happy to help review once it is refreshed.

addyosmani (commented):
Good intent, and dead links on per-skill installs are a real problem (#361). My hesitation is that hardcoding `blob/main` URLs pins the links to main rather than the installed version, and breaks if the repo ever moves. There's also the skill-local references/ direction in play. I'd rather settle which #361 fix we're committing to before this lands, so we don't ship two half-approaches. Holding for that.

addyosmani:
Closing in favor of #469. We talked it through and landed on fixing the relative paths properly (`../../references/`) plus a CI validator, rather than rewriting to `blob/main` GitHub URLs, mainly because the URL approach pins every link to main and needs a network round-trip. Genuinely appreciate you surfacing the #361 problem and pushing on it, that's what got us to the right fix.

PR #372, opened 2026-07-09
fix: harden eval and command validation

ShiroKSH:
﻿## Summary

- make deterministic eval skill loading tolerate CRLF frontmatter
- reject behavioral eval fixture paths that escape the fixtures or temp workspace roots
- fail command parity validation when command descriptions are missing or malformed
- fix Copilot setup directory creation and stale reference docs

## Why

On Windows checkouts with CRLF line endings, the eval runner skipped every skill and reported all eval cases as missing. The command validator also treated parse failures as if they were already covered by the missing-file check. Behavioral fixture paths now fail early if an eval case tries to escape the intended roots.

## Validation

- node scripts\validate-skills.js
- node scripts\validate-commands.js
- node scripts\run-evals.js
- node scripts\run-evals.js --behavioral test-driven-development --dry-run
- node --check scripts\run-evals.js
- node --check scripts\validate-commands.js
- git diff --check

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/README.md
+++ b/README.md
@@ -339,7 +339,7 @@ agent-skills/
 │   ├── shipping-and-launch/           #   Ship
 │   └── using-agent-skills/            #   Meta: how to use this pack
 ├── agents/                            # 4 specialist personas
-├── references/                        # 5 supplementary checklists
+├── references/                        # 7 supplementary checklists
 ├── hooks/                             # Session lifecycle hooks
 ├── .claude/commands/                  # 8 slash commands (Claude Code)
 ├── .gemini/commands/                  # 8 slash commands (Gemini CLI)
--- a/docs/copilot-setup.md
+++ b/docs/copilot-setup.md
@@ -7,7 +7,7 @@
 Copilot supports creating agent skills using a `.github/skills`, `.claude/skills`, or `.agents/skills` directory in your repository.
 
 ```bash
-mkdir -p .github
+mkdir -p .github/skills/test-driven-development .github/skills/code-review-and-quality
 
 # Create files for essential skills
 cat /path/to/agent-skills/skills/test-driven-development/SKILL.md > .github/skills/test-driven-development/SKILL.md
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -132,6 +132,9 @@ The `references/` directory contains supplementary checklists:
 | `performance-checklist.md` | performance-optimization |
 | `security-checklist.md` | security-and-hardening |
 | `accessibility-checklist.md` | frontend-ui-engineering |
+| `definition-of-done.md` | all skills / every change |
+| `observability-checklist.md` | observability-and-instrumentation |
+| `orchestration-patterns.md` | context-engineering |
 
 Load a reference when you need detailed patterns beyond what the skill covers.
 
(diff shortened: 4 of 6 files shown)


nucliweb (commented):
Thanks for the contribution, @ShiroKSH, and for the thorough validation list in the description. Went through all four changes against the current code; they are individually correct and the two hardenings are genuinely useful. Notes below, plus one packaging ask.

### What checks out

- **CRLF frontmatter regex (`run-evals.js`).** Real bug: on CRLF checkouts `---\r\n` never matched `^---\n`, so `loadSkills()` skipped every skill and reported all evals missing. Worth calling out that `validate-skills.js` already uses this exact CRLF-tolerant pattern (`^---[ \t]*\r?\n...`), so this aligns `run-evals.js` with the pattern the repo already follows rather than inventing one. The stray `\r` the capture could carry is cleaned by the later `.trim()`, so the fix is complete.
- **`resolveFixturePath` traversal guard.** Correct, and applied symmetrically on source and dest. One residual worth a line: `path.resolve` is lexical and does not resolve symlinks, so a fixture that is a symlink pointing outside the root would still pass. Low risk since fixtures are in-repo and reviewed, just noting it.
- **`validate-commands.js` malformed-description check.** Closes a real gap. The old `if (descX == null) continue` conflated "file missing" (caught by the parity check) with "file present but description malformed" (silently passed). Distinguishing them with `hasOwnProperty(...) && commands[stem] == null` is the right fix, and the defensive `Object.prototype.hasOwnProperty.call` is good.
- **Docs.** The `copilot-setup.md` mkdir fix is a real bug (the following `cat > .github/skills/<skill>/SKILL.md` would fail without the subdirs). The `5 -> 7` references count and the three added `getting-started.md` rows are correct stale-doc corrections.

### Packaging ask

This bundles four independent concerns across three subsystems (eval CRLF, fixture traversal, command validation, docs). CLAUDE.md asks for small focused PRs, especially in shared `scripts/` files. Ideally this is 2-3 PRs (evals hardening / command-validation fix / docs), so a revert of one does not drag the others. Not blocking given it is 36 lines and all correct, but at minimum splitting the docs out from the script changes would help.

### Systemic guardrail: add `.gitattributes`

The CRLF issue is worth preventing at the root, not just per-parser. There is no `.gitattributes` in the repo today, so a Windows checkout with `autocrlf=true` reintroduces this class of bug anywhere frontmatter is parsed by a stricter matcher. A `.gitattributes` normalizing line endings (e.g. `* text=auto eol=lf`, or scoped to `*.md`/`*.json`) would guard the whole repo. The regex fix here is still the right immediate change and matches the existing precedent; the `.gitattributes` is a complementary guardrail on top.

### One process note

CI has not run on this branch (no checks reported), likely because fork Actions need maintainer approval. Worth getting the deterministic evals + validators to run green here before merge, since nothing has validated it in-repo yet.

Net: mergeable, four correct fixes. The split and the `.gitattributes` are the two things I would do before or alongside merge.

ShiroKSH:
﻿Thanks for the review. I added the repository-level line-ending guardrail in eb5acf1:

- added .gitattributes with text normalization to LF
- reran validate-skills, validate-commands, and run-evals successfully

I left the existing fixes in this PR rather than force-pushing a history split, since the current scope is small and the split was noted as non-blocking. Happy to split docs into a separate PR if that is preferred.

addyosmani (approved):
Appreciate the defensive pass. The line-ending handling and the fixture path guard both only bite once you have a Windows contributor or an untrusted path, so closing them before that happens is the right instinct, and normalizing to LF is overdue hygiene anyway. Thanks for tightening the edges here.

PR #373, opened 2026-07-09
Add grounding-and-abstention skill

enached134-ctrl:
## What

Adds a `grounding-and-abstention` skill (Build phase): the design discipline for LLM/RAG features that **ground every claim in retrieved evidence and abstain** (refuse, say "I don't know", or escalate to a human) instead of hallucinating a confident wrong answer.

## Why / the gap

Confident hallucination is the #1 production LLM failure, and it is a **design** problem, not only an eval one. The pack covers deterministic engineering well, and `context-engineering` covers feeding context *in* - but nothing covers what a feature should do when that context is **insufficient**: ground and refuse rather than invent.

Per CONTRIBUTING, I checked the catalog and open PRs. This is deliberately **complementary to, not a duplicate of, #286 (`evaluating-llm-output`)**:

- **#286 *measures* output quality and guards regressions.** This skill is the **build-time discipline** for the grounding + abstention behavior it measures, and it points at an eval skill for the harness rather than re-implementing one.
- Distinct from **`source-driven-development`** (grounds *your code* in framework docs) and **`doubt-driven-development`** (verifies *your reasoning*): this grounds **the product's runtime answers** in retrieved data.

It references the eval skill by name only, so it carries **no hard dependency on the unmerged #286**.

## What's in it

- `skills/grounding-and-abstention/SKILL.md` - standard anatomy (Overview, When to Use, a 6-step design process, grounding + abstention technique tables, the over-refusal <-> hallucination tradeoff, Common Rationalizations, Red Flags, Interaction with Other Skills, Verification).
- `evals/cases/grounding-and-abstention.json` - 3 positive triggers, 3 negative (owned by `test-driven-development` / `source-driven-development` / `evaluating-llm-output`), 1 behavioral eval.
- README Build-section table row + directory-tree entry.

## Conventions

Frontmatter is `name` + `description` (what, then Use-when). Standard section anatomy, under 130 lines, references other skills by name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,7 @@ The commands above are entry points. The pack includes 24 skills total — 23 li
 | [test-driven-development](skills/test-driven-development/SKILL.md) | Red-Green-Refactor, test pyramid (80/15/5), test sizes, DAMP over DRY, Beyonce Rule, browser testing | Implementing logic, fixing bugs, or changing behavior |
 | [context-engineering](skills/context-engineering/SKILL.md) | Feed agents the right information at the right time - rules files, context packing, MCP integrations | Starting a session, switching tasks, or when output quality drops |
 | [source-driven-development](skills/source-driven-development/SKILL.md) | Ground every framework decision in official documentation - verify, cite sources, flag what's unverified | You want authoritative, source-cited code for any framework or library |
+| [grounding-and-abstention](skills/grounding-and-abstention/SKILL.md) | Ground LLM/RAG answers in retrieved evidence and abstain - refuse, say "I don't know", or escalate to a human - instead of hallucinating; test the unanswerable cases, not just the hits | Building any feature where an LLM answers from documents, data, or tools and a confident wrong answer is costly |
 | [doubt-driven-development](skills/doubt-driven-development/SKILL.md) | Adversarial fresh-context review of every non-trivial decision in-flight - CLAIM → EXTRACT → DOUBT → RECONCILE → STOP, with optional user-authorized cross-model escalation | Stakes are high (production, security, irreversible), working in unfamiliar code, or a confident output is cheaper to verify now than to debug later |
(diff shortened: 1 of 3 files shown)


addyosmani:
Genuinely useful topic, and grounding-plus-abstention is a discipline RAG features get wrong constantly. The scope question is where it sits: there's an open cluster of LLM-runtime skill proposals (cost, eval, agent loops) and I haven't decided whether this pack expands into LLM-app territory or stays lifecycle-focused. Nice that you shipped it with an eval case though, that's the bar now. Holding on the scope call, not the quality.

enached134-ctrl:
Thanks, that's fair. Better to settle the scope before any of these land.

One thing that might help: I think this is one decision, not four PRs. #286, #285, #287 and this one all fall on the same side of the line. They're about the LLM app you're building, not about how the agent builds software. So whichever way you call it, the whole cluster resolves at once.

If you do expand, you could give them their own group instead of folding them into Build. A separate section, framed as "this is for the feature you're shipping, not for how you ship it". Then the existing phases stay clean and the new territory is opt-in.

If it stays lifecycle-only, happy to close this one. I'd rather the pack stay tight than get my skill in.

And good to know the eval case is the bar now. That's the right call.

addyosmani:
Good framing, and you're right that this and #285/#286/#287 are one decision rather than four: they're all about the LLM app you're shipping, not how the agent ships software. The "give them their own group instead of folding into Build" idea is the right instinct if the answer turns out to be yes, since it keeps the existing phases clean. That whole call is deliberately on hold, and I'd rather make it once for the cluster than piecemeal. Not a no, a not-yet-decided, and your comment is a useful input to it.

addyosmani:
Appreciate it, but grounding and abstention is LLM-application behavior, not general engineering practice, so it's out of scope here. Closing.

PR #375, opened 2026-07-09
docs: decline documentation and skill translations

nucliweb:
## Why

Maintainers have decided not to accept translations of the documentation or skills. Translated copies drift out of sync as skills and docs evolve, and there is no sustainable way to maintain them long-term (it would mean relying on agent translations plus community corrections, for limited value).

## What

Add a **Translations** section to CONTRIBUTING.md stating the policy, keeping all skills, docs, and contributions in English.

Docs only.


addyosmani (approved):
LGTM.

We discussed this and believe that we cannot support translations at this time due to inability to keep those docs updated and in sync faithfully. While we could use agents to help with this, its just as feasible for a user to point their agent at the repo and do a translation pass as needed. Merging.

PR #380, opened 2026-07-10
docs: separate repo orchestration rules from Claude Code platform limits

### LINKED ISSUE #377: Docs: distinguish repo orchestration convention from current Claude Code behavior
## Context

The Claude Code orchestration notes in this repo appear to predate the June 2026 Claude Code updates, so this looks like documentation drift rather than incorrect original guidance.

The repo currently describes `subagents cannot spawn other subagents` as a hard Claude Code platform constraint, while also using that constraint to reinforce the repo rule that personas should not invoke other personas.

That repo rule still seems valuable and should remain. The part that may need updating is the platform-mapping wording.

## Current drift candidates

- `AGENTS.md`, `docs/agents.md`, and `references/orchestration-patterns.md` describe nested subagents as impossible on Claude Code.
- The Agent Teams worked example still includes a separate `Clean up the team` instruction, which reads like older `TeamDelete`-style behavior.

## Current Claude Code behavior

- As of [Claude Code v2.1.172](https://code.claude.com/docs/en/changelog#2-1-172), subagents can spawn their own subagents, up to 5 levels deep.
- The official docs still allow projects to prevent that by omitting the `Agent` tool from an explicit tools allowlist or by using `disallowedTools`.
- As of Claude Code v2.1.178, Agent Teams no longer use `TeamCreate` / `TeamDelete`; with `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS=1`, each session has one implicit team and teammates are spawned directly with the Agent tool `name` parameter.
- Agent Teams still cannot nest: teammates cannot spawn their own teammates.

## Suggested direction

Keep the repo orchestration convention:

- personas do not invoke other personas;
- composition remains the job of slash commands or the user;
- Agent Teams remain the only endorsed pattern when teammates need to communicate directly.

But reword the Claude Code compatibility notes so the repo rule is framed as a deliberate project convention, not as a current hard platform limitation.

For the Agent Teams example, remove or soften the separate `Clean up the team` instruction; current docs describe no separate `TeamDelete`-style cleanup step, while still documenting graceful teammate shutdown and possible orphaned tmux cleanup.

## Non-goal

This issue is not asking the repo to adopt nested persona trees. It is only about keeping the Claude Code interop documentation accurate while preserving the repo orchestration model.

## Sources

- Claude Code subagents docs: https://code.claude.com/docs/en/sub-agents
- Claude Code Agent Teams docs: https://code.claude.com/docs/en/agent-teams
- Claude Code changelog: https://code.claude.com/docs/en/changelog

Pitchfork-and-Torch:
﻿## Summary

Fixes #377 (collaborator-confirmed docs drift).

Separates **repo orchestration policy** (personas do not invoke personas) from **current Claude Code platform behavior**, which agents following `source-driven-development` would otherwise treat as contradictory.

## What changed

| File | Change |
|------|--------|
| `references/orchestration-patterns.md` | Rewrite "Platform-enforced rules" → convention vs platform; update Agent Teams cleanup (no TeamCreate/TeamDelete); soft spawn prompt |
| `docs/agents.md` | Rule 2 + interop section no longer claim nested subagents are impossible |
| `AGENTS.md` | Claude Code interop paragraph aligned with v2.1.172 / v2.1.178 |
| `.claude/commands/ship.md` | Fan-out constraints rephrased as repo rule + platform mapping |
| `skills/doubt-driven-development/SKILL.md` | Nested-context guidance no longer assumes platform hard-block |

## Verified against official docs

- Nested subagents allowed as of Claude Code v2.1.172 (depth 5): https://code.claude.com/docs/en/sub-agents#spawn-nested-subagents
- Agent Teams: no TeamCreate/TeamDelete as of v2.1.178; no nested teams; session-end cleanup: https://code.claude.com/docs/en/agent-teams
- Repo still endorses: personas do not invoke personas; composition via slash commands / user; Agent Teams for multi-persona messaging

## Non-goals

- Does **not** adopt nested persona trees
- Does **not** change skill workflows beyond accuracy of Claude Code mapping
- Does **not** touch hooks / plugin discovery paths

## Test plan

- [x] `node scripts/validate-skills.js` — 24 skills, 0 errors, 0 warnings
- [x] Grep: no remaining "subagents cannot spawn other subagents" as a hard platform claim
- [ ] Maintainer skim: orchestration-patterns section still matches pack intent


addyosmani:
This is the right cut to make. The persona-never-invokes-persona rule is ours and permanent, but the Claude Code platform behavior around it drifts, and conflating the two is how the docs went stale in the first place. Splitting them so the policy stands on its own and the platform note can age independently is exactly the fix. Reads clean, I'll give it a proper look for merge.

federicobartoli (changes requested):
The split itself looks right to me, and it matches what Addy said about separating our convention from what the platform enforces. One concern before this lands though.

The diff states the Claude Code version eight times across five files ("as of v2.1.172", "v2.1.178"), including skills/doubt-driven-development/SKILL.md and the ship command, which agents load as instructions. Version-pinned platform facts rot silently: the next CLI release makes the text stale and nothing flags it. Anthropic's skill authoring guide calls this out under avoiding time-sensitive information.

Could we drop the pins entirely? For agent-facing text the version adds nothing actionable: "current Claude Code releases allow nested subagent spawn; this pack forbids persona chains regardless" stays true either way. And where the detail matters for a human reader, linking the official Claude Code docs beats any dated statement here, since that page is the only thing that stays current. The PR already does exactly this for Agent Teams limitations, so it would just be applying the same pattern to the rest.

Pitchfork-and-Torch:
@federicobartoli Good catch on the version pins — agreed they rot silently in agent-loaded docs.

Dropped the Claude Code patch-version pins across the five files and kept the repo-convention vs platform-capability split without timestamps. PTAL.

Pitchfork-and-Torch:
@federicobartoli @addyosmani Rebased onto current main.

Version pins are fully dropped (commit on this branch after your review). Agent-facing text now states the convention vs platform capability without dated Claude Code patch versions, matching Anthropic skill-authoring guidance.

Ready for re-review when you have a moment. Happy to tweak wording if anything still reads time-sensitive.

Pitchfork-and-Torch:
### Rebased onto current `main`

Rebased `docs/claude-code-orchestration-interop-377` onto latest `addyosmani/agent-skills` main (resolved conflict in `skills/doubt-driven-development/SKILL.md`).

- Kept the relative path fix to `../../references/orchestration-patterns.md` from upstream
- Preserved the repo-convention vs Claude Code platform split
- Version pins remain dropped (review feedback)
- No remaining `v2.x.y` pins in the PR files

Ready for re-review when you have a moment.

Pitchfork-and-Torch:
Friendly ping — still ready for re-review after the Aug 10 rebase (conflict resolved; version pins remain dropped). Happy to adjust if anything still feels time-sensitive.

Pitchfork-and-Torch:
@federicobartoli The version pins you flagged are gone on the current tip (dropped after your review; later rebases kept them out). Agent-facing text is now convention vs platform capability, with no dated Claude Code patch versions. Ready for re-review when you have a moment.

Issue #180, opened 2026-05-18
Support for multiple/parallel specs


afshinm:
Thanks for building this!

Currently, the `spec.md` file is global and is written to the root folder of the project (https://github.com/addyosmani/agent-skills/blob/main/.claude/commands/spec.md), which is probably ok if you are working on a single task/spec at a time, but I was wondering if the flow can be improved so that specs/tasks/todos are scoped to a particular `/spec` command as opposed to a global `spec.md` file?

I believe OpenSpec, Spec Kit, etc. have this level of scoping already which is useful.

Thanks

addyosmani:
Real need, and it's partly getting addressed: #383 adds spec-evolution guidance (V1 scoping plus how V2 work branches off), which is the adjacent half of this. True parallel specs, several live at once, is the piece still open, and the honest constraint is where they live on disk without stepping on each other. Keeping this open as the tracking issue for the parallel case. Good one to have documented.

PR #383, opened 2026-07-10
docs(spec): add spec evolution guidance — V1 scoping and delta specs

### LINKED ISSUE #150: Recommended way to handle V2 features without polluting V1 spec?

ayobamiseun:
## Summary

Answers two long-open user questions about what happens to a spec *after* V1:

- **#150** — should the spec mark itself as a V1 and mention the V2 backlog, and what's the workflow for V2 features (e.g. SQLite → PostgreSQL, manual step → OCR) after `/build` and `/ship`?
- **#180** (workflow half) — `/spec` writes a single global `SPEC.md`, which gets overwritten when you spec something else.

Closes #150

## What changed

**`skills/spec-driven-development/SKILL.md`**
- The spec template gains an **Out of Scope (Deferred)** section — V2 ideas are captured in the spec without expanding V1, and the V1 boundary becomes enforceable during implementation (answering #150's first question directly).
- New **Spec Evolution: V1 Scoping and Follow-Up Changes** section: once a spec has shipped, a change gets its own small **delta spec** (objective / current behavior / new behavior / success criteria / boundary deltas), saved alongside the original (`SPEC-postgres-migration.md` next to `SPEC.md`), run through the same gated workflow, and folded back into the original after shipping — the original spec always describes the current system, delta specs describe transitions (answering #150's second question).
- Two new rationalization rows, two red flags, and two verification items covering mid-build scope creep and "the feature already has a spec" changes.

**`.claude/commands/spec.md` + `.gemini/commands/spec.toml`** (kept in parity; `validate-commands` passes)
- One guard line: if `SPEC.md` exists and the request is a change to what it describes, write a delta spec instead of overwriting. This fixes the overwrite pain in #180 at the workflow level.

**`evals/cases/spec-driven-development.json`**
- Frontmatter description gains the evolution vocabulary ("shipped spec", "V1", "V2 backlog", "follow-up spec"), and two positive trigger prompts phrased the way the issue authors actually asked verify it routes:
  - *"We shipped the V1 we specced — how do I spec the V2 features we deferred?"*
  - *"This feature already has a spec, but now we want OCR instead of the manual step — spec that change before building"*

## Non-overlap check

- **#73** (spec directory restructuring, stale since April) is the other approach to #180's storage question. This PR deliberately stays out of that territory — no directory conventions changed, just "don't overwrite; write a delta file". If #73 revives, the delta-spec guidance composes with any directory layout it picks.
- No other open PR touches the Spec Evolution topic; #195 and #175 touch this SKILL.md for unrelated reasons (confirmation UX, trigger phrasing) with no line overlap in the sections edited here.

## Verification

Rebased on current main (6bcfeb9), all tiers green locally:
- `validate-skills.js` — 24 skills, 0 errors
- `validate-commands.js` — 8 commands, parity holds across Claude/Gemini twins
- `run-evals.js` — 126 checks passed, 0 errors; both new trigger prompts route within top-3; rank-1 rate unchanged at 85%


addyosmani:
I like this one a lot. V1-scoping-with-a-V2-delta is a real question people hit the moment a spec survives contact with a second feature, and answering #150 and #180 inside the spec skill itself is the right home for it rather than a separate doc. Nice that you kept the command files in parity and shipped an eval case with it, that's the bar now. Let me read the guidance end to end to make sure it doesn't bloat the spec flow, but this is on-mission and close to a yes.

.claude/commands/spec.md
```diff
@@ -13,3 +13,5 @@ Begin by understanding what the user wants to build. Ask clarifying questions ab
 Then generate a structured spec covering all six core areas: objective, commands, project structure, code style, testing strategy, and boundaries.
 
 Save the spec as SPEC.md in the project root and confirm with the user before proceeding.
+
+If SPEC.md already exists and this request is a change to what it describes, do not overwrite it: write a delta spec (e.g. SPEC-<change-slug>.md) covering only the change, per the skill's Spec Evolution section.
```
federicobartoli: Could we mirror this guard in commands/spec.toml too? Antigravity still instructs the agent to always overwrite SPEC.md; validate-commands does not compare prompt bodies.

skills/spec-driven-development/SKILL.md
```diff
@@ -177,6 +180,41 @@ The spec is a living document, not a one-time artifact:
...
+[What will be true after the change.]
+
+## Success Criteria
+[Specific, testable conditions — including what must NOT change.]
+
+## Boundaries
+[Only deltas from the original spec's boundaries; everything else inherits.]
+```
+
+Save it alongside the original (e.g. `SPEC-postgres-migration.md` next to `SPEC.md`) and run it through the same gated workflow: human review, then plan, tasks, implement. Keep one change per delta spec — that keeps each plan/build cycle small and reviewable.
```
federicobartoli: Could we align this filename with the downstream workflow? /build auto only recognizes SPEC.md, docs/SPEC.md, or files under spec/, and does not allow root-level SPEC-*.md files as uncommitted planning artifacts. It may therefore select the old spec or stop at the clean-baseline check.

skills/spec-driven-development/SKILL.md
```diff
@@ -177,6 +180,41 @@ The spec is a living document, not a one-time artifact:
...
+### Scope each version explicitly
+
+During Specify, ideas that won't make this version go under **Out of Scope (Deferred)** — captured, not lost. This makes the V1 boundary enforceable: during implementation, anything not in the spec's requirements is out, even if it's written in the same file under Deferred. When an idea arrives mid-build, park it in the Deferred list and keep going; re-scoping happens between versions, not during one.
+
+### Follow-up changes get a delta spec
+
+Once a spec has shipped, a new requirement — swapping the storage engine, replacing a manual step with OCR, promoting a deferred V2 item — gets its own small spec rather than a hand-edit of the shipped one. A delta spec covers only what changes:
+
+```markdown
+# Spec Delta: [Change Name]
```
federicobartoli, line 198: How should a delta spec satisfy the existing “covers all six core areas” verification? This template intentionally omits Commands, Project Structure, Code Style, and Testing Strategy. Could the verification distinguish full specs from delta specs and require only the affected core areas?

skills/spec-driven-development/SKILL.md
```diff
@@ -177,6 +180,41 @@ The spec is a living document, not a one-time artifact:
 - **Commit the spec** — The spec belongs in version control alongside the code.
 - **Reference the spec in PRs** — Link back to the spec section that each PR implements.
 
+## Spec Evolution: V1 Scoping and Follow-Up Changes
+
+A spec describes one version of the work. New ideas and post-ship changes don't expand it — they queue behind it.
+
+### Scope each version explicitly
```
federicobartoli, line 187: Could we explicitly answer whether the spec itself should be labeled V1? The Deferred section explains how to capture the V2 backlog, but the version-labeling part of #150 remains ambiguous.

.claude/commands/spec.md
ayobamiseun: Good catch — I'd missed that `commands/` is a third twin and that `validate-commands` only diffs presence/descriptions, not bodies. Added the same guard line to `commands/spec.toml` in ce60dbc.

skills/spec-driven-development/SKILL.md
ayobamiseun: Confirmed on both counts: discovery wouldn't select a root-level delta spec, and the clean-baseline check would stop on it. Rather than move delta specs into `spec/` (which would drift into #73's storage-convention territory), ce60dbc updates all three `/build` twins: a root-level `SPEC-<change>.md` is now an accepted spec path (taking precedence over `SPEC.md` when the run is building that follow-up change) and `SPEC-*.md` is included in the planning-artifact exclusions. SKILL.md now states the convention is recognized downstream.

skills/spec-driven-development/SKILL.md
ayobamiseun: Agreed, the checklist item contradicted the delta template. In ce60dbc it now distinguishes the two: full specs cover all six areas; delta specs cover only the areas the change affects (at minimum objective, success criteria, and boundary deltas) and inherit the rest from the original.

skills/spec-driven-development/SKILL.md
ayobamiseun: Made it explicit in ce60dbc: the spec template title is now `# Spec: [Name] — V[N]`, and Spec Evolution says to label the version in the title and bump it when a shipped delta is folded back in, so the V1/V2 boundary is stated in the artifact itself rather than implied.

.claude/commands/build.md
```diff
@@ -28,8 +28,8 @@ Pick the next pending task from the plan. Then:
...
 
-1. **Require a spec.** Look only for a spec at a known path: `SPEC.md` at the repo root, `docs/SPEC.md`, or a file under `spec/`. A README or arbitrary doc does **not** count. If none exists, stop and tell the user to run `/spec` first — do not invent requirements.
-2. **Establish a clean baseline.** Run `git status --porcelain`. If there are uncommitted changes outside the expected planning artifacts (`SPEC.md`, `docs/SPEC.md`, `spec/*`, `tasks/plan.md`, `tasks/todo.md`), stop and ask the user to commit, stash, or confirm how to handle them. Autonomous per-task commits must not absorb unrelated local work, or the clean-rollback guarantee breaks.
+1. **Require a spec.** Look only for a spec at a known path: `SPEC.md` at the repo root, a root-level `SPEC-<change>.md` delta spec (per the spec-driven-development skill's Spec Evolution section — when one exists and the run is building that follow-up change, it takes precedence over `SPEC.md`), `docs/SPEC.md`, or a file under `spec/`. A README or arbitrary doc does **not** count. If none exists, stop and tell the user to run `/spec` first — do not invent requirements.
```
federicobartoli, line 31: The existing global tasks/plan.md behavior predates this change, but it becomes ambiguous with the newly introduced delta-spec workflow. After V1 ships, an existing V1 plan may remain; /build auto will then skip planning for SPEC-ocr.md because step 3 checks only whether tasks/plan.md exists.
One low-scope option would be to record the source spec in the generated plan—for example, Source spec: SPEC-ocr.md—and reuse an existing plan only when that value matches the selected spec. On a mismatch, /build auto could stop and ask whether to archive or replace the existing plan rather than silently reusing it. Would that fit the intended workflow? The same guard should be mirrored across the three /build twins.

.claude/commands/spec.md
```diff
@@ -13,3 +13,5 @@ Begin by understanding what the user wants to build. Ask clarifying questions ab
 Then generate a structured spec covering all six core areas: objective, commands, project structure, code style, testing strategy, and boundaries.
 
 Save the spec as SPEC.md in the project root and confirm with the user before proceeding.
+
```
federicobartoli, line 16: Consider: Would it be clearer to express this as an explicit new-spec/delta-spec branch? The preceding instructions are written unconditionally for a full SPEC.md, while this line introduces the exception for an existing shipped spec.

federicobartoli (changes requested):

.claude/commands/build.md
ayobamiseun: Agreed this was a real gap — spec selection became version-aware but plan reuse didn't. Implemented your suggestion in 5971d80 across all three /build twins: plans are stamped with `Source spec: <path>` at generation, reused only on a match, and a mismatch stops to ask whether to archive or replace. One addition: a plan with no `Source spec:` line is treated as a mismatch rather than assumed current, since silently trusting an unstamped legacy plan is the same failure mode. The planning skill's plan template now includes the line so standalone /plan output participates in the guard too.

.claude/commands/spec.md
ayobamiseun: Yes, clearer as a branch. 5971d80 restructures all three /spec twins to decide the flow up front: New spec (no SPEC.md, or unrelated request) runs the full six-area flow; Delta spec (SPEC.md exists and the request changes it) writes SPEC-<change-slug>.md per Spec Evolution. No behavior change, just the structure you suggested.

ayobamiseun:
@federicobartoli gentle nudge on this one — all six points from your two review rounds are addressed (ce60dbc for the first round, 5971d80 for the plan source-spec guard and the new/delta branch), with replies on each thread. Anything else you need from me here?

PR #384, opened 2026-07-10
docs: clarify native adapter directories

TheDenStudios:
## Summary

- Add a README section that explains the shared `skills/` directory versus native tool adapter directories.
- Document why `.claude-plugin/`, `.claude/commands/`, `.gemini/commands/`, `.codex-plugin/`, and `.github/workflows/` keep their tool-specific names.
- Frame those folders as discovery conventions/adapters rather than branding.

## Why

This makes the repo structure easier to understand for contributors and reinforces that portability comes from shared skills plus native per-tool adapters, not from renaming discovery paths.

## Validation

- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,28 @@ Skills are plain Markdown - they work with any agent that accepts system prompts
 
 </details>
 
+---
 
+## Repository Structure
+
+The repo keeps shared skill content separate from native tool adapters:
+
+| Path | Purpose |
+|------|---------|
+| `skills/` | Shared `SKILL.md` workflows used across agents |
+| `commands/` | Antigravity slash-command definitions |
+| `.claude/commands/` | Claude Code slash-command definitions |
+| `.gemini/commands/` | Gemini CLI slash-command definitions |
+| `.claude-plugin/` | Claude Code marketplace and plugin metadata |
+| `.codex-plugin/` | Codex plugin metadata |
+| `.agents/` | Agent/marketplace metadata for tools that read the shared agent format |
+| `.github/workflows/` | GitHub Actions validation for skills, commands, and plugin installs |
+| `agents/` | Reusable persona definitions for review, security, testing, and performance |
+| `docs/` | Tool-specific setup guides |
+| `references/` | Supporting checklists and engineering references |
+| `scripts/` | Validation and evaluation scripts used by CI |
+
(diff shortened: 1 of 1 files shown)


addyosmani:
This is the productive version of the earlier rename idea, and I'm glad it landed here instead. Documenting the shared skills/ directory versus the per-tool adapter dirs is what actually reduces the cross-agent confusion, without the blast radius of renaming anything that existing installs depend on. Reads clean and merges green. I appreciate you following the thread to a better answer.

Interested in @federicobartoli @nucliweb takes here as well.

federicobartoli (approved):
LGTM. This is the right resolution to the earlier rename thread: it clarifies the portability model without changing any native discovery paths that existing installs may depend on.

I especially like the “shared skills vs per-tool adapters” framing. It makes the repo structure easier to reason about for contributors: `skills/` remains the canonical shared workflow layer, while `.claude-plugin/`, `.gemini/commands/`, `.codex-plugin/`, `.github/workflows/`, etc. are documented as tool-native integration points rather than repo branding.

No blockers from me. This preserves compatibility and resolves the confusion in docs, which is exactly the safer fix here.

nucliweb:
Agree with Addy and Federico that this is the right resolution to the rename thread: the shared-skills-vs-per-tool-adapters model is exactly the useful framing, and documenting it instead of renaming discovery paths is the safe call. I checked the table against the tree and every directory it lists exists.

One thing worth resolving before this lands, which I think improves on the goal rather than blocking it: this adds a second structure section. The README already has a `## Project Structure` section further down (the ASCII tree), so we would end up with two layout listings ~130 lines apart, which tends to drift rather than reduce confusion. It is already drifting in practice:

- The existing ASCII tree still says `references/ # 5 supplementary checklists` (it is 7) and does not list `evals/` (added later).
- The new table, in turn, omits `hooks/` and `evals/` (both real and CI-relevant), and `.opencode/`.

So both listings are incomplete, in different ways, and now there would be two to keep in sync.

Suggestion: fold the adapter framing into the existing `## Project Structure` section instead of adding a parallel one. Concretely, replace the ASCII tree with this clearer table plus the "tool-native discovery conventions" paragraph (that paragraph is the real value here), and while doing so add `hooks/` and `evals/` and fix the `5 -> 7` count. That keeps a single source of truth for the layout, delivers exactly the shared-vs-adapter clarity everyone likes, and removes the duplication.

Happy to see the framing land; I would just consolidate to one structure section first.

PR #392, opened 2026-07-11
docs(security): harden package-manager supply-chain guidance

### LINKED ISSUE #391: Improve dependency supply-chain guidance for npm, pnpm, and Yarn
## Problem

PR #346 added a disciplined dependency-upgrade workflow and correctly delegates audit triage and supply-chain risk to `security-and-hardening`. However, that skill and its checklist still use npm-only examples (`npm audit`, `npm ci`) and only warn about dependency install scripts rather than defining a safe review gate.

That leaves two gaps:

1. pnpm and Yarn projects can receive the wrong audit or reproducible-install command.
2. A known-advisory audit does not detect a newly malicious package, and lifecycle scripts may execute before they are reviewed unless the pinned client has a verified default-deny policy.

## Proposed focused change

Extend the existing `security-and-hardening` skill and `references/security-checklist.md` to:

- locate the installation boundary correctly (workspace root vs independent nested project), then corroborate `packageManager`, lockfile, and CI signals;
- use native frozen/immutable installs and full-project audits for npm, pnpm, Yarn Modern, and Yarn Classic;
- block dependency lifecycle scripts before first execution, then record narrow, version-aware approvals using native controls (`allowScripts`, `allowBuilds`, or top-level `dependenciesMeta` as supported);
- require manual review for forced audit remediation, lockfile/script-policy diffs, registry signatures/provenance, and dependency ownership/release-age signals;
- keep detailed, version-sensitive commands in the existing security checklist so the main skill stays concise.

## Scope boundaries

- Preserve the existing skill description/frontmatter and routing behavior.
- Do not add a new skill, eval routing, CI workflow, or third-party dependency such as LavaMoat.
- Treat different lockfiles as a conflict only within the same installation boundary; independent subprojects may use different package managers.
- For unlisted managers, require checking the pinned version's official documentation rather than guessing an equivalent command.

## Acceptance criteria

- [ ] Guidance distinguishes npm, pnpm, Yarn Modern, and Yarn Classic.
- [ ] Workspace-root detection cannot mistake a member package for an independent installation root.
- [ ] Unreviewed dependency scripts are blocked before first execution.
- [ ] Version-specific controls and defaults are stated precisely.
- [ ] Automatic forced remediation is explicitly rejected.
- [ ] `node scripts/validate-skills.js` and `node scripts/run-evals.js` pass.

## Primary references

- [npm install-script approvals](https://docs.npmjs.com/cli/v11/commands/npm-install-scripts/) and [install policy](https://docs.npmjs.com/cli/v11/commands/npm-install/)
- [npm audit and signature/provenance verification](https://docs.npmjs.com/cli/v11/commands/npm-audit/)
- [pnpm approve-builds](https://pnpm.io/cli/approve-builds), [build settings](https://pnpm.io/settings#allowbuilds), and [audit](https://pnpm.io/cli/audit)
- [Yarn security defaults](https://yarnpkg.com/features/security), [audit scope](https://yarnpkg.com/cli/npm/audit), and [top-level dependenciesMeta policy](https://yarnpkg.com/configuration/manifest#dependenciesMeta)
- [OWASP npm supply-chain guidance](https://cheatsheetseries.owasp.org/cheatsheets/NPM_Security_Cheat_Sheet.html)

federicobartoli:
## Summary

Strengthens the existing dependency supply-chain guidance by making it package-manager-aware and safe before dependency lifecycle scripts are allowed to execute.

* resolves the correct installation boundary for workspaces and independent nested projects;
* distinguishes reproducible installs and full-project audits across npm, pnpm, Yarn Modern, and Yarn Classic;
* documents version-aware native install-script controls and a safe bootstrap path before first execution;
* requires review of forced remediation, lockfile and script-policy changes, signatures/provenance, ownership, and release-age signals;
* keeps detailed version-sensitive commands in the security checklist so the main skill remains concise.

## Why

The dependency-upgrade workflow added in #346 correctly delegates audit triage and supply-chain risk to `security-and-hardening`, but the existing guidance primarily uses npm examples and does not define a complete gate for reviewing lifecycle scripts before they execute.

Known-vulnerability auditing and install-script containment address different risks. This change makes that distinction operational without introducing a third-party dependency or a new workflow.

## Scope

* Updates the existing `security-and-hardening` skill and checklist.
* Preserves the existing frontmatter and routing behavior.
* Adds no new skill, CI workflow, or third-party dependency.
* Does not attempt exhaustive support for every JavaScript package manager; unlisted clients must be checked against the pinned version’s official documentation.

Fixes #391

### DIFF
--- a/references/security-checklist.md
+++ b/references/security-checklist.md
@@ -121,10 +121,13 @@ Never discover dependency lifecycle scripts by first executing an ordinary insta
 3. Record the narrowest native allow/deny policy at the installation boundary and commit it.
 4. Run a clean frozen/immutable install with that policy and verify the required packages still build.
 
+**Point-in-time snapshot:** Package-manager defaults and command names change quickly. Verify this matrix against the pinned client's current official documentation before relying on it.
+
 | Manager version | Native policy |
 |---|---|
-| npm 11.18 | Treat unreviewed scripts as allow-with-warning unless `strict-allow-scripts=true` is already enforced; otherwise bootstrap with `npm ci --ignore-scripts`. From the installation boundary, use the workspace-unaware `npm install-scripts ls`, keep approvals version-pinned, and record denials name-wide. |
-| Older or unknown npm | Bootstrap with `npm ci --ignore-scripts`. Keep scripts disabled unless the pinned version documents an enforceable policy and each required script has been reviewed. |
+| npm without verified granular approvals | Bootstrap with `npm ci --ignore-scripts`, or persist `ignore-scripts=true` when project-wide blocking is intended. Keep scripts disabled or deliberately upgrade before allowing any reviewed dependency script. |
+| npm 11.18.x (verified on 11.18.0) | Unreviewed dependency scripts run with a warning by default. Enforce `strict-allow-scripts=true` before a normal install, then use the workspace-unaware `npm install-scripts ls` from the installation boundary; keep approvals version-pinned and denials name-wide. |
+| npm 12.x (verified on 12.0.1) | Unreviewed dependency scripts are skipped by default; `strict-allow-scripts=true` makes their presence fail the install before execution. Use the same `npm install-scripts` review and approval flow. |
 | pnpm 11+ | Use `pnpm approve-builds` and commit `allowBuilds` decisions; `strictDepBuilds` defaults to `true`, so unreviewed builds fail. |
 | pnpm 10.26–10.x | Configure `allowBuilds` explicitly, or use `pnpm approve-builds` with the legacy `onlyBuiltDependencies` / `ignoredBuiltDependencies` lists. Set `strictDepBuilds: true`; its v10 default is `false`. |
 | pnpm 10.1–10.25 | `pnpm approve-builds` records the legacy lists; enable `strictDepBuilds` where supported (10.3+). |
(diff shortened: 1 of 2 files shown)


addyosmani:
This is a genuinely good hardening pass and the timing is right: blocking dependency lifecycle scripts before the first install is exactly the hole the recent npm postinstall worms drove through, so making it the default posture rather than an afterthought is the correct instinct. I also like that you kept the skill body at the principle level and pushed the version-specific matrix into the checklist, since that's the right split for keeping the skill readable while the detail lives where it belongs. The new "the audit passed, so it's safe" rationalization is a nice touch too, it's the exact wrong mental model this whole section is fighting.

My one real question is maintenance. The per-version bands (pnpm 10.1-10.25 vs 10.26 vs 11+, npm 11.18, Yarn 4.14+) are precise today, but package-manager defaults churn fast, so this matrix is the part most likely to quietly go stale and mislead someone in six months. You've hedged it well with the "verify against the official docs before changing policy" line and the links, which I think is the right call, so I'm not asking you to drop the specificity. It might just be worth a one-line note at the top of that table that it's a point-in-time snapshot to check against current docs, plus a quick spot-check that the newest flag names (npm install-scripts ls, strict-allow-scripts) match current npm, since a couple of those are recent enough that I'd want to be sure. Everything's green and it's a clear improvement over the npm-only version, so I'm inclined to take it once we've figured out how to keep that matrix honest over time.

nucliweb (approved):

federicobartoli:
> This is a genuinely good hardening pass and the timing is right: blocking dependency lifecycle scripts before the first install is exactly the hole the recent npm postinstall worms drove through, so making it the default posture rather than an afterthought is the correct instinct. I also like that you kept the skill body at the principle level and pushed the version-specific matrix into the checklist, since that's the right split for keeping the skill readable while the detail lives where it belongs. The new "the audit passed, so it's safe" rationalization is a nice touch too, it's the exact wrong mental model this whole section is fighting.
> 
> My one real question is maintenance. The per-version bands (pnpm 10.1-10.25 vs 10.26 vs 11+, npm 11.18, Yarn 4.14+) are precise today, but package-manager defaults churn fast, so this matrix is the part most likely to quietly go stale and mislead someone in six months. You've hedged it well with the "verify against the official docs before changing policy" line and the links, which I think is the right call, so I'm not asking you to drop the specificity. It might just be worth a one-line note at the top of that table that it's a point-in-time snapshot to check against current docs, plus a quick spot-check that the newest flag names (npm install-scripts ls, strict-allow-scripts) match current npm, since a couple of those are recent enough that I'd want to be sure. Everything's green and it's a clear improvement over the npm-only version, so I'm inclined to take it once we've figured out how to keep that matrix honest over time.

Good call. I’ve now checked the version boundary directly rather than treating the current npm behavior as uniform.

The command surface is valid in npm 11.18.0 and 12.0.1, and an isolated npm 12.0.1 probe with a real dependency `postinstall` confirmed the complete flow:

* the default install succeeded but blocked the unreviewed script;
* `npm install-scripts ls` surfaced it as pending;
* explicit approval followed by `npm rebuild` executed it;
* `strict-allow-scripts=true` failed with `ESTRICTALLOWSCRIPTS` before execution.

The lower boundary matters too: npm 11.12.1 does not expose the `install-scripts` namespace at all. I’m therefore going to make the checklist explicit across three cases rather than leaving older clients implicit:

* older npm versions without granular native approvals: bootstrap with lifecycle scripts disabled (`npm ci --ignore-scripts`, or a project-level `ignore-scripts=true` policy when persistence is intended), then keep them disabled or upgrade deliberately before allowing reviewed dependency scripts;
* npm 11.18: the approval interface exists, but unreviewed scripts require strict enforcement to become a hard failure;
* npm 12+: unreviewed scripts are skipped by default, while strict mode makes their presence fail the install.

I’ll also add the one-line version-sensitive snapshot note above the matrix and keep the first-party links as the authoritative check for the project’s pinned version. That should preserve the useful precision without silently treating either the newest defaults or the newest commands as universally available.

federicobartoli:
Implemented in 509281d.
The matrix is now explicitly marked as a point-in-time snapshot and separates:
npm without verified granular approval support;
npm 11.18.x, verified on 11.18.0;
npm 12.x, verified on 12.0.1.
Unlisted or future versions are directed back to the pinned client’s official documentation rather than inheriting newer defaults. All PR checks are green.

addyosmani (approved):
This is exactly the response I was hoping for, and then some. Actually probing the version boundaries beats the "add a caveat" fix I suggested, and finding that 11.12.1 has no install-scripts namespace at all is the kind of thing that would have quietly bitten someone. The three-case split plus the snapshot note lands it. Concern fully addressed, checks green, good to go from my side. Nice work @federicobartoli.

PR #393, opened 2026-07-11
Add fixture-backed behavioral evals

ayobamiseun:
## What changed

- Added fixture workspaces for three behavioral evals: test-driven-development, debugging-and-error-recovery, and code-simplification.
- Updated the corresponding eval case files to reference real files via files[].
- Removed provisional trust markers from these three evals because they now run against concrete fixture workspaces.

## Why

Behavioral evals without fixtures are useful as sanity checks, but they do not prove that an agent can operate on real code. These fixtures let the Tier 3 runner materialize small Node.js projects so graders can inspect actual tool calls, edits, and test runs instead of relying on narration.

## What was implemented

- test-driven-development: a small invoice total module and passing test suite. The eval asks the agent to add a failing fractional-cent rounding test before making the minimal fix.
- debugging-and-error-recovery: a pagination module with an intentional off-by-one bug. The eval requires reproducing the failure, identifying the root cause, fixing it minimally, and rerunning tests.
- code-simplification: a deliberately nested config parser with passing coverage. The eval asks the agent to simplify the parser while preserving behavior and avoiding feature creep.

## Validation

- node scripts/run-evals.js
- npm test in evals/fixtures/test-driven-development
- npm test in evals/fixtures/code-simplification
- npm test in evals/fixtures/debugging-and-error-recovery fails intentionally before the agent fix, showing the seeded pagination bug is present

### DIFF
--- a/evals/cases/code-simplification.json
+++ b/evals/cases/code-simplification.json
@@ -28,15 +28,19 @@
   "evals": [
     {
       "id": 1,
-      "prompt": "Simplify the provided 80-line function that parses config files, preserving exact behavior.",
+      "prompt": "Simplify the provided config parser while preserving exact behavior. Run the existing npm test suite before and after the refactor, reduce the nesting in src/config-parser.js, and do not add new features.",
       "expected_output": "A simpler implementation with identical behavior and a summary of what was removed and why",
+      "files": [
+        "code-simplification/package.json",
+        "code-simplification/src/config-parser.js",
+        "code-simplification/test/config-parser.test.js"
+      ],
       "expectations": [
         "Behavior is preserved (tests unchanged and passing, or equivalence argued concretely)",
         "Complexity is reduced rather than relocated",
         "The response explains what was removed and why it was safe",
         "No new features are added during the simplification"
-      ],
-      "trust_level": "provisional"
+      ]
     }
   ]
 }
--- a/evals/cases/debugging-and-error-recovery.json
+++ b/evals/cases/debugging-and-error-recovery.json
@@ -29,15 +29,19 @@
   "evals": [
     {
       "id": 1,
-      "prompt": "A test that passed yesterday now fails with an off-by-one error in pagination. Find and fix the root cause.",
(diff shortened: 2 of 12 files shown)


addyosmani:
Right direction, and the fixtures are real (`package.json` plus `src` and `test`, not stubs), which is exactly what makes a behavioral eval mean something. The snag is #381 is doing the same graduation across the whole catalog, including these three skills, so the two will collide on the same case files and fixtures.

Rather than land overlapping halves, let's coordinate: either fold these three into #381, or have #381 rebase around them. I don't want to merge this and then force a messy three-way conflict on the bigger PR. Let me know which way you and the #381 author want to split it.

ayobamiseun:
@addyosmani  That makes sense. Given #381 is the broader graduation PR and already covers these three skills, I think the cleanest path is to let #381 land first rather than make it rebase around this smaller slice.

I’m happy for any useful bits from this PR to be folded into #381 if they differ in a helpful way, but I don’t want to create extra conflict on the same case files and fixture paths. Once #381 has these covered, I can close this PR as superseded.

Thanks for the heads-up and for calling out the overlap before it became messy.

addyosmani:
Agreed, that's the clean path: let #381 land, then close this as superseded. I'll make sure #381 covers your three skills before it merges. Thanks for catching the overlap gracefully rather than racing it.

federicobartoli:
Closing this as superseded, following the plan agreed in the thread: on Jul 13 you offered to close this once #381 had the three skills covered, and on Jul 16 Addy confirmed to let #381 land and then close this one. #381 merged on Jul 16, and its files include eval cases and fixtures for all three skills (code-simplification, debugging-and-error-recovery, test-driven-development), so the condition is met. The branch has also gone conflicting against main in the meantime, so closing seems better than asking you to rebase work that now exists upstream.
Thank you for coordinating this instead of racing the conflict, it made the overlap easy to resolve. If you notice anything your fixtures covered that #381's versions miss, a small follow-up PR scoped to that gap would be very welcome.

PR #394, opened 2026-07-11
Add Kiro setup guide

ayobamiseun:
## What changed

- Added a dedicated Kiro setup guide at docs/kiro-setup.md.
- Updated the README Kiro section to point to the new guide.
- Clarified the supported Kiro install paths for this repo's skills: individual GitHub imports, workspace skills under .kiro/skills/, and global skills under ~/.kiro/skills/.

## Why

The README mentioned Kiro, but it did not have a dedicated setup guide like the other supported tools. Kiro also has a few important details worth documenting clearly: GitHub imports target an individual skill folder or SKILL.md, workspace and global skills use different directories, and project-specific guidance belongs in Kiro steering rather than copying this repo's root AGENTS.md.

## What was implemented

- Documented how to import individual skills through Kiro's Agent Steering & Skills panel.
- Added workspace and global copy-based setup examples.
- Added a recommended starter skill set.
- Explained how Kiro uses Skills, Steering, and Powers differently.
- Added maintenance notes for refreshing imported skills.

## Validation

- node scripts/run-evals.js

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,8 +159,10 @@ Use agent definitions from `agents/` as Copilot personas and skill content in `.
 </details>
 
 <details>
-  <summary><b>Kiro IDE & CLI </b></summary>
-  Skills for Kiro reside under ".kiro/skills/" and can be stored under Project or Global level. Kiro also supports Agents.md. See Kiro docs at https://kiro.dev/docs/skills/
+<summary><b>Kiro</b></summary>
+
+Import individual skills through Kiro's Agent Steering & Skills panel, or copy this repo's `skills/` directories into `.kiro/skills/` for workspace scope or `~/.kiro/skills/` for global scope. See [docs/kiro-setup.md](docs/kiro-setup.md).
+
 </details>
 
 <details>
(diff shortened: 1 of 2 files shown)


addyosmani:
Sensible addition. Kiro was already in the README, so this just gives it the same dedicated guide the other platforms have. One ask before it lands: double-check the three install paths against current Kiro docs (the `.kiro/skills` and `~/.kiro/skills` locations especially), since platform paths drift and a setup guide that's subtly wrong is worse than the one-liner it replaces. If those check out, it's good to go.

ayobamiseun:
Thanks, agreed. I double-checked the current Kiro docs before writing this and re-checked them now.

The three path/import details in the guide match the docs:

- Workspace skills live under `.kiro/skills/`
- Global skills live under `~/.kiro/skills/`
- GitHub imports must point to a skill folder or directly to `SKILL.md`, not the repository root

The docs also confirm that workspace skills take priority over global skills when names conflict, and that imported skills are copied into the skills directory.

Reference checked: https://kiro.dev/docs/skills/

So I think the setup guide is aligned with current Kiro behavior.

addyosmani:
Thanks for re-checking against the current docs, that's exactly what I wanted. The three paths and the priority/copy behavior all match kiro.dev, and Kiro is already a supported platform so this just gives it the dedicated guide the others have. Verified and good to merge.

ayobamiseun:
@addyosmani gentle nudge — this one has your "verified and good to merge" from Jul 16 and is still mergeable with no conflicts. Anything still needed from me before it lands?

PR #395, opened 2026-07-12
Feature/tests skill test agent

summergan:


addyosmani:
I can't really review this as-is. It's 23 files and +1100 with no description, it's showing conflicts against main, and it overlaps your own #390 (both touch `test-engineer.md` and the testing eval cases). Right now I can't tell what's intentional versus leftover.

Could you rebase it, add a description of what it changes and why, and scope it down to one coherent thing? Happy to look again once it's focused and green, and let's make sure it and #390 aren't stepping on each other.

PR #396, opened 2026-07-12
docs: add adoption guide for greenfield vs brownfield rollout

nucliweb:
## What

Adds `docs/adoption-guide.md`: how to roll out agent-skills depending on where the codebase is in its life.

- **Greenfield:** full lifecycle from day one, with the quality gates compounding from the first commit.
- **Brownfield:** an incremental, verification-first path that reads and protects the codebase (context, review, characterization tests) before it changes it, then runs the full lifecycle on new work.
- A "which path are you on?" signal table, per-path anti-patterns, and a convergence summary showing both paths reaching the same steady state.

## References

Linked from two places so it is discoverable without duplicating content:
- **README** gains a short `## Adoption` section pointing to the guide.
- **docs/getting-started.md** links it from Recommended Setup as the in-depth companion to the quick setup (getting-started covers installation mechanics; the guide covers rollout strategy).

Docs only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,12 @@ Skills are plain Markdown - they work with any agent that accepts system prompts
 
 ---
 
+## Adoption
+
+Already installed? How you roll the pack out depends on your codebase. The **[Adoption Guide](docs/adoption-guide.md)** covers two paths: the full lifecycle from day one for a greenfield project, or an incremental, verification-first rollout for an established codebase.
+
+---
+
 ## All 24 Skills
 
 The commands above are entry points. The pack includes 24 skills total — 23 lifecycle skills plus the `using-agent-skills` meta-skill. Each skill is a structured workflow with steps, verification gates, and anti-rationalization tables. You can also reference any skill directly.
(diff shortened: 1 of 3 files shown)


addyosmani:
This is really good, and the brownfield half is what earns it: context-first, characterization tests before any change, Chesterton's Fence on simplification, the Beyonce Rule, Hyrum's Law at the seams. That's the genuinely hard adoption story to get right, and you mapped it onto the actual skills accurately, including the two-speed "legacy stays under the Phase 1-2 regime, new work gets the greenfield treatment" framing. The convergence table is a nice close.

One accuracy fix before it lands: describing the code-review severity labels as "(Nit/Optional/FYI)" undersells the taxonomy, since the point of it is the Required (no-prefix) and Critical labels that actually gate a merge, with Nit/Optional/FYI being the don't-block-on-this side. In a doc about protecting a legacy codebase that's the half you'd most want to name. I'd either include a blocking label or just say "severity labels" and let the skill define them. Small thing, everything else checks out and CI is green.

nucliweb:
Good catch, fixed in `27dcc32`. You're right that naming only Nit/Optional/FYI sold the taxonomy short, especially in a doc about guarding legacy where the blocking half is the point. Reworded to "severity labels (which separate what blocks a merge, Critical and Required, from what doesn't)", so it names the gating side without re-enumerating the full taxonomy the skill owns.

addyosmani (approved):
That's the fix, and the reword is cleaner than what I suggested: naming Critical and Required as the blocking half without re-enumerating the taxonomy the skill owns. My one concern's resolved, the doc reads well, CI is green. Good to merge from me.

addyosmani:
Looks good to me!

PR #397, opened 2026-07-12
docs: add developer onboarding guide for repo contributors

nucliweb:
## What

Adds `docs/developer-onboarding.md`: a guided tour for people working **on** the repo itself (as opposed to `getting-started.md`, which is for people *using* the skills in their own projects).

It covers:
- **The mental model:** the five composable layers (skills, personas, commands, references, evals) and the two structural rules (user/command is the orchestrator; do not duplicate, reference).
- **Local setup** and **the verification loop**, mapping each CI check to the local command that reproduces it.
- **Contribution paths** (fix an existing skill / propose a new one / docs and harness), a **pre-PR checklist**, and a **suggested reading order**.

It is a map, not a rulebook: it points to `CONTRIBUTING.md`, `skill-anatomy.md`, and `evals/README.md` as the authoritative sources and tells you when to read each, rather than restating them.

## Reference

Linked from the top of `CONTRIBUTING.md` as the guided tour that complements the rules. Deliberately not added to the README in this PR, since the README currently has several open PRs in flight and this stays discoverable via README to CONTRIBUTING to onboarding.

Docs only. All internal links resolve; validators stay green.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,8 @@
 
 Thanks for your interest in contributing! This project is a collection of production-grade engineering skills for AI coding agents.
 
+New here? [docs/developer-onboarding.md](docs/developer-onboarding.md) is a guided tour of how the repo fits together (the five layers, the verification loop, and the contribution paths) and tells you when to read this document, [skill-anatomy.md](docs/skill-anatomy.md), and [evals/README.md](evals/README.md). This file is the authoritative rulebook; the onboarding guide is the map.
+
 ## Adding a New Skill
 
 ### Before proposing a new skill
--- a/docs/developer-onboarding.md
+++ b/docs/developer-onboarding.md
@@ -1,0 +1,121 @@
+# Developer Onboarding
+
+This guide is for people working **on** the agent-skills repository itself: contributing skills, fixing docs, improving the eval harness. If you want to *use* the skills in your own projects, you're looking for [getting-started.md](getting-started.md) instead.
+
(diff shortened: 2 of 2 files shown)


addyosmani:
Clean and genuinely useful, and I like that you built it as a map rather than another rulebook. The five-layer table plus the "user/command is the orchestrator" and "don't duplicate, reference" rules are exactly the two things new contributors get wrong. Mapping each CI check to the local command to run is the detail that'll save people a bounced PR, and the suggested reading order is a nice touch. I checked the specifics and they hold up, including the translations policy link and the three-command-directory parity claim.

One thing to keep an eye on: a couple of sections (the new-skill requirements, the pre-flight) restate CONTRIBUTING rather than pointing at it, which is the "don't duplicate, reference" rule you're the one championing. It's light and you frame the doc as a tour, so I'm not blocking on it, just flagging so the two don't drift apart later. Good to go otherwise.

nucliweb:
Fair, and a bit ironic given the doc preaches exactly that rule. Fixed in `7209bd1`: the "ships as a set" block now keeps only the shape (SKILL.md + eval case + optional scripts, references never inside the skill) and defers the exact frontmatter rules, section anatomy, and eval-case minimums to CONTRIBUTING.md and skill-anatomy.md, so they can't drift out of sync. Left the pre-flight paragraph as is since it already links out; happy to trim it further if you'd prefer.

addyosmani:
Thanks so much @nucliweb!

PR #399, opened 2026-07-13
feat(skills): add dependency-management skill

Devendra2016:
Adds the dependency upgrade lifecycle skill (inventory, assess, batch, upgrade, verify) covering routine upgrades, major-version bumps, lockfile hygiene, and recovery from broken updates. Fills a gap not covered by security-and-hardening (advisory triage, supply-chain vetting) or deprecation-and-migration (replacing one system with another); the skill explicitly routes to both when appropriate.

Includes evals/cases/dependency-management.json with 3 positive triggers, 2 negatives (routed to security-and-hardening and deprecation-and-migration), and 1 behavioral eval (provisional trust level).

Validation:
- validate-skills.js: 25 skills, 0 errors, 0 warnings
- run-evals.js: 129/129 checks pass, trigger rank-1 rate 86%
- validate-commands.js: command parity PASSED

### DIFF
--- a/evals/cases/dependency-management.json
+++ b/evals/cases/dependency-management.json
@@ -1,0 +1,44 @@
+{
+  "skill_name": "dependency-management",
+  "trigger": {
+    "positive": [
+      {
+        "prompt": "Should I bump React from 17 to 18, and how do I do it safely?",
+        "top_k": 3
+      },
+      {
+        "prompt": "Help me batch these outdated npm packages into reviewable PRs",
+        "top_k": 3
+      },
+      {
+        "prompt": "Review the lockfile diff on this dependency bump before we merge",
+        "top_k": 3
+      }
+    ],
+    "negative": [
+      {
+        "prompt": "Audit this file upload handler for vulnerabilities",
+        "owner": "security-and-hardening"
+      },
+      {
+        "prompt": "Sunset the v1 API and move users over to v2",
+        "owner": "deprecation-and-migration"
+      }
+    ]
+  },
+  "evals": [
+    {
+      "id": 1,
+      "prompt": "Upgrade a Node service from Express 4 to Express 5, keeping the test suite green and shipping the change in reviewable increments.",
+      "expected_output": "An assessment of the 4→5 breaking changes from the changelog, a one-major-per-PR batching plan, the lockfile diff inspected, the suite green, runtime smoke verified, and a regression test added for any behavior change",
+      "expectations": [
+        "The Express 4→5 changelog is read and breaking changes are named before the bump is applied",
+        "The major upgrade is shipped in its own PR/commit, not bundled with other dependency changes",
+        "The lockfile diff is inspected for unexpected transitive changes",
+        "Verification goes beyond the test suite: the build succeeds and a runtime smoke check is performed",
+        "Any behavior change introduced by the bump is covered by a regression test before re-bumping"
+      ],
+      "trust_level": "provisional"
+    }
+  ]
+}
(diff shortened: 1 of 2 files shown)


addyosmani:
The five-step lifecycle is well-shaped and you clearly thought about the boundaries with the other skills. The problem is timing: since you opened this, #346 landed a dependency-upgrade workflow directly in `code-review-and-quality` (changelog-first, one-package-per-change, lockfile discipline), and #392 hardened the supply-chain side in `security-and-hardening`. So the gap this claims is now mostly covered across two existing skills.

Before this can move I'd need the case for why it's a standalone skill rather than folding the extra lifecycle depth into the code-review section that now exists, per the pre-flight in CONTRIBUTING. New skills are also a catalog-growth call I'm still holding. Good content, most likely the wrong shape.

federicobartoli:
Thanks for the contribution , there are useful ideas here, particularly the inventory, risk assessment, runtime verification, and recovery workflow.
That said, I’m not yet convinced that this should be a separate skill in its current form.
#346 already added the dependency-upgrade review workflow to code-review-and-quality, while #392 expanded security-and-hardening with package-manager and supply-chain guidance. This PR adds some operational coverage that those changes do not fully provide, but it also repeats much of their core guidance: changelog review, isolated upgrades, lockfile inspection, audit handling, and verification.
The ownership boundary is also unclear. For example, the new description and evals treat “reviewing a lockfile diff before merge” as a trigger for dependency-management, even though that naturally belongs to code-review-and-quality. Because the eval only requires top_k: 3 and does not test code-review-and-quality as a negative owner, it can pass without resolving that routing ambiguity.
I also have concerns about the “across any package manager” scope. The proposed pip and Cargo commands do not necessarily update both the project manifest and lockfile in the way the workflow requires. We should either narrow the initial scope to package managers whose workflows are fully verified, or ground each ecosystem-specific path in its official documentation.
Before merging this as a new skill, I think we need:
A sharper explanation of what behavior is not already covered by #346 and #392.
A clear ownership boundary: performing/planning upgrades here, reviewing the resulting PR in code-review-and-quality, and security triage in security-and-hardening.
Pairwise routing evals against code-review-and-quality.
Verified or narrower package-manager guidance.
If the remaining unique workflow is small after that separation, I would prefer a focused extension to an existing skill rather than a new directory. If it remains a substantial, independently triggered operational workflow, then a dedicated skill could be justified.
I’d like to hear the other maintainers’ views as well, but I would not approve the PR in its current form.

addyosmani:
Having read the whole thing now, I can be specific about the overlap Federico flagged. The genuinely new material is the batching strategy, the routine cadence, and the recovering-from-a-broken-upgrade workflow, none of which #346 or #392 cover. But the Assess step (changelog, classify the bump, deprecation) and the lockfile-hygiene section largely restate what #346 put in `code-review-and-quality` and #392 put in `security-and-hardening`, and the skill points at both mid-section, which is the tell. So it's roughly half novel, half duplication.

Trimming the duplicated halves down to references (the way you already do for the security triage) would cut this a lot and show whether the novel remainder, keep-your-deps-current-over-time, is really its own skill or a section elsewhere. And "is it a new skill at all" is the catalog-growth call that's on hold right now, so I don't want to greenlight a 250-line standalone ahead of that. The batching/cadence/recovery thinking is the part worth keeping whichever way it lands.

addyosmani:
Thanks. Dependency work is covered by the dependency-upgrade workflow in code-review-and-quality, so a separate skill would duplicate it. Closing.

PR #400, opened 2026-07-14
fix(code-review): keep standalone review self-contained

astandrik:
## Problem

`skills add --skill code-review-and-quality` copies only the selected skill directory. The skill referenced two repository-level files, so standalone installs pointed to files that were not present.

## Fix

Keep the existing security and performance baseline in `SKILL.md`. Use `security-and-hardening` and `performance-optimization` when available, and state when deeper review was not performed. Remove the two broken root-reference links.

Shared-reference packaging for other skills remains out of scope and is tracked in #361.

## Verification

- isolated `skills@1.5.17` install contains only `SKILL.md`, no missing references, and both fallback notices
- `node scripts/validate-skills.js`: 24/24
- `node scripts/run-evals.js`: 124/124
- `node scripts/validate-commands.js`: 8/8
- `git diff --check origin/main...HEAD`

### DIFF
--- a/skills/code-review-and-quality/SKILL.md
+++ b/skills/code-review-and-quality/SKILL.md
@@ -63,7 +63,9 @@ Does the change fit the system's design?
 
 ### 4. Security
 
-For detailed security guidance, see `security-and-hardening`. Does the change introduce vulnerabilities?
+The checks below are self-contained. Use `security-and-hardening` for deeper analysis when it is available.
+
+Does the change introduce vulnerabilities?
 
 - Is user input validated and sanitized?
 - Are secrets kept out of code, logs, and version control?
@@ -76,7 +78,9 @@ For detailed security guidance, see `security-and-hardening`. Does the change in
 
 ### 5. Performance
 
-For detailed profiling and optimization, see `performance-optimization`. Does the change introduce performance problems?
+The checks below are self-contained. Use `performance-optimization` for measurement-driven investigation when it is available.
+
+Does the change introduce performance problems?
 
 - Any N+1 query patterns?
 - Any unbounded loops or unconstrained data fetching?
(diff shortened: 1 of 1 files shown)


federicobartoli:
Thanks for addressing the broken standalone references. The patch works for this individual skill, but I don’t think we should establish a skill-specific workaround while the repository-wide portability approach in #361 is still unresolved.
This PR chooses one particular policy , remove the shared references and replace them with inline baselines plus optional specialist-skill fallbacks , while the same underlying issue remains across the skill pack. Depending on the eventual #361 decision, this change may need to be revised or reverted.
I would prefer to hold this PR until the shared-reference strategy is settled, then apply the chosen approach consistently. The implementation itself is reasonable; my concern is that merging it now would prematurely commit one skill to a design that has not yet been agreed upon.

addyosmani:
Sensible fix for the per-skill-install gap, same class as #361, and keeping the security/performance baseline inline while soft-referencing the fuller skills "when available" is the right call for a standalone install.

One thing to double-check: make sure the inline baseline doesn't drift from what `security-and-hardening` and `performance-optimization` actually say, since the whole reason we reference instead of duplicate is to keep two copies from going out of sync. If it's a genuinely minimal baseline that points at the real skills for depth, that's fine. Merges clean and green.

nucliweb:
This is the same #361 class I have been tracking on #236 and #366, so I took a close look, especially at the drift concern @addyosmani raised.

On drift: the inline checklists here are not a new second copy. The security and performance checklist items already lived in this `SKILL.md`; the diff only relabels them as "the self-contained baseline," softens the sibling-skill links to "when available," and drops the two See Also links to `references/security-checklist.md` and `references/performance-checklist.md`. So it introduces no new duplication to keep in sync, and it still points at `security-and-hardening` / `performance-optimization` for depth, which is the "genuinely minimal baseline pointing at the real skills" case you said is fine.

Verified standalone and green:
- No `references/` root links remain in the SKILL.md, so a per-skill install has no dangling links.
- `validate-skills.js` (24/24) and `run-evals.js` (124/124) pass locally.

On @federicobartoli's hold-for-#361 point, which is fair: this is now the third distinct #361 approach in flight, alongside #366's canonical GitHub URLs and the skill-local `references/` convention that merged in #236. But this PR is narrower than a pack-wide policy: the two links it removes are already broken for standalone installs today, and everything else is a relabel of pre-existing content. So it is a strict improvement for this one skill no matter how #361 settles, and trivial to re-align to the eventual convention. That is why I would lean it can land without pre-empting the #361 decision.

PR #402, opened 2026-07-14
docs(adr): match an existing ADR convention before creating one

CybotTM:
## What

Adds a short "Match the existing convention first" step to the `documentation-and-adrs` skill, before the ADR Template section.

## Why

The template hard-codes `docs/decisions/` Markdown with its own numbering and headings. Many repos already have a different ADR convention — a different location or markup (e.g. `Documentation/Decisions/*.rst`, MADR, `adr-tools`), sometimes introduced in a **not-yet-merged PR** or an org-level `.github` repo. Following the template blindly then creates a second, inconsistent scheme.

The added step: detect an existing convention (location/format, numbering/naming, headings) — scanning open PRs and the org `.github` repo — and continue it; fall back to the `docs/decisions/` default only when none exists.

One line added, no restructuring. markdownlint adds no new findings (the file has pre-existing default-ruleset warnings unrelated to this change; the repo ships no markdownlint config).

### DIFF
--- a/skills/documentation-and-adrs/SKILL.md
+++ b/skills/documentation-and-adrs/SKILL.md
@@ -33,9 +33,19 @@ ADRs capture the reasoning behind significant technical decisions. They're the h
 - Choosing between build tools, hosting platforms, or infrastructure
 - Any decision that would be expensive to reverse
 
+### Match the existing convention first
+
+Before creating an ADR, check whether the project already has one — an established convention overrides the defaults below. Scan the repo **and**, for a GitHub org, its `.github` repo, including **open PRs** (a convention may be introduced in a not-yet-merged PR):
+
+- **Location and format** — e.g. `docs/adr/*.md`, `Documentation/Decisions/*.rst`, a MADR layout, or an `adr-tools` setup. Match the existing directory, file extension, and markup (Markdown vs reStructuredText).
+- **Numbering and naming** — continue the existing sequence and filename pattern (`ADR-004-Title.rst`, `0004-title.md`, …); don't restart at 001 or introduce a second scheme.
+- **Section headings** — reuse the project's heading set rather than imposing this template's.
+
+Only when no convention exists do you apply the default below.
+
 ### ADR Template
 
-Store ADRs in `docs/decisions/` with sequential numbering:
+Store ADRs in `docs/decisions/` with sequential numbering (unless the project already uses another location — see above):
 
 ```markdown
 # ADR-001: Use PostgreSQL for primary database
--- a/skills/documentation-and-adrs/SKILL.md
+++ b/skills/documentation-and-adrs/SKILL.md
@@ -35,13 +35,13 @@ ADRs capture the reasoning behind significant technical decisions. They're the h
 
 ### Match the existing convention first
 
-Before creating an ADR, check whether the project already has one — an established convention overrides the defaults below. Scan the repo **and**, for a GitHub org, its `.github` repo, including **open PRs** (a convention may be introduced in a not-yet-merged PR):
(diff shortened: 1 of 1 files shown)


addyosmani:
Good addition, and the right instinct. The template hardcoding `docs/decisions/` Markdown was quietly telling agents to impose our convention on repos that already have one (MADR, `.rst` under `Documentation/Decisions`, whatever), which is exactly what an ADR skill shouldn't do. Matching the existing convention first is the correct default. Merges clean, suite's green. Happy to take it.

CybotTM:
Btw. discovered with https://github.com/netresearch/retro-skill ;-)

skills/documentation-and-adrs/SKILL.md
```diff
@@ -33,9 +33,19 @@ ADRs capture the reasoning behind significant technical decisions. They're the h
 - Choosing between build tools, hosting platforms, or infrastructure
 - Any decision that would be expensive to reverse
 
+### Match the existing convention first
+
+Before creating an ADR, check whether the project already has one — an established convention overrides the defaults below. Scan the repo **and**, for a GitHub org, its `.github` repo, including **open PRs** (a convention may be introduced in a not-yet-merged PR):
```
federicobartoli: Could we keep this check entirely repository-local and forge-agnostic? A reusable skill should not require network access or assume GitHub, a particular collaboration platform, or any remote transport. Unmerged remote changes are also proposals rather than established project conventions.
Suggested wording:
“Before creating an ADR, inspect the available repository context for an established convention, including existing ADRs, project instructions, and ADR-related configuration or tooling. Match its location, format, numbering, naming, and section structure. If the available evidence conflicts, surface the conflict rather than silently introducing another scheme. Only use the default below when no convention can be established.”
This achieves the core goal without depending on GitHub, SSH, external permissions, or network availability.

federicobartoli (changes requested):

addyosmani:
Re-reviewing this more carefully than my first pass. The core point, match an existing ADR convention before imposing `docs/decisions/`, is right and worth adding; an agent genuinely will steamroll a repo's existing MADR or `.rst` setup.

Where I'd push back is the "scan open PRs (and the org's `.github` repo) for a convention introduced in a not-yet-merged PR" clause. An ADR convention lives in the repo you're already in, and sending the agent off to read open PRs for markup conventions is real instruction weight for a case that almost never fires. Trim that to "check the repo (and its `.github`)" and drop the open-PR scanning, and it's a clean merge.

CybotTM:
I did not invent the issue, It is from real world. Anyway, let me check ...

skills/documentation-and-adrs/SKILL.md
CybotTM: Reworded in a0eba71 per your suggestion: detection is now repository-local only (existing ADRs, project instructions, ADR tooling config), the `.github`-repo and open-PR scanning is gone, and conflicting evidence is surfaced instead of silently adding another scheme. Kept the three match-criteria bullets (location/format, numbering/naming, section headings) as the concrete checklist.

addyosmani:
Good question, and your forge-agnostic read is better than what I wrote. When I said "check the repo (and its `.github`)" I was loosely carrying over the org-level `.github` repo from the original text, not asking to preserve a GitHub-specific mechanism, so nothing I intended to keep was lost. You're right that GitHub shouldn't be a special case in a general ADR skill.

Sebastian's current wording is the right shape: "existing ADRs, project instructions, and ADR-related configuration or tooling (e.g. `.adr-dir`)" covers a local `.github/` naturally through "project instructions" without naming a forge, and it reads the same for GitLab, Gitea/Forgejo, or plain git. The `.adr-dir` example is a nice concrete, tool-level signal rather than a platform one, and the conflict-surfacing line is a good add. I'm happy with it as-is. Go ahead and resolve the review, good to merge from me.

federicobartoli (approved):

federicobartoli:
LGTM!

PR #222, opened 2026-06-09
feat(agents): add web-performance-auditor persona and /webperf command

nucliweb:
Closes #85

## Summary
- Add `agents/web-performance-auditor.md`: a specialist Web Performance Engineer persona with Quick mode (source analysis) and Deep mode (Lighthouse JSON, PageSpeed Insights JSON, CrUX API, DevTools trace, live Chrome DevTools MCP)
- Add `.claude/commands/webperf.md`: `/webperf` slash command that invokes the persona for performance-focused audits on web applications
- Update `agents/README.md`: register the new persona in the table and add usage examples for direct invocation and via `/webperf`

## Test plan
- [ ] Quick mode: point `/webperf` at a component with no artifacts — findings are labeled `potential impact`, scorecard shows `not measured`
- [ ] Deep mode: pass a Lighthouse JSON report — scorecard populates with `Lab (Lighthouse)` source labels
- [ ] Framework detection: verify the persona does not recommend React-specific fixes to a Vue or vanilla project
- [ ] `/webperf` is not triggered on CLI tools or server-only code

Issue #398, opened 2026-07-13
Proposal: Selectively Preserve Tiny, Scoped Repository-Memory Capsules


federicobartoli:
## Summary

This issue proposes a pragmatic agent skill, tentatively named `repository-memory`.

The central idea is not to generate repository documentation in a separate phase. It is to preserve the smallest durable semantic discoveries agents make while doing verified engineering work:

> As agents develop a repository, they selectively promote non-obvious, durable knowledge into tiny, scoped capsules. Future agents retrieve only relevant capsules that pass freshness checks; when validation is uncertain, they fall back to investigating the repository.

Most tasks should produce no memory update. A capsule is created or changed only when an agent had to reconstruct knowledge that would be expensive and useful to reconstruct again.

This is specialized living documentation distinguished by its protocol and governance: admission during real work, minimal deltas, explicit invalidation dependencies, human review, selective retrieval, and refusal of uncertain memory.

## Problem

Agents repeatedly inspect the same files to rediscover:

- what a subsystem is responsible for;
- how control flows across several files;
- which invariants must remain true;
- which boundaries and dependencies are non-obvious;
- which recurring failure modes have already required investigation.

Code remains the source of truth, but this knowledge is often distributed across implementation, tests, configuration, and architecture decisions. Once a task ends, the reconstructed mental model disappears and the next agent pays the same discovery cost.

A broad repository overview is not an adequate substitute: it is loaded too often, contains irrelevant information, and becomes stale as a large unit.

## Core workflow

Colocate semantic capsules with the high-value subsystems or code scopes they describe:

```text
src/
├── auth/
│   ├── AGENT_MEMORY.md
│   ├── session.ts
│   └── oauth-state.ts
├── billing/
│   ├── AGENT_MEMORY.md
│   └── reconciliation.ts
└── jobs/
    ├── AGENT_MEMORY.md
    └── scheduler.ts
```

Do not create one capsule per folder. Add `AGENT_MEMORY.md` only when real development work exposes durable knowledge worth preserving for that area. Cross-cutting knowledge belongs at the nearest common ancestor of the code it describes, with explicit scope metadata.

```text
DEVELOP normally
  → DISCOVER a non-obvious fact
  → VERIFY it against current code, tests, configuration, or an ADR
  → DECIDE whether it is durable and likely to prevent future rediscovery
  → PROPOSE the smallest possible capsule delta
  → REVIEW it with the code change
  → RETRIEVE it selectively in future tasks
[...]

PR #403, opened 2026-07-14
docs(performance): add Step 4 (verify with keep-or-revert)

asiridalugoda:
## What

The optimization workflow declares five steps, but only three have sections of their own. `VERIFY` and `GUARD` stop at their one-line summary in the workflow code block. This adds the missing **Step 4: Verify (Keep or Revert)**, plus matching entries in Common Rationalizations, Red Flags, and Verification.

## Why

The skill is strong on *measure first* and quiet on what to do with the measurement afterwards. That is where performance work usually goes wrong: not in choosing the optimization, but in what happens when the number comes back.

- The result is re-measured under different conditions than the baseline, so the delta is noise.
- Several changes ship together and no single one can be attributed.
- The result is flat, but the code is already written, so it lands anyway: complexity that never bought anything.
- A "win" that turned a test red gets merged as a win.
- A failed experiment leaves no trace in git history, so someone re-runs it two quarters later.

Step 4 makes the decision explicit and strict: past the threshold with tests green keeps; within noise, worse, or test-breaking reverts. Neutral is a revert. And it says to log every attempt, including the reverted ones, because those are the ones that get repeated.

The discipline is the one Karpathy's [autoresearch](https://github.com/karpathy/autoresearch) uses for overnight experiment loops (fixed measurement budget, one change at a time, git advance-or-reset on the result, a ledger per trial). It generalises to any metric-driven work.

## Pre-flight checks (per CONTRIBUTING)

- **Searched the catalog.** Step 1 establishes a baseline and Verification asks for before/after numbers, but no skill states a keep-or-revert rule, the one-change-at-a-time constraint, a noise threshold, or an attempt ledger. `references/performance-checklist.md` does not cover it either.
- **Checked open PRs.** None touch `skills/performance-optimization/SKILL.md`. #287 (`llm-cost-optimization`) is a different metric in a proposed new skill; #222 (`web-performance-auditor`, merged) is a persona, not this workflow step.
- **Extends rather than adds.** Focused edit to an existing skill, no new directory.

## Validation

- `node scripts/validate-skills.js` -> 24 skills checked, 0 errors, 0 warnings
- `node scripts/run-evals.js` -> 124 checks passed, 0 errors
- `git diff --check` -> clean; frontmatter unchanged

Scope: one file, +47 / -1.

### DIFF
--- a/skills/performance-optimization/SKILL.md
+++ b/skills/performance-optimization/SKILL.md
@@ -33,7 +33,7 @@ Measure before optimizing. Performance work without measurement is guessing — 
 1. MEASURE  → Establish baseline with real data
 2. IDENTIFY → Find the actual bottleneck (not assumed)
 3. FIX      → Address the specific bottleneck
-4. VERIFY   → Measure again, confirm improvement
+4. VERIFY   → Measure again; keep or revert
 5. GUARD    → Add monitoring or tests to prevent regression
 ```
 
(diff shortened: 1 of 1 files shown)


addyosmani:
This is really good, and it fills the exact gap the skill had: it said "measure again, confirm improvement" but had no teeth for when the number doesn't move. "Neutral is a revert" is the line that earns it, that's the step teams actually skip, and pairing it with correctness-gates-the-metric (a win that drops work the product needed is a regression) closes both ways a fake optimization sneaks in. The attempt ledger with the worked example is the other standout.

Merges clean, suite is green, and it's not duplicating the existing GUARD step. One thing I like beyond the perf angle: the "log reverted attempts so the next agent doesn't re-run them" ledger is a domain-specific version of the cross-session memory idea we're chewing on in #398, same instinct, keep the durable non-obvious finding so nobody rediscovers it. Good to go from me.

asiridalugoda:
@addyosmani thanks for the review. for your #398 i have some suggestions based on a similar skill i derived out of agent-skills, ill comment directly on it.

asiridalugoda:
any update on this team? @federicobartoli @addyosmani

addyosmani:
Merged. Sorry for the lag on this one, it had my sign-off back on the 15th and then sat. Re-ran the full validation suite against the merged tree first since CI hadn't triggered on the branch, all green. Thanks for the sharp write-up, "neutral is a revert" and the attempt ledger are the parts that earn it. And thanks for the detailed data points over on #398, that thread is better for them.

Issue #387, opened 2026-07-10
Resolve skill-validator heuristic limitations before the PR-B test battery locks them in


nucliweb:
Surfaced during review of #379. These are pre-existing behaviors of `scripts/lib/skill-lint.js`, not regressions from the extraction, but they should be triaged as bug-vs-intended before the PR-B test battery codifies the current heuristics as the contract:

- **Frontmatter parsing** is a YAML-like subset: it accepts malformed YAML and non-string descriptions, rejects valid folded/commented YAML, and mismeasures the real length of multiline descriptions.
- **Section checks** use substring matching, so headings inside fenced code blocks or `###` sub-headings can satisfy a required-section rule. Contract question to resolve: `docs/skill-anatomy.md` describes the sections as recommended and allows equivalent headings, while the validator treats the exact headings as required.
- **Cross-reference patterns** miss some reference styles already used in the corpus and can match references inside fenced examples.
- **Trigger check** is satisfied by a negative phrase such as "Do not use when...".

For each: decide whether the current behavior is the intended contract (and document it) or a bug to fix, so PR-B tests assert the decided behavior rather than freezing today's heuristics. Credit to @federicobartoli for surfacing these while reviewing #379.

addyosmani:
Agreed these should be triaged before the #324 test battery freezes the current behavior as the contract, good call pulling them out of #379. The one that actually matters is the section check: `skill-anatomy.md` is explicit that the sections are recommended and equivalent headings are fine, so the validator treating the exact headings as required is stricter than our own stated contract. I'd fix the validator to match the anatomy rather than tighten the anatomy to match the validator, since that flexibility is intentional.

The frontmatter YAML-subset issues and the fenced-code-block false matches are real bugs worth fixing; the negative "Do not use when..." satisfying the trigger check is smaller but also clearly wrong. Want to turn this into a checklist we knock out before #324 lands?

kevglynn:
Happy to take a couple of these once #379 lands (starting against `main`'s validate-skills.js now would just conflict with the extraction). Specifically claiming:

- **Section checks**: move from substring matching to actual heading parsing, so headings inside fenced code blocks and `###` sub-headings stop satisfying required sections — and align with the anatomy contract that equivalent headings are acceptable, per @addyosmani's triage above.
- **Trigger check negation**: stop accepting "Do not use when..." as a trigger phrase.

[...]

PR #405, opened 2026-07-16
fix(validator): reject fenced-block headings and negated triggers

kevglynn:
## Summary

Fixes the two validator heuristic bugs I [claimed on #387](https://github.com/addyosmani/agent-skills/issues/387#issuecomment-4971842600), now that #379 has landed:

**Section check** (`scripts/lib/skill-lint.js`): the old check used `content.includes('## Overview')`, which matches headings inside fenced code blocks (a `## Overview` in a template example would satisfy the real section requirement) and sub-headings (`### Verification` contains the substring `## Verification`). Replaced with a line-anchored regex match against content that has had fenced code blocks stripped via a new `stripFencedCodeBlocks()` helper. Only real `##`-level headings in prose now satisfy the check.

**Trigger check**: a description like "Does stuff. Do not use when testing." satisfied `DESCRIPTION_TRIGGER` because it contains "use … when". Added `DESCRIPTION_TRIGGER_NEGATE` to catch "do not / don't / never use when" and reject descriptions where the only trigger match is a negated form. Descriptions with both a positive and negated trigger still pass (the positive is what matters; the negated describes exclusions, which is fine).

The remaining two items from #387 — frontmatter YAML-subset parsing and cross-reference patterns matching inside fences — are unclaimed and available for follow-up.

## Test plan

- `node scripts/validate-skills.js` → 24 skills, 0 errors, 0 warnings — output is **byte-identical** to the pre-change run (no current skill hits either bug pattern, so these are purely defensive fixes)
- Crafted edge cases exercised against `lintSkillContent` directly:
  - Headings only inside fenced blocks → 5 missing-section errors (previously: 0 — they were silently satisfied)
  - `### Verification` outside a fence → missing `## Verification` error (previously: passed via substring match)
  - Description with only "Do not use when …" → missing-trigger error (previously: passed)
  - Description with both "Use when …" and "Do not use when …" → passes (the positive trigger counts)
- `node scripts/run-evals.js` → 124 checks passed, rank-1 rate 86% — unchanged

Partial fix for #387 (section check + trigger negation).

### DIFF
--- a/scripts/lib/skill-lint.js
+++ b/scripts/lib/skill-lint.js
@@ -34,7 +34,10 @@ const KEBAB_CASE = /^[a-z0-9]+(-[a-z0-9]+)*$/;
 // A description must state WHEN to use the skill, not just what it does
 // (docs/skill-anatomy.md → Required). Accept the canonical "Use when …"
 // plus the equivalent "Use before/after/during …" phrasings in use today.
-const DESCRIPTION_TRIGGER = /\buse (this )?when\b|\buse (before|after|during)\b/i;
+// Reject negated forms ("Do not use when …", "Don't use when …") — those
+// describe exclusions, not trigger conditions.
+const DESCRIPTION_TRIGGER        = /\buse (this )?when\b|\buse (before|after|during)\b/i;
+const DESCRIPTION_TRIGGER_NEGATE = /\b(do not|don't|never) use (this )?(when|before|after|during)\b/i;
 
 // Sections every standard SKILL.md must contain.
 // Each entry is an array of acceptable heading strings — the first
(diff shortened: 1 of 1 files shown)


addyosmani (approved):
Read the actual change, and it's the right fix for both. The section check now strips fenced blocks and anchors to a full line, so `### Verification` inside an example stops satisfying `## Verification` via substring, which was the real hole. The negation logic is careful too: a description with both a real trigger and a "Do not use when" exclusion still passes, and only a solely-negated one fails, which is the correct behavior. It's clean on all 24 current skills, so it tightens the validator without turning anything red, and you left the frontmatter-YAML and cross-ref items for later rather than scope-creeping this. Approving.

PR #408, opened 2026-07-18
feat: add open-code-review-delegate skill

lizhengfeng101:
## Summary

- Adds `skills/open-code-review-delegate/SKILL.md` as an independent skill
- OCR delegation mode: OCR handles deterministic engineering (file selection, rule resolution), the host agent drives the actual review with its own LLM
- No modification to existing skills — complements `code-review-and-quality` for large changesets (10+ files)

## Motivation

Addresses the coverage gap identified in #327 discussion with a different approach:
- Not an "integrations pointer" in an existing skill (rejected concern)
- Not a product usage manual (intelligence is in the agent, not ocr)
- Follows the same pattern as `browser-testing-with-devtools` — external tool provides a capability the agent otherwise lacks (reliable file-level coverage for large changesets)

## Key differences from #327

| Original concern | This PR |
|---|---|
| "Third-party product manual" | Agent drives review; ocr is infrastructure (like git) |
| "Links in core skill files" | Independent skill, no changes to code-review-and-quality |
| "Vendor LLM dependency" | Delegation mode is LLM-free on ocr side |
| "Maintenance coupling" | CLI surface is 2 commands only: `delegate preview` and `delegate rule` |

## Test plan

- [ ] Frontmatter validates (name + description only, single-line string)
- [ ] All required sections present (Overview, When to Use, Process, Common Rationalizations, Red Flags, Verification)
- [ ] No modifications to other skill files

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,7 @@ The commands above are entry points. The pack includes 24 skills total — 23 li
 | Skill | What It Does | Use When |
 |-------|-------------|----------|
 | [code-review-and-quality](skills/code-review-and-quality/SKILL.md) | Five-axis review, change sizing (~100 lines), severity labels (Nit/Optional/FYI), review speed norms, splitting strategies | Before merging any change |
+| [open-code-review-delegate](skills/open-code-review-delegate/SKILL.md) | Delegation-mode review using OCR for deterministic file selection and rule resolution while the agent drives the actual review | Reviewing PRs with 10+ changed files where prompt-only review risks skipping files |
 | [code-simplification](skills/code-simplification/SKILL.md) | Chesterton's Fence, Rule of 500, reduce complexity while preserving exact behavior | Code works but is harder to read or maintain than it should be |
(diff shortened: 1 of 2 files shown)


federicobartoli:
I'll be direct rather than string this along. The redesign moves the integration point but not the question already settled in #327: whether this pack carries a first-class skill for one vendor's product. Delegate mode still means the catalog documents your CLI (npm install -g @alibaba-group/open-code-review as a prerequisite, two subcommands, eight shared flags) and inherits every change to that surface. Addy was explicit in #327 that he keeps the pack's skill files clear of specific third-party products, because one pointer invites the next, and a full catalog skill is more of that, not less.

So from my side this is a pass on the same grounds as #327, and I don't think another restructuring changes it. The right home for this is your own repo's plugin and skill distribution, which already exists. Catalog additions are Addy's ( @addyosmani ) call, so I'll leave the formal close to him.

addyosmani:
Closing this one. It's a skill built around delegating to a specific external tool (Open Code Review), which is the vendor-specific direction I keep the core pack clear of, the same line as the other tool-integration skills. The lighter "See Also" pointer in #327 is about as far as I want a third-party review tool referenced from inside a skill, and even that is still open rather than merged; a standalone delegate skill is a step past it. Genuinely appreciate the work, it would just live better as your own skill than in the core repo.

PR #410, opened 2026-07-18
feat: add Test Planner handoff to TDD

summergan:
### TDD Test Planner Handoff

Teaches the existing test-driven-development skill to consume a reviewed READY Test Planner before writing any code:

- Validates the planner's oracle, READY status, and selected case against current requirements.
- Preserves the planned case meaning; test code translates but does not change the oracle, layer, or expected result.
- Blocks unresolved requirements instead of encoding a guess.
- For a single obvious bug with no supplied planner, records a visible mini planner checkpoint before the first file edit.
- Preserves RED-GREEN execution evidence with the discovered project command.

### Scope

7 files, +140/-6. Only modifies test-driven-development and its own evals/fixtures. Preserves the existing 80/15/5 Test Pyramid. No dependency on unmerged skills.

### Validation

- validate-skills: 24/24 passed
- run-evals: 125 checks passed, 0 warnings
- validate-commands: 8/8 passed
- Behavioral dry-run: 2 fixture-backed evals planned
- Fixture npm test: passed

### DIFF
--- a/evals/cases/test-driven-development.json
+++ b/evals/cases/test-driven-development.json
@@ -13,6 +13,10 @@
       {
         "prompt": "What tests should cover this new parsing logic before I write it?",
         "top_k": 3
+      },
+      {
+        "prompt": "The invoice rounding Test Planner is reviewed and READY; implement its case test-first without changing the oracle.",
+        "top_k": 3
       }
     ],
     "negative": [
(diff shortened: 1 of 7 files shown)


addyosmani:
This is the one I'd push back on hardest. Adding a "Planner Handoff Before RED" gate to test-driven-development, including a mandatory "Mini Test Planner" block before the first edit even for a single obvious bug, adds real ceremony to the pack's leanest and most-used skill.

TDD earns its keep on the tight red-green-refactor loop; requiring an oracle/READY/status ritual upstream of every RED works against that, and it couples a core skill to the new test-case-design methodology by default. If test-case-design lands, the handoff belongs as a light optional pointer ("if a reviewed planner exists, follow it"), not a required gate that reshapes TDD for everyone. I'd keep TDD lean and let the planner be opt-in.

addyosmani:
@nucliweb @federicobartoli I'd value a second opinion here, because my objection is a judgment call rather than a defect, and I don't want to reshape summergan's work on my instinct alone.

This adds a Planner Handoff gate to `test-driven-development`, including a mandatory "Mini Test Planner" block that must be written before the first edit even for a single obvious bug. My read is that it over-processes the leanest and most-used skill in the pack, and couples it to the new test-case-design methodology by default, so I argued the handoff should be an optional pointer ("if a reviewed planner exists, follow it") rather than a required gate. If you both think the ceremony earns its keep, I'll reconsider. Worth noting it needs a rebase now that #419 has landed on the same file.

PR #409, opened 2026-07-18
feat: add test case design skill

summergan:
### Test Case Design

- Separates Test Analysis, Test Design, and Case Specification.
- Builds a Test Planner from the test basis/oracle, changed surfaces, affected contracts, risks, quality concerns, and exit criteria.
- Selects the smallest sufficient layer across Unit, Component, Frontend Integration, API, Contract, Backend Integration, and E2E/System testing.
- Distinguishes backend API behavior from consumer/provider contract compatibility.
- Applies boundary values, equivalence partitions, decision tables, state transitions, pairwise reduction, and regression reproduction.
- Produces concrete Case Specifications with names, objectives, preconditions, test data, actions, expected results/oracles, coverage traceability, priority, and readiness status.
- Retains the familiar 80/15/5 pyramid as a cost guideline rather than a universal quota.
- Prevents disputed, incomplete, or unsupported assumptions from becoming executable tests.
- Hands READY cases to TDD without allowing implementation to silently change their layer, scope, or oracle.

### Scope

6 files, +935/-4. New skill only; no edits to existing skills, commands, agents, or unrelated docs.

### Validation

- validate-skills: 25/25 passed
- run-evals: 132 checks passed, 0 warnings
- validate-commands: 8/8 passed
- Behavioral dry-run: 8 fixture-backed evals planned

Note: If the standalone `tests` skill PR also lands, update README skill count from 25 to 26.

PR #411, opened 2026-07-18
feat: add standalone tests execution skill

summergan:
Discover and run project-owned test suites with bounded evidence while keeping test authoring and failure diagnosis out of scope.### Tests Skill

Discovers and executes existing project-owned test commands with bounded, non-interactive execution and exact per-command evidence:

- Three bounded modes: Reproduce failure, Verify change, Release gate.
- Discovers real commands from project configuration; never invents script names.
- Keeps API, Contract, Frontend Integration, Backend Integration, and E2E distinct.
- Classifies results honestly: PASS, FAIL, NO_TESTS, NOT_RUN.
- Redacts secrets from evidence; never diagnoses or fixes failures.
- Consumes an available Test Planner as read-only scope input without requiring one.

### Scope

7 files, +415/-4. New skill only; no edits to existing skills, commands, agents, or unrelated docs.

### Validation

- validate-skills: 25/25 passed
- run-evals: 132 checks passed, 0 warnings
- validate-commands: 8/8 passed
- Behavioral dry-run: 5 fixture-backed evals planned
- Fixture smoke tests confirmed pass, failure, zero-test, and missing-prerequisite behavior

Note: If the test-case-design PR also lands, update README skill count from 25 to 26.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Skills also activate automatically based on what you're doing — designing an A
 **Fastest path — any agent, one command.** The open [skills CLI](https://github.com/vercel-labs/skills) installs into 70+ agents (Claude Code, Cursor, Codex, Copilot, Cline, and more):
 
 ```bash
-npx skills add addyosmani/agent-skills            # install all 24 skills
+npx skills add addyosmani/agent-skills            # install all 25 skills
 npx skills add addyosmani/agent-skills --list     # browse before installing
 ```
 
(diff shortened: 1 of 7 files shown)


addyosmani:
Same catalog-growth hold as #409, plus a specific scope question here: "discover and run the project's existing tests, authoring and diagnosis out of scope" is a fairly thin slice, and it overlaps both TDD's verification step and what the `/test` command already does. Before it's a standalone skill I'd want the case for what it does that those don't, rather than a third place that talks about running tests. Not a no, but the boundary needs to be sharper than "runs tests but doesn't write or debug them."

PR #414, opened 2026-07-18
docs(documentation): add Runbooks section to documentation-and-adrs

HMAKT99:
## What this adds

A `## Runbooks` section in `skills/documentation-and-adrs/SKILL.md`, covering:

- **When to write a runbook** — any alert that fires, any non-obvious recovery path, any escalation path that shouldn't be memorised
- **The three-line minimum** — what it means / first thing to check / who to escalate to (already referenced in `observability-and-instrumentation` as the required minimum for alert links)
- **A full runbook template** stored in `docs/runbooks/`, with trigger, likely causes, immediate actions, escalation, and related links
- **Keep-current discipline** — update the runbook as part of closing every incident it was used in

Plus two rationalization rows, two red flags, and two verification checkboxes in the existing sections.

## The gap it fills

The `observability-and-instrumentation` skill already says: *"It links to a runbook — even three lines: what it means, first query to run, escalation path."* That's a hard requirement on every alert. But `documentation-and-adrs` — the skill that teaches what documentation to write and how — has no guidance on what a runbook is or how to write one. Engineers are told to link to a runbook without being taught the artifact.

## Why here

Runbooks are a documentation artifact, not an observability artifact. The observability skill is the right place to say "link to one"; `documentation-and-adrs` is the right place to define what it contains, where to store it, and how to keep it current.

## Non-duplication check

`observability-and-instrumentation` mentions runbooks as a required alert field but teaches nothing about their content. `debugging-and-error-recovery` (PR #312) covers the incident response *process* — declaring severity, stabilising, postmortem. This section is the *documentation artifact* that enables that process: a pre-written recovery guide rather than a live triage workflow. The two cross-link naturally but teach different things.

## Checklist

- [x] Fully additive — no edits to existing prose, only new content and new rows/bullets in existing lists
- [x] Cross-references resolve to existing skills (`observability-and-instrumentation`, `shipping-and-launch`)
- [x] Touches only `skills/documentation-and-adrs/SKILL.md`
- [x] Rationalization rows, red flags, and verification checkboxes match host skill's style

### DIFF
--- a/skills/documentation-and-adrs/SKILL.md
+++ b/skills/documentation-and-adrs/SKILL.md
@@ -256,6 +256,79 @@ Special consideration for AI agent context:
 - **ADRs** — Help agents understand why past decisions were made (prevents re-deciding)
 - **Inline gotchas** — Prevent agents from falling into known traps
 
+## Runbooks
+
+Runbooks are operational documentation: step-by-step guides for known failure modes and recurring on-call tasks. Where an ADR records *why* a decision was made, a runbook records *what to do when something goes wrong*. They exist so an engineer can act at 3am without reconstructing knowledge from first principles.
+
+The `observability-and-instrumentation` skill requires every alert to link to a runbook. This is where you write it.
+
+### When to Write a Runbook
+
+- Any alert that has fired, or is expected to fire, in production
+- Any recurring operational task with a non-obvious sequence of steps
+- Any failure mode whose recovery path isn't self-evident from the alert text
+- Any escalation path that engineers shouldn't have to remember under pressure
+
+### The Three-Line Minimum
+
+A runbook doesn't have to be comprehensive to be useful. The minimum viable runbook answers three questions:
+
+1. **What does this mean?** — the likely cause in plain language
+2. **What is the first thing to check?** — one command or dashboard link
+3. **Who to escalate to if that doesn't resolve it?** — team, channel, or rotation
+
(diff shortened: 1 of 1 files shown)


addyosmani:
This one I'd redirect rather than take. observability-and-instrumentation already owns runbooks: it requires every alert to link one and specifies the same three-line minimum (what it means, first query to run, escalation path) that this PR adds. So a Runbooks section in documentation-and-adrs both duplicates that and sits in the wrong skill, ADRs are decision records, runbooks are operational incident docs tied to alerts. If runbooks deserve more depth, expand them in observability where they already live rather than forking a second copy into the ADR skill.

HMAKT99:
Fair redirect — I see the issue now. The three-line minimum is already anchored in `observability-and-instrumentation` as an alert requirement; putting a fuller version in `documentation-and-adrs` creates a second copy in the wrong conceptual home. I'll close this and expand the runbook guidance where it already lives in the observability skill instead.

HMAKT99:
Closing in favour of #422 which expands the runbook guidance where it already lives in `observability-and-instrumentation`, per the redirect above.

addyosmani:
Superseded by #422, which puts the runbook guidance in observability-and-instrumentation where the alerting rules already require it, and does it in 17 lines instead of 79. That's the right outcome, so closing this one. Thanks for taking the redirect rather than arguing it.

PR #415, opened 2026-07-18
docs(shipping): add Error Budget Release Gate section

HMAKT99:
## What this adds

A `## SLOs and Error Budgets` section in `skills/shipping-and-launch/SKILL.md`, covering:

- **The three concepts** — SLI (what you measure), SLO (the target), error budget (the allowed failure)
- **Budget-gates-releases policy** — a clear four-state decision table: ship normally → slow rollouts → freeze feature work → resume after recovery
- **Setting targets from measurement** — instrument first, set the SLO at your current p75, tighten incrementally; with a specific warning about the four-nines trap
- **Burn rate alerting** — how to catch slow degradations that never breach the SLO threshold individually but drain the budget silently
- Cross-link to the rollout thresholds table already in the skill (burn rate connects directly to the hold/rollback signals)

Plus two rationalization rows, three red flags, and three verification checkboxes.

## The gap it fills

The skill already references SLOs — the Staged Rollout section says thresholds should be "justified by the SLO or by historical data." But the skill never defines what an SLO is, how to set one, or what happens when the error budget runs out. Engineers are told to justify thresholds against an SLO without being taught the framework that produces one.

## Non-duplication check

`observability-and-instrumentation` teaches how to instrument for visibility. This section teaches the *target* that gives instrumentation its meaning — what "good" looks like and how much failure is acceptable before feature work stops. The skills are complementary, not overlapping.

## Checklist

- [x] Touches only `skills/shipping-and-launch/SKILL.md`
- [x] Rationalization rows, red flags, and verification checkboxes match host skill's style

### DIFF
--- a/skills/shipping-and-launch/SKILL.md
+++ b/skills/shipping-and-launch/SKILL.md
@@ -237,7 +237,7 @@ In the first hour after launch:
 
 ## Error Budget Release Gate
 
-Your service's error budget (defined by its SLO in `observability-and-instrumentation`) determines whether it's safe to ship. Use it as an objective gate — not a negotiation:
+Your service's error budget — the fraction of requests or time your SLO allows to fail — determines whether it's safe to ship. Use it as an objective gate — not a negotiation:
 
 ```
 Budget remaining > 20%  →  Ship normally; monitor closely
@@ -282,6 +282,7 @@ Every deployment needs a rollback plan before it happens:
 - For security pre-launch checks, see `../../references/security-checklist.md`
 - For performance pre-launch checklist, see `../../references/performance-checklist.md`
 - For accessibility verification before launch, see `../../references/accessibility-checklist.md`
+- For the alerting rules and SLO-tied thresholds, see `observability-and-instrumentation`
 
 ## Common Rationalizations
 
@@ -326,4 +327,4 @@ After deploying:
 
 For every shipped service:
 
-- [ ] Error budget policy in place: what happens when budget drops below 20% and when it's exhausted (see `observability-and-instrumentation` for SLO/SLI setup)
+- [ ] Error budget policy in place: know what action to take when budget drops below 20% and when it's exhausted
--- a/skills/shipping-and-launch/SKILL.md
+++ b/skills/shipping-and-launch/SKILL.md
@@ -235,36 +235,9 @@ In the first hour after launch:
 6. Confirm rollback mechanism works (dry run if possible)
 ```
 
-## SLOs and Error Budgets
+## Error Budget Release Gate
 
-Monitoring tells you what is happening. SLOs tell you what *should* be happening and how much failure you can afford. Without them, alert thresholds are guesses and "is this good enough to ship?" has no shared answer.
-
-### The Three Concepts
-
-**SLI (Service Level Indicator):** The actual measurement — what you observe.
-
-```
-Error rate    = failed requests / total requests
-Latency       = P95 response time over a rolling window
-Availability  = successful health checks / total checks
-```
-
-**SLO (Service Level Objective):** The target for the SLI — the promise you make to users.
-
-```
-99.9% of API requests succeed in < 500ms, measured over a 30-day window
-```
-
-**Error budget:** The allowed failure implied by the SLO.
-
-```
-100% − 99.9% = 0.1% of requests may fail
-Over 30 days (43,200 minutes): 43 minutes of full outage, or proportionally more partial failure
-```
-
-### Error Budgets Gate Releases
-
-The error budget is the mechanism that balances feature velocity against reliability:
+Your service's error budget (defined by its SLO in `observability-and-instrumentation`) determines whether it's safe to ship. Use it as an objective gate — not a negotiation:
 
 ```
 Budget remaining > 20%  →  Ship normally; monitor closely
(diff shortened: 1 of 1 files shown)


addyosmani:
Good content, but the placement is off. observability-and-instrumentation already works in SLO terms (its alerting section justifies thresholds "by the SLO"), so defining SLOs, error budgets, and burn-rate alerts in shipping-and-launch splits the concept across two skills. The one piece that's genuinely shipping's is "the error budget gates the release", that decision belongs here. I'd move the SLO, budget, and burn-rate definitions into observability (or just reference them) and keep only the release-gate call in shipping, rather than redefining the concept in a second place.

HMAKT99:
Agreed on the placement. The SLI/SLO/error-budget definitions belong in `observability-and-instrumentation` where the alerting threshold language already uses them. I'll trim this down to just the release-gate decision — the four-state policy (ship / slow rollouts / freeze / resume) that answers 'can we ship?' — and cross-reference observability for the definitions. That's the piece genuinely owned by shipping.

addyosmani:
Agreed, and the four-state ship / slow rollouts / freeze / resume policy is precisely the piece shipping owns. Trim to that plus a cross-reference to observability for the definitions and it's a clean fit.

HMAKT99:
The trim is already in — pushed the cut at the same time I replied earlier. The section is now just the four-state policy table plus a cross-reference to `observability-and-instrumentation` for the SLO/SLI definitions. Checker passes 0/0. Let me know if anything else needs adjusting.

addyosmani (approved):
This is a real improvement on the earlier cut, and it lands the concern I had before. Last time the SLO material read like it was setting up definitions that belong in `observability-and-instrumentation`. This version doesn't do that: it defers the SLO/SLI setup to that skill and keeps only the part that is genuinely a shipping decision, which is the error budget as a release gate. That is the right split. The budget-remaining table maps cleanly to "should I ship right now," and wiring burn rate into the existing rollout thresholds as a hold signal is a nice touch because it reuses machinery the skill already has rather than inventing a parallel one. Clean, green, additive. Approving.

federicobartoli (changes requested):
Thanks @HMAKT99 — the trimmed version is a clean fit: the release-gate table, the burn-rate note, and the rationalization/red-flag/verification additions all match the file's existing structure, and the "hold" signal claim lines up with the existing Rollout Decision Thresholds table.

Blocking: both new cross-references in skills/shipping-and-launch/SKILL.md ("defined by its SLO in observability-and-instrumentation", and the checklist's "see observability-and-instrumentation for SLO/SLI setup") point at content that doesn't exist yet — that skill currently has no SLO/SLI/error-budget definitions (just one passing "justified by the SLO" under Alerting) and no burn-rate coverage. Please land the definitions there first (companion PR), or soften these two pointers until it does.
Nit: the PR title still says "add SLOs and Error Budgets section", but the section is now "Error Budget Release Gate" — worth updating before squash-merge.
Nit: consider also listing the observability pointer under this file's existing "See Also" section, where its cross-references live.

@addyosmani looping you in since the approval note says this "defers the SLO/SLI setup to that skill" — I checked and that content isn't in observability-and-instrumentation yet (one passing "justified by the SLO" under Alerting, same single mention in references/observability-checklist.md; no SLI/error-budget/burn-rate coverage). Am I missing something? If not, this actually matches your Jul 21 point: the definitions were cut from shipping but never landed in observability. I'd suggest a companion PR there, or softening the two pointers here in the meantime.

HMAKT99:
Good catch, @federicobartoli — you're right that the cross-references were overpromising. Fixed in the latest push:

- **Blocking (pointer 1):** Replaced `(defined by its SLO in `observability-and-instrumentation`)` with a self-contained definition — `the fraction of requests or time your SLO allows to fail`. No longer implies definitions live in observability.
- **Blocking (pointer 2):** Removed `(see `observability-and-instrumentation` for SLO/SLI setup)` from the checklist item — it now stands alone.
- **Nit (See Also):** Added `observability-and-instrumentation` to the See Also section — that's the right home for a cross-reference, not buried in body copy.
- **Nit (title):** Updated PR title to `docs(shipping): add Error Budget Release Gate section`.

Validator 0/0.

federicobartoli:
All four points from my July review are addressed :both pointers, the title, and the See Also placement. Thanks for the thorough pass.

One residual slipped through in the move, though: the new See Also line carries the same "SLO setup" phrasing that the latest push scrubbed from the body, and that content still doesn't exist in observability-and-instrumentation. Could you reword it to what's actually there today:

"- For the alerting rules and SLO-tied thresholds, see `observability-and-instrumentation`"

And restating the other half of my July either/or, since it's still open: the SLO/SLI/error-budget definitions you trimmed from the first cut are exactly what observability is missing, and Addy already pointed at that home. If you ever want to bring them back as the companion PR there, this See Also can grow back to cover it once it lands.

With the one-line reword I'll flip my review to approve.

HMAKT99:
Thanks @federicobartoli — the one-line reword is in the latest push. The See Also entry now reads exactly as you suggested: *"For the alerting rules and SLO-tied thresholds, see `observability-and-instrumentation`"* — no forward promise about definitions that don't exist yet. Validator 0/0.

federicobartoli (approved):
The See Also line now reads as agreed and no longer promises SLO definitions that observability does not have yet. Everything else from the July round is still in place. Approving. The companion PR bringing the SLO, SLI and error-budget definitions into observability-and-instrumentation is still a good follow-up if you feel like it.

PR #416, opened 2026-07-18
docs(tdd): add Flaky Test Management section

HMAKT99:
## What this adds

A `## Flaky Test Management` section in `skills/test-driven-development/SKILL.md`, covering:

- **Common causes table** — timing/async, shared state, external dependencies, date sensitivity, random data, test ordering — with a concrete fix for each
- **The quarantine workflow** — move to a non-blocking suite, file a fix-or-delete ticket with a ≤ 2-week deadline, then fix or delete; never let quarantine become permanent
- **Prevention rules** — prefer small tests, freeze non-determinism at the boundary, assert on events not timing, run tests in random order in CI

Plus two rationalization rows (including the "add a sleep" anti-pattern), two red flags (auto-retry as standard workflow; orphaned `skip`/`xit` annotations), and two verification checkboxes.

## The gap it fills

The skill already lists "Flaky tests (timing, order-dependent)" in its anti-patterns table and says to "use deterministic assertions, isolate test state" — but gives no process for what to do when a flaky test appears in a live CI pipeline. The quarantine pattern, the fix-or-delete deadline, and the root-cause table are entirely absent. Engineers know flaky tests are bad; they don't have a workflow for dealing with them.

## Non-duplication check

`ci-cd-and-automation` covers pipeline configuration. This section is about managing test reliability as a property of the test suite itself — a `test-driven-development` concern, not a CI configuration concern.

## Checklist

- [x] Touches only `skills/test-driven-development/SKILL.md`
- [x] Style matches host skill (table format for causes, workflow diagram, bullet prevention rules)

### DIFF
--- a/skills/test-driven-development/SKILL.md
+++ b/skills/test-driven-development/SKILL.md
@@ -360,6 +360,47 @@ This separation ensures the test is written without knowledge of the fix, making
 
 For JavaScript/TypeScript testing patterns illustrating these principles — Jest, React Testing Library, Supertest, Playwright — see `references/testing-patterns.md`. The principles transfer to any ecosystem; the syntax and tools there are JS/TS-specific.
 
+## Flaky Test Management
+
+A flaky test is one that passes and fails on the same code without any change — non-deterministic by nature. Flakiness isn't a minor inconvenience: it trains engineers to ignore red CI runs, which is how real failures get missed.
+
+### Common Causes
+
+| Cause | Example | Fix |
+|---|---|---|
+| Timing / async | `setTimeout` assumed to fire before assertion | Await the actual event or state change; never sleep |
+| Shared state | Test A leaves data that breaks test B | Each test creates and tears down its own state |
+| External dependency | Real API call succeeds sometimes | Mock or stub at the boundary |
+| Date/time sensitivity | Test assumes current year | Inject or freeze the clock |
+| Random data | Auto-generated values that occasionally collide | Use deterministic seeds or fixed fixtures |
+| Test ordering | Tests pass individually but fail together | Enforce random order in CI to surface hidden coupling |
+
+### The Quarantine Workflow
+
+Don't let a flaky test block CI indefinitely — but don't ignore it either:
+
+```
+Flaky test detected
+       │
+       ▼
+  Move to quarantine suite (separate CI job, non-blocking)
+       │
+       ▼
+  File a fix-or-delete ticket with a deadline (≤ 2 weeks)
+       │
+       ├── Root cause found → fix and restore to main suite
+       └── No fix in 2 weeks → delete the test
+```
+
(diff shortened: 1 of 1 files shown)


addyosmani:
Flaky-test quarantine is a reasonable fit for TDD and +47 is proportionate. Two checks before it lands: it's the third open PR touching this SKILL.md (with #419 and #410), so it'll need a rebase, and make sure the quarantine and prevention guidance doesn't restate what the skill's existing test-level and isolation material already says. If it's additive rather than overlapping, I'm fine with it.

HMAKT99:
Thanks for the check. I've looked at #419 and #410 to make sure the flaky-test content is additive rather than overlapping — the existing TDD skill covers isolation as a test-writing principle (each test sets up/tears down its own state) but has nothing on the operational workflow once flakiness appears in CI: the quarantine step, the fix-or-delete deadline, or the root-cause table. The content is additive. I'll rebase now.

addyosmani:
Good, that's exactly the distinction I wanted: isolation as a test-writing principle is already there, the operational quarantine-and-deadline workflow isn't. Additive then. Rebase and I'll take another look.

HMAKT99:
Rebased onto current main — one commit, sole contributor, 0/0 validation. Ready when you are.

addyosmani (approved):
Good section and it belongs here. Flaky tests are a test-discipline problem, and this is opinionated in the right places: quarantine is a ticking clock not a home, fix-or-delete with a deadline, no sleep-based fixes, random order in CI to surface coupling. The causes table is accurate and the rationalizations match the skill's voice. Test-merged clean, validators pass.

One thing to keep an eye on, not blocking this: TDD's SKILL.md is at ~445 lines after this, closing on the 500-line soft ceiling in skill-anatomy. This section earns its place, but the next substantial addition to this skill should probably go to `references/` rather than inline so we don't blow the budget. Approving.

skills/test-driven-development/SKILL.md
```diff
@@ -383,6 +426,8 @@ For JavaScript/TypeScript testing patterns illustrating these principles — Jes
 - Test names that don't describe the expected behavior
 - Skipping tests to make the suite pass
 - Running the same test command twice in a row without any intervening code change
+- CI configured to auto-retry on failure as standard workflow (masking flakiness rather than fixing it)
+- Tests annotated `skip` or `xit` with no owner, no ticket, and no deadline
```
federicobartoli, line 430: One line to tighten before I approve. This file is consumed by models, and they pattern-match literally: in a pytest or JUnit project, an agent told to look for xit will grep for xit, miss @pytest.mark.skip or @ Disabled, and report a clean bill. Naming the behavior instead of the JS token keeps the red flag working in every ecosystem, in line with the #419 neutrality pass:

```suggestion
- Disabled tests (`skip`, `xit`, or the framework's equivalent) with no owner, no ticket, and no deadline
```
With this in, I'm approving; everything else checked out, including the additive claim across the whole catalog and the 445-line count.

federicobartoli (changes requested):

PR #417, opened 2026-07-18
docs(api): add Webhook Design section to api-and-interface-design

HMAKT99:
## What this adds

A `## Webhook Design` section in `skills/api-and-interface-design/SKILL.md`, covering the outbound side of API design — pushing events to consumer-provided URLs:

- **Event envelope** — consistent structure with `id`, `type`, `timestamp`, `version`, and `data`; emphasis on the stable `id` field as the idempotency key consumers need to de-duplicate retries
- **Signature verification** — HMAC-SHA256 signing with `timingSafeEqual` comparison, with the explicit guidance to design it in from day one (adding required auth headers after consumers are live forces a coordinated migration)
- **Reliable delivery** — queue-backed async fan-out, exponential backoff retry sequence, dead-letter queue for exhausted attempts; why at-least-once means consumers must be idempotent
- **Consumer contract** — the four rules every webhook consumer must implement: verify signature, ack immediately, de-duplicate on `id`, ignore unknown event types

Plus two rationalization rows, three red flags, and three verification checkboxes.

## The gap it fills

The skill covers inbound REST and GraphQL thoroughly — resource design, pagination, error semantics, validation at boundaries. It has nothing on the outbound side: webhooks are a distinct design problem (push vs pull, at-least-once delivery, consumer isolation, signature auth) that the skill name implies but doesn't teach.

## Non-duplication check

`security-and-hardening` covers authentication patterns for inbound requests. The signature verification here is outbound-specific (signing what you send, not validating what you receive) and is framed as an API design decision, not a security hardening step.

## Checklist

- [x] Touches only `skills/api-and-interface-design/SKILL.md`
- [x] Code examples use the same TypeScript style as the host skill

### DIFF
--- a/skills/api-and-interface-design/SKILL.md
+++ b/skills/api-and-interface-design/SKILL.md
@@ -208,97 +208,32 @@ PATCH /api/tasks/123
 
 ## Webhook Design
 
-Webhooks are the outbound side of an API: instead of consumers polling for changes, you push events to a URL they provide. The design challenges are different from inbound REST — delivery is inherently unreliable, consumers must be idempotent, and a slow consumer should never block others.
+Webhooks are the outbound side of an API: instead of consumers polling for changes, you push events to a URL they provide. Four decisions determine whether your webhook API is reliable and safe to build on:
 
-### Event Envelope
-
-Every webhook event should follow a consistent envelope so consumers can handle all events with the same code:
+**1. Give every event a stable, unique ID.** Delivery is at-least-once — retries happen. Without a stable `id` field, consumers have no way to de-duplicate:
 
 ```typescript
 interface WebhookEvent {
-  id: string;           // Unique, stable event ID — consumers use this for idempotency
-  type: string;         // e.g. "task.created", "task.completed"
-  timestamp: string;    // ISO 8601
-  version: string;      // Payload schema version — lets you evolve the shape
-  data: unknown;        // Event-specific payload
-}
-
-// Example
-{
-  "id": "evt_01HX9K3MNPQ7RSTUVWXYZ",
-  "type": "task.completed",
-  "timestamp": "2025-06-12T14:32:00Z",
-  "version": "1",
-  "data": { "taskId": "task_abc", "completedBy": "user_123" }
+  id: string;        // Stable across retries — consumers key idempotency on this
+  type: string;      // e.g. "task.completed"
+  timestamp: string; // ISO 8601
+  version: string;   // Schema version — lets you evolve the payload shape
+  data: unknown;
 }
 ```
 
-The `id` field is the most important part: without a stable event ID, consumers cannot de-duplicate retries.
-
-### Signature Verification
-
-Sign every outbound payload so consumers can verify it came from you and wasn't tampered with:
+**2. Sign every outbound payload.** Use HMAC-SHA256; attach as a header; document consumer verification. Design this in from day one — adding required auth after consumers are live forces a coordinated migration:
 
 ```typescript
-import { createHmac } from 'node:crypto';
+import { createHmac, timingSafeEqual } from 'node:crypto';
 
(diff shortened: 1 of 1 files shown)


addyosmani:
Webhooks genuinely fit here, and the decomposition is right (envelope, signature verification, delivery/retries, consumer contract). api-and-interface-design is 294 lines so +102 lands it under the size budget. My one push is length: +102 for a single section is heavy, so make sure every line is API-design guidance an agent will act on, not general webhook background it could get anywhere. If it's tight, it's a real gap filled.

HMAKT99:
Good call — some of the delivery section reads as webhook background rather than API design decisions an agent acts on. I'll cut it down to the decisions: envelope shape (stable event ID for idempotency), signature verification (design it in from day one, not after consumers are live), async fan-out (never in the request path), and the consumer contract. The retry mechanics detail will come out.

addyosmani:
That's the right cut. Envelope shape, signature verification from day one, async fan-out, and the consumer contract are the decisions an agent acts on; retry mechanics is the background. Ping me when it's trimmed.

HMAKT99:
Trimmed and pushed — four decisions only (envelope ID, HMAC-SHA256 signature, async fan-out, consumer contract), +35 lines net. The delivery mechanics background and retry sub-section are out. Ready for your re-review.

addyosmani (approved):
This one belongs. Webhooks are the outbound half of an API surface and the skill genuinely didn't cover them, so this fills a real gap rather than padding. The four rules are the right four, in the right order: stable event id for idempotency, sign from day one, deliver off the request path, publish the consumer contract. The rationalizations and red flags you added match the skill's existing voice, and the HMAC snippet is correct (timingSafeEqual, raw body). Test-merged clean, validators pass.

Small thing, take it or leave it: the signing snippet shows the produce side but names `timingSafeEqual` in the import without showing the verify-side compare, so a reader has to infer it. A one-liner on the consumer end would close the loop. Not blocking. Approving.

federicobartoli (changes requested):
The four decisions are the right four and the trim made this section dense in the good sense. Two small asks before I approve, both building on Addy's note about the unused `timingSafeEqual` import:

1. Could you add the consumer-side compare? Agents reading this will be asked to write the verify side, which is where the classic mistakes live (string `===` compares, and `timingSafeEqual` throwing on length mismatch, which turns a malformed attacker header into a crash inside the security check):

```typescript
// Verify (consumer side): length-check first, timingSafeEqual throws on length mismatch
const expected = Buffer.from('sha256=' + createHmac('sha256', secret).update(rawBody).digest('hex'));
const received = Buffer.from(signatureHeader);
const valid = expected.length === received.length && timingSafeEqual(expected, received);
```

2. Could you add the illustration hedge the catalog uses elsewhere? Something like: "The snippets use TypeScript and Node's crypto for illustration; the rules (sign the raw body, constant-time compare, length-check first) are the same on any platform."

With those two in I'll approve. Optional, take it or leave it: one clause in the consumer contract about rejecting deliveries older than a tolerance window, since the envelope already carries `timestamp`.

HMAKT99:
Both points addressed in the latest push, @federicobartoli:

1. **Consumer-side verify** — added the snippet showing length-check first (so `timingSafeEqual` can't throw on a malformed attacker header) then constant-time compare. The two classic mistakes are now explicitly guarded.
2. **Illustration hedge** — added a one-liner before the code block: *"The snippets below use TypeScript and Node's `crypto` for illustration; the rules (sign the raw body, constant-time compare, length-check first) are the same on any platform."*

On the optional timestamp clause: the envelope already carries `timestamp`, and a consumer contract note about rejecting stale deliveries is a solid addition for a real production setup. Happy to include it — or leave it for a follow-on edit if you'd rather keep this PR focused. Your call.

Validator 0/0.

Issue #404, opened 2026-07-15
Proposal: make test-driven-development ecosystem-neutral with practitioner-validated references


federicobartoli:
## Context

This proposal was prompted by a public LinkedIn discussion with Renan França.

Renan noticed that, although `test-driven-development` is described as a
general development workflow, its examples and supporting material appear
specific to JavaScript and TypeScript. He asked whether the workflow had been
used and validated in a Java project.

Java is the context that surfaced the issue, but this is not intended to be a
Java-specific proposal. The broader question is whether the skill currently
represents TDD across software ecosystems or unintentionally assumes an
npm-based project.

LinkedIn discussion: [https://www.linkedin.com/feed/update/urn:li:activity:7482684721563508736/?dashCommentUrn=urn%3Ali%3Afsd_comment%3A%287482933051820498944%2Curn%3Ali%3Aactivity%3A7482684721563508736%29](https://www.linkedin.com/feed/update/urn:li:activity:7482684721563508736/?dashCommentUrn=urn%3Ali%3Afsd_comment%3A%287482933051820498944%2Curn%3Ali%3Aactivity%3A7482684721563508736%29)
## Current mismatch

The core TDD principles are ecosystem-independent, but the current execution
guidance is not:

- the RED, GREEN, and bug-reproduction examples use TypeScript;
- the final verification checklist hardcodes `npm test`;
- `references/testing-patterns.md` uses Jest, React Testing Library,
  Supertest, and Playwright;
- the reference describes itself as covering patterns “across the stack,”
  although it currently represents one ecosystem;
- the skill does not first require agents to discover the repository's
  language, build system, test runner, wrappers, and existing conventions.

This may cause agents working in JVM, .NET, Python, Go, Rust, Ruby, PHP, and
other projects to receive irrelevant commands or testing patterns.

### 1. Keep one TDD skill

Keep the universal workflow in `test-driven-development/SKILL.md`:

- RED → GREEN → REFACTOR
- reproduce bugs with a failing test
- test observable behavior rather than implementation details
- choose an appropriate test level
- prefer deterministic and isolated tests
- run focused tests during development and the relevant suite before completion

Creating one TDD skill per language would duplicate the workflow, create
overlapping triggers, and allow the different versions to drift.

### 2. Make the core ecosystem-neutral

Before selecting commands or frameworks, require the agent to inspect the
repository for:

- language and runtime;
- build system;
- checked-in build wrappers;
- test framework and configuration;
[...]

PR #418, opened 2026-07-19
docs: add skill-gap issue form for reporting guidance mismatches

### LINKED ISSUE #412: Proposal: add a structured way to report skill gaps
Issue #404 ( the number is perfect lol )  surfaced a useful problem in `test-driven-development`: the skill assumed an npm-based workflow, while the user was working in a Java project.

What stood out to me was not only the specific gap, but how it reached the repository. It came out of a conversation on LinkedIn rather than through an existing reporting path in the repo.

`CONTRIBUTING.md` already asks users to open an issue when a skill contains incorrect or outdated guidance, but the repository currently has no issue template for this kind of report. Writing a useful issue from scratch requires more effort than many users will reasonably spend, especially when they encounter the problem in the middle of real work.

Would it be useful to add a small `skill-gap` issue form asking for:

* the affected skill;
* the relevant instruction or excerpt from `SKILL.md`;
* the project context, such as language and build tool;
* what did not work or did not apply;
* what the user had to do instead.

The goal would not be to treat every report as a confirmed defect. It would simply make real-world gaps easier to capture with enough context for maintainers to evaluate them.

This could be accompanied by a short pointer in `CONTRIBUTING.md` under the existing guidance for incorrect or outdated skills.

I would keep the first version limited to intake only: one issue form, no new automation or review process. If the reports prove useful, we could later consider whether recurring patterns can be detected or audited more systematically.

Alex7develop:
## Summary
- Adds a minimal GitHub issue form for reporting skill gaps (wrong, outdated, or ecosystem-mismatched guidance), as proposed in #412.
- Five fields only: affected skill, relevant excerpt, project context, what failed, and optional workaround — matching the maintainer green light to keep intake faster than freeform.
- Points contributors to the form from `CONTRIBUTING.md` under Reporting Issues.

## Test plan
- [ ] Open https://github.com/addyosmani/agent-skills/issues/new/choose (after merge) and confirm **Skill gap** appears
- [ ] Submit a draft with the sample Maven/`npm test` scenario and verify all required fields render in the issue body
- [ ] Confirm blank/freeform issues remain available
- [ ] Confirm the CONTRIBUTING.md link resolves to `?template=skill-gap.yml`

Closes #412

### DIFF
--- a/.github/ISSUE_TEMPLATE/skill-gap.yml
+++ b/.github/ISSUE_TEMPLATE/skill-gap.yml
@@ -1,0 +1,80 @@
+name: Skill gap
+description: Report when a skill's guidance was wrong, outdated, or did not apply in your project.
+title: "[skill-gap]: "
+labels: ["enhancement"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Use this form when a skill gave incorrect, outdated, or ecosystem-mismatched guidance.
+        Keep answers short — enough context for maintainers to triage, not a full write-up.
+
+  - type: dropdown
+    id: skill
+    attributes:
+      label: Affected skill
+      description: Which skill did not work as expected?
+      options:
+        - api-and-interface-design
+        - browser-testing-with-devtools
+        - ci-cd-and-automation
+        - code-review-and-quality
+        - code-simplification
+        - context-engineering
+        - debugging-and-error-recovery
+        - deprecation-and-migration
+        - documentation-and-adrs
+        - doubt-driven-development
+        - frontend-ui-engineering
+        - git-workflow-and-versioning
+        - idea-refine
+        - incremental-implementation
+        - interview-me
+        - observability-and-instrumentation
+        - performance-optimization
+        - planning-and-task-breakdown
+        - security-and-hardening
+        - shipping-and-launch
+        - source-driven-development
+        - spec-driven-development
+        - test-driven-development
+        - using-agent-skills
+        - other / not sure
+    validations:
+      required: true
+
+  - type: textarea
+    id: excerpt
+    attributes:
+      label: Relevant instruction or excerpt
+      description: Paste the line or short passage from SKILL.md (or a linked reference) that misled the agent or did not apply.
+      placeholder: e.g. the verification checklist said to run `npm test`
+    validations:
+      required: true
+
+  - type: textarea
+    id: context
+    attributes:
+      label: Project context
+      description: Language, build tool, test runner, or other conventions that matter here.
+      placeholder: e.g. Java 21, Maven, JUnit 5
+    validations:
+      required: true
+
+  - type: textarea
+    id: what-failed
+    attributes:
+      label: What did not work or did not apply
+      description: What happened when the agent followed the skill? What was wrong or irrelevant?
+      placeholder: e.g. agent insisted on npm scripts in a Maven repo; no package.json exists
+    validations:
+      required: true
+
+  - type: textarea
+    id: workaround
+    attributes:
+      label: What you did instead
+      description: How did you (or the agent) get unblocked? Optional but very helpful.
+      placeholder: e.g. pointed the agent at `./mvnw test` and ignored the npm checklist
+    validations:
+      required: false
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -111,6 +111,13 @@ Open an issue if you find:
 - Missing coverage for a common engineering workflow
 - Inconsistencies between skills
 
+If a skill's guidance was wrong, outdated, or did not apply in your project
+(for example, it assumed `npm test` in a Maven or Gradle repo), use the
+[Skill gap](https://github.com/addyosmani/agent-skills/issues/new?template=skill-gap.yml)
+issue form. It asks for the affected skill, the relevant excerpt, your project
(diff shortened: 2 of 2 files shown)


nucliweb:
Reviewed and verified this end to end. It matches the #412 green light cleanly: five fields (four required plus an optional workaround), faster than freeform, and it keeps blank issues available since there is no `config.yml` disabling them. Checked concretely: the YAML parses as a valid issue form, the skill dropdown matches all 24 skills under `skills/` exactly plus an "other / not sure" escape hatch, the `enhancement` label exists so the form will submit, and the CONTRIBUTING link resolves to `?template=skill-gap.yml`.

Two small, non-blocking notes:

1. The `labels: ["enhancement"]` declared in the template auto-applies `enhancement` to every issue opened through the form, and "New feature or request" reads a little off for a wrong-or-outdated-guidance report. The #412 proposal framed this as triage intake rather than a confirmed defect, so `bug` is not right either. A dedicated `skill-gap` label would fit best and make these filterable, if you are up for adding one; otherwise it is a minor thing.
2. The CONTRIBUTING blurb lists "the affected skill, the relevant excerpt, your project context, and what you did instead" but skips the required **what did not work or did not apply** field while naming the optional workaround. Worth listing the required field there for accuracy.

Nice, minimal intake form. This closes the exact gap #404 surfaced through LinkedIn.

federicobartoli:
https://github.com/user-attachments/assets/08e7be78-47c0-4c99-97bd-7e682ab7eb16

I pushed this to a fork to see the form rendered, video attached for evidence. It builds fine and the dropdown populates correctly. I also checked the option list against skills/ and all 24 are there, correctly spelled, with none missing, which was the part I expected to have drifted.

Small heads up if you want to reproduce it yourself: forks have Issues disabled by default, so the tab just isn't there until you enable it in Settings. Took me a minute to work out that it wasn't the template's fault.

Three things, none blocking.

The enhancement label reads as "New feature or request" in this repo, and a skill gap isn't really that. It isn't a confirmed bug either, since the idea was to capture reports without pre-judging them as defects. Would a dedicated skill-gap label be worth it? It would need creating before this merges, otherwise the form just drops it silently.

The skill list is static, and there's no way around that since issue forms don't support templating, so it will go stale the next time a skill lands. I'd still keep the dropdown rather than switch to a free text field, since a clean enum is the whole point for triage. Maybe just a line in the new-skill checklist in CONTRIBUTING.md pointing here? WDYT @addyosmani  @nucliweb ? 

On Addy's "expected vs actual", I think the current "what did not work" field covers it if the description also asks what the contributor expected. Probably doesn't need a sixth field.

Optional and feel free to ignore: since this is the repo's first template, it introduces the chooser page. Blank issues still work by default, but would a config.yml with blank_issues_enabled: true be worth it, so that stays a decision rather than a default?

federicobartoli:
> Reviewed and verified this end to end. It matches the #412 green light cleanly: five fields (four required plus an optional workaround), faster than freeform, and it keeps blank issues available since there is no `config.yml` disabling them. Checked concretely: the YAML parses as a valid issue form, the skill dropdown matches all 24 skills under `skills/` exactly plus an "other / not sure" escape hatch, the `enhancement` label exists so the form will submit, and the CONTRIBUTING link resolves to `?template=skill-gap.yml`.
> 
> Two small, non-blocking notes:
> 
> 1. The `labels: ["enhancement"]` declared in the template auto-applies `enhancement` to every issue opened through the form, and "New feature or request" reads a little off for a wrong-or-outdated-guidance report. The [Proposal: add a structured way to report skill gaps #412](https://github.com/addyosmani/agent-skills/issues/412) proposal framed this as triage intake rather than a confirmed defect, so `bug` is not right either. A dedicated `skill-gap` label would fit best and make these filterable, if you are up for adding one; otherwise it is a minor thing.
> 2. The CONTRIBUTING blurb lists "the affected skill, the relevant excerpt, your project context, and what you did instead" but skips the required **what did not work or did not apply** field while naming the optional workaround. Worth listing the required field there for accuracy.
> 
> Nice, minimal intake form. This closes the exact gap #404 surfaced through LinkedIn.

Full disclosure on how this happened: I wrote the first version of this on #366, which is a completely unrelated PR, then deleted it and moved it over here. Which is also neatly why I missed @nucliweb's review sitting directly above mine. They got here sixteen minutes ahead of me and landed on the same conclusion about the label, so please read my first point as a duplicate vote rather than a second opinion. More coffee next time, or possibly glasses.

Their second point is the better catch and I missed it completely: the CONTRIBUTING blurb lists the optional workaround field but skips the required "what did not work" one. Worth fixing regardless of what happens with the label.

I will leave the rest of mine up, since the static list question and the fork gotcha are still their own thing.

addyosmani:
This hits the weight I asked for on #412: five fields (affected skill, the excerpt, project context, what didn't apply, what you did instead), enough to make a report actionable without turning it into a chore people bounce off. The 80 lines are just YAML verbosity for those five, not field bloat. It closes the loop on the gap that let #404 reach us through LinkedIn instead of the tracker. Fine by me.

Alex7develop:
Thank you all for the thoughtful reviews and for merging this! I really appreciate the detailed feedback and the welcoming contribution process. It was a great learning experience, and I'm looking forward to contributing more to the project

PR #419, opened 2026-07-19
docs(tdd): make test-driven-development ecosystem-neutral (#404 Phase 1)

ayobamiseun:
Implements Phase 1 of #404 as scoped in the issue thread. The TDD workflow now discovers the repository's stack (language, build system, wrappers, test framework, documented commands) before choosing any test command, and runs RED–GREEN–REFACTOR against the repo's own focused/full-suite commands instead of a hardcoded `npm test`. `references/testing-patterns.md` is relabeled in place as JavaScript/TypeScript examples illustrating the universal principles — no relocation, per the #361/#236 sequencing agreed there. A new behavioral eval backs it with a Python/unittest fixture verifying the skill adapts to non-npm tooling. Ecosystem-specific references remain practitioner-authored follow-ups (Phase 2).

Validation: `validate-skills.js` and `run-evals.js` pass (rank-1 86%); the fixture's suite passes with `python3 -m unittest`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ node_modules/
 .claude/.simplify-ignore-cache/
 .claude/sdd-cache/
 evals/results/
+__pycache__/
--- a/README.md
+++ b/README.md
@@ -280,7 +280,7 @@ Quick-reference material that skills pull in when needed:
 | Reference | Covers |
 |-----------|--------|
 | [definition-of-done.md](references/definition-of-done.md) | Project-wide standing bar every change clears, contrasted with per-task acceptance criteria |
-| [testing-patterns.md](references/testing-patterns.md) | Test structure, naming, mocking, React/API/E2E examples, anti-patterns |
+| [testing-patterns.md](references/testing-patterns.md) | Test structure, naming, mocking, React/API/E2E examples, anti-patterns (JavaScript/TypeScript) |
 | [security-checklist.md](references/security-checklist.md) | Pre-commit checks, auth, input validation, headers, CORS, OWASP Top 10 |
 | [performance-checklist.md](references/performance-checklist.md) | Core Web Vitals targets, frontend/backend checklists, measurement commands |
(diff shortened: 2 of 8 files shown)


nucliweb:
Reviewed this against the Phase 1 scope settled in #404, and it lands it faithfully. The core now discovers the stack before choosing any command, `npm test` is demoted from *the* default to one example among gradle/pytest/go everywhere it appears, and `testing-patterns.md` is relabeled as JS/TS in place, no relocation, which is exactly the #361/#236 sequencing we agreed to keep this PR from stepping on the packaging discussion.

Verified locally:
- The new `test-driven-development-ecosystem` fixture runs clean (`python3 -m unittest`, 3 passing).
- `node scripts/run-evals.js` passes: 124 checks, 0 errors, and eval case 3 is well-formed and consistent with cases 1 and 2.
- No stale "across the stack" / "across frameworks" labels remain in the tree.

Scope discipline is good too: it deliberately leaves the other four skills with npm-family commands in verification checkboxes (the set @federicobartoli grepped) for the separate systematic pass, rather than widening this PR.

One small, non-blocking note on eval case 3: the prompt says a debit "must not drive the balance below zero" but does not pin the behavior when it would (raise, like the existing `ValueError` on unknown kinds, vs. clamp at zero). The eval grades TDD process rather than domain semantics, so it still does its job, but spelling out the expected behavior would remove a bit of grading ambiguity on the "below-zero rule gets its own test case" expectation.

Nice work, this matches what we scoped.

addyosmani (approved):
This is the right fix for the TDD half, and I like that you didn't just soften the wording: you pulled the hardcoded `npm test` out of the verification checklist, reframed `testing-patterns.md` honestly as JS/TS examples of universal principles, and added a Python fixture with a real `unittest` eval so the ecosystem-neutrality is actually guarded rather than asserted. Suite stays green.

Two notes. This is TDD only, and Federico's grep on #404 found the same npm assumption in nine more spots across four other skills, so treat this as Phase 1 of a catalog-wide pass rather than the whole fix. And heads up that #410 and #416 also touch this SKILL.md, so whoever lands second will rebase. Approving this one.

PR #420, opened 2026-07-20
fix(hooks): quote opt-in hook paths in setup docs

jagadishs-oss:
## Summary
- Quote `${CLAUDE_PROJECT_DIR}` in the `SDD-CACHE.md` and `SIMPLIFY-IGNORE.md` setup examples so opt-in hooks work when the project path contains spaces.
- Same class of failure as #214 (SessionStart already fixed on `main`); these docs still shipped the unquoted form.
- No overlap with open hooks PRs (#294/#117/#385 touch the scripts, not these setup snippets).

## Test plan
- [ ] Confirm the five `command` examples in `hooks/SDD-CACHE.md` and `hooks/SIMPLIFY-IGNORE.md` use `bash "${CLAUDE_PROJECT_DIR}/hooks/..."`.
- [ ] Mentally expand a path like `/Users/First Last/project` and verify the quoted form stays a single argv to `bash`.
- [ ] Optional: copy a quoted snippet into `.claude/settings.json` under a directory with a space and confirm the hook script is invoked.


federicobartoli (approved):
LGTM.

addyosmani:
Small but correct: quoting the hook paths so a space in the install path doesn't break the command is the same class of bug as the SessionStart path issue in #214. Fine to take.

PR #422, opened 2026-07-21
docs(observability): expand runbook guidance in alerting section

HMAKT99:
## What this adds

A `### Writing Runbooks` subsection directly under the existing alerting rules in `skills/observability-and-instrumentation/SKILL.md`.

## Why here (not in documentation-and-adrs)

PR #414 tried to add runbook guidance to `documentation-and-adrs`. The review redirect was right: the skill already names runbooks as a hard requirement on every alert ("It links to a runbook — even three lines: what it means, first query to run, escalation path") and is the natural home for teaching the artifact it requires. This PR expands that one-liner into a usable guide without moving it.

## What it teaches

- The three-question format every runbook must answer: what is happening / what to check first / who to escalate to
- A minimum viable runbook template (stored in `docs/runbooks/`, named after the alert)
- When to expand beyond three lines: add steps only when the first check alone isn't enough to decide
- Keep-current discipline: update as part of closing every incident the runbook was used in

## Scope

+17 lines. One file. Fully additive — no edits to existing prose. `validate-skills.js` → 0 errors, 0 warnings.

### DIFF
--- a/skills/observability-and-instrumentation/SKILL.md
+++ b/skills/observability-and-instrumentation/SKILL.md
@@ -154,6 +154,23 @@ Rules for every alert you create:
 3. **It has a threshold and duration** justified by the SLO or by historical data, not by a guess.
 4. Use two severities only: **page** (user-facing, act now) and **ticket** (degradation, act this week). A third tier becomes noise that trains people to ignore everything.
 
+### Writing Runbooks
+
+A runbook is the document your alert links to. Its job is to answer three questions without requiring the reader to think: what is happening, what to check first, and who to call if that doesn't resolve it. Store in `docs/runbooks/` named after the alert.
+
+**Minimum viable runbook (three lines):**
+
+```markdown
+# Runbook: High Error Rate on /api/tasks
+**Means:** DB connection pool likely exhausted, or a bad deploy.
+**First check:** `SELECT count(*) FROM pg_stat_activity;` — if > pool limit, see Step 2.
+**Escalate to:** #db-oncall or engineering on-call rotation.
+```
+
+**When to expand beyond three lines:** add steps only when the first check alone isn't enough to decide. A five-step runbook that covers the three most common causes is better than a twenty-step document that covers every edge case and gets skimmed.
+
+**Keep runbooks current.** Update the runbook as part of closing every incident it was used in — a stale runbook builds false confidence. If a step was wrong or missing, fix it before marking the incident resolved.
+
 ### 7. Verify the telemetry itself
 
 Instrumentation is code; it can be wrong. Before calling the work done, trigger the paths and look at the actual output:
--- a/skills/observability-and-instrumentation/SKILL.md
+++ b/skills/observability-and-instrumentation/SKILL.md
@@ -154,16 +154,17 @@ Rules for every alert you create:
 3. **It has a threshold and duration** justified by the SLO or by historical data, not by a guess.
 4. Use two severities only: **page** (user-facing, act now) and **ticket** (degradation, act this week). A third tier becomes noise that trains people to ignore everything.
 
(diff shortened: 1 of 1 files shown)


addyosmani (approved):
This is the right home and much tighter than the original: +17 under the alerting rules that already require a runbook link, versus +79 in the ADR skill. The worked three-line example, the "expand only when the first check isn't enough" judgment, and updating the runbook as part of closing the incident are all genuinely new, the last one especially since a stale runbook builds false confidence.

One nit, not blocking: the opening sentence restates the same three elements the alerting rule fifteen lines above already lists, so it could point at that rule instead of re-listing them. Approving, and I'll close #414 as superseded by this.

addyosmani:
@nucliweb @federicobartoli when either of you has a minute, this one just needs a second pair of eyes to clear our bar. It's 17 lines adding a `### Writing Runbooks` subsection under the alerting rules in observability, redirected here from #414 after I pushed back on runbooks living in the ADR skill. I've approved it and it's clean and green locally, but I'd rather not merge on my own review alone. Small one, and the only nit I left is that its opening sentence restates the three-element runbook rule a few lines above it.

nucliweb (approved):
**Second review as Addy requested. This crosses the bar for me, happy to see it merge.**

This is the right home: the alerting rules already make a runbook a hard requirement on every alert (rule 2), so teaching the artifact right where it's mandated is the natural fit, and step 7 already verifies "the runbook link works", so the new guidance is consistent with what's here rather than bolted on. The worked three-line example, the "expand only when the first check isn't enough" judgment, and tying runbook upkeep to incident closure are all genuinely new and opinionated in the right way.

One structural nit, non-blocking. The Process section is a numbered sequence (`### 1.` … `### 7.`), and `### Writing Runbooks` lands unnumbered between `### 6. Alerting` and `### 7. Verify…`, so it reads as a peer step that breaks the 6→7 flow. Since it's really a sub-topic of Alerting, `#### Writing Runbooks` nested under step 6 would keep the numbering intact. That also lines up with the nit Addy already left: the opening sentence restates the three elements from alerting rule 2 a few lines above, so nesting it under that rule and pointing back at it (instead of re-listing them) would fix both at once.

Neither is a blocker; approving.

federicobartoli:
Thanks, this already crossed the bar with @addyosmani  and @nucliweb , so treat everything here as one optional round before merge, and we can land it as-is if you prefer.

Could we nest the section as #### Writing Runbooks under step 6 and open it by pointing at alerting rule 2 instead of re-listing its three elements? That would pick up both non-blocking notes from the earlier reviews in one commit.

Two small touches on the example while you are in the file, if you agree. pg_stat_activity also counts background processes and other apps' connections, so the comparison with the app's pool limit reads tighter as SELECT count(*) FROM pg_stat_activity WHERE backend_type = 'client backend';. And since the catalog labels worked examples as illustrations (the TDD skill does this explicitly), maybe one short clause like "the example assumes Postgres; swap in the equivalent first check for your store"?

None of this blocks: if you would rather land it as-is, these can ride along in a future edit.

HMAKT99:
All three points addressed, @federicobartoli — took the optional round:

- **Nested as `####`** under step 6 alerting, opening with a direct pointer to rule 2 (*"Rule 2 above requires every alert to link to a runbook"*) instead of re-listing its three elements.
- **SQL tightened** — `pg_stat_activity WHERE backend_type = 'client backend'` so it counts only app connections, not background processes. Added a parenthetical: *"Swap in the equivalent for your database."*
- **Platform hedge** already implied by that parenthetical — if you'd like it made more explicit, happy to adjust.

Validator 0/0.

federicobartoli (approved):
Landing this, thanks for taking the optional round.

PR #424, opened 2026-07-22
Make shared skill workflows agent-agnostic

ujjwalks:
## Summary

Two shared skills currently instruct the agent in Claude Code's tool vocabulary, which reads as a hard requirement on harnesses that have the same capability under a different name.

- `idea-refine`: replace `AskUserQuestion` and `` `Glob`/`Grep`/`Read` `` with the capability each one stands for — ask focused questions; discover, search, and read relevant files. Workflow and stopping conditions unchanged.
- `browser-testing-with-devtools`: lead with the transport-neutral requirement (launch `chrome-devtools-mcp` as a local STDIO MCP server), then give the existing `.mcp.json` under a labeled **Claude Code and compatible clients** heading and add a **Codex** equivalent (`codex mcp add` plus project-scoped `.codex/config.toml`).

Scoped down from the original submission per review on #424:

- the `superpowers/` plan and spec docs are removed — working artifacts against an external framework, and `docs/` carries no `plans/`/`specs/` convention;
- the portability lint rule is removed — a blocking catalog-wide gate is a policy decision that deserves its own thread, and it overlaps #428's rework of the same file;
- the `doubt-driven-development` edit is removed — #380 rewrites the same paragraph and has been open longer. It can be re-cut on top of #380 if still wanted.

No overlap with the #404 ecosystem-neutrality track (#419, #425): different axis, no shared files.

## Verification

- `node scripts/validate-skills.js` — 24 skills, 0 errors, 0 warnings
- `node scripts/run-evals.js` — 124 checks passed, 0 errors, 0 warnings; rank-1 86% (65/76)
- `git diff --check origin/main...HEAD` — clean
- Rebased onto `main` at 7829ffd (post-#425)

### DIFF
--- a/skills/browser-testing-with-devtools/SKILL.md
+++ b/skills/browser-testing-with-devtools/SKILL.md
@@ -25,7 +25,11 @@ Use Chrome DevTools MCP to give your agent eyes into the browser. This bridges t
 
 ### Installation
 
-Add the following to your project's `.mcp.json` or Claude Code settings:
+Your MCP-capable agent harness must launch `npx -y chrome-devtools-mcp@latest --isolated` as a local STDIO MCP server.
+
+#### Claude Code and compatible clients
+
+For Claude Code and compatible clients, `.mcp.json` is one harness-specific representation of that server configuration:
 
 ```json
 {
@@ -37,6 +41,24 @@ Add the following to your project's `.mcp.json` or Claude Code settings:
   }
 }
 ```
+
+#### Codex
+
+Add the server from the command line:
+
+```bash
+codex mcp add chrome-devtools -- npx -y chrome-devtools-mcp@latest --isolated
+```
+
+Or use the equivalent project-scoped `.codex/config.toml` configuration:
+
+```toml
+[mcp_servers.chrome_devtools]
+command = "npx"
+args = ["-y", "chrome-devtools-mcp@latest", "--isolated"]
+```
+
+Codex loads project configuration only for trusted projects.
 
 `-y` skips the npx install confirmation. By default the server launches Chrome with its own dedicated profile (under `~/.cache/chrome-devtools-mcp/`), separate from your personal browser; `--isolated` goes one step further and uses a temporary profile that is wiped when the browser closes. This is the right setup for most testing.
 
--- a/skills/idea-refine/SKILL.md
+++ b/skills/idea-refine/SKILL.md
@@ -66,7 +66,7 @@ When the user invokes this skill with an idea (`$ARGUMENTS`), guide them through
    - What's been tried before?
    - Why now?
 
-   Use the `AskUserQuestion` tool to gather this input. Do NOT proceed until you understand who this is for and what success looks like.
+   Ask the user these questions using the interaction mechanism available in the current agent. Do NOT proceed until you understand who this is for and what success looks like.
 
 3. **Generate 5-8 idea variations** using these lenses:
    - **Inversion:** "What if we did the opposite?"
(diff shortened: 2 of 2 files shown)


addyosmani (changes requested):
The underlying goal here is right and it's one we're actively working. The browser-testing change in particular (harness-neutral MCP requirement up top, then labeled Claude Code and Codex config blocks) is exactly the shape we want. But two things stop me from merging this as it stands.

First, it overlaps an in-flight track. #404 is doing the same agent-agnostic pass surgically, skill by skill, human-curated: #419 already landed for TDD and #425 is Phase 2 in review right now. Two parallel efforts rewriting the same skills will collide. Before more skill edits go in here, this needs to be reconciled with that track rather than run alongside it.

Second, the packaging brings in things I don't think we want in the canonical tree. `docs/superpowers/plans/` and `docs/superpowers/specs/` are working docs written against an external framework, with lines like "REQUIRED SUB-SKILL: use superpowers:subagent-driven-development." This repo is deliberately vendor-neutral and doesn't carry third-party framework references, so those files don't belong here even as planning artifacts. And the new lint is a blocking CI rule that errors on `AskUserQuestion` / backticked `Glob`/`Grep`/`Read` anywhere outside a labeled platform heading. That's a real policy decision (we ship as a Claude Code plugin and reference those tools deliberately in places), and it's a big blast radius to fold into a contributor PR. A blocking, catalog-wide "no Claude tool names" gate is the kind of thing that needs to be decided on its own, not arrive as a side effect.

Net: the Codex/MCP-neutral examples are salvageable and good. If you drop the superpowers docs and the blocking lint, and rebase the skill edits onto the #404 track so they don't fight #419/#425, there's something mergeable here. As-is I can't take it.

ujjwalks:
Acting on all three. Force-pushed: this PR is now 25 lines across two files — `idea-refine` and `browser-testing-with-devtools`. Everything else is out.

**superpowers docs** — dropped. `docs/` is flat here and those were working artifacts written against an external framework; they shouldn't have been in the diff.

**The lint** — dropped from this PR. You're right that a blocking, catalog-wide gate is policy, not a side effect of a docs change, and it's policy that currently cuts against `docs/codex-setup.md`, which explicitly accepts Claude Code-specific pieces with per-harness guidance in `docs/`. If you want it litigated on its own I'll open an issue proposing it as a **warning** first, with the offending term named in the message. Not opening one until you say it's worth the thread.

**On the overlap** — one correction for the record: #419/#425 and this PR don't share a file or a line. That track is ecosystem-neutrality (`npm test` → the repo's own commands); this is harness-neutrality (Claude tool names → capability language). #425 has also since merged, so the collision window is closed either way.

But I did find real in-flight overlap you didn't mention, and it's worse than the #404 one:

- **#380 rewrites the exact bullet** in `doubt-driven-development` that this PR rewrote — same paragraph, incompatible direction (it keeps the Claude Code framing and adds nuance about nested spawn being technically available; mine deleted that nuance). #380 has been open since July 10. I've pulled the `doubt-driven-development` commit out of this PR entirely and will re-cut it on top of #380 once that lands, if you still want it then.
- **#428 rewrites the region of `skill-lint.js`** the linter commit touched and adds its own test file at `scripts/skill-lint-test.js`. Another reason the lint doesn't belong here. I've taken the one genuinely reusable piece of that commit — a fence-parsing bug in `stripFencedCodeBlocks` that makes #428's own cross-reference rule miss tilde and indented fences — and offered it there instead, rebased on their head.
- **#195** (draft) adds `AskUserQuestion` to 10 skills, including `idea-refine`. Whichever direction you pick, those two can't both be right — more evidence the tool-name question wants its own decision.

So: is harness-neutrality worth its own tracking issue, curated the way #404 is? Happy to write it up as a phased proposal rather than keep sending drive-by PRs at it. Or if the answer is "not a direction we're taking," say so and I'll stop at these two files.

Verification on the current head: `validate-skills.js` 24 skills 0 errors, `validate-commands.js` 8 commands 0 errors, `run-evals.js` 124 checks, rank-1 86%. Rebased onto `main` at 7829ffd (post-#425).

PR #425, opened 2026-07-22
docs(skills): extend ecosystem-neutral commands catalog-wide (#404 Phase 2)

ayobamiseun:
Follow-up to #419, which addyosmani asked to treat as Phase 1 of a catalog-wide pass. This addresses the remaining normative npm-family commands from @federicobartoli's grep on #404: incremental-implementation's four exit criteria, planning-and-task-breakdown's task-template verification lines, shipping-and-launch's `npm audit` checkbox, and the imperative command blocks in debugging-and-error-recovery — each now uses or points to the repository's own commands, referencing the TDD skill's Discover the Stack First section instead of duplicating it. `references/security-checklist.md`'s OWASP row generalizes likewise. security-and-hardening needed no change: its SKILL.md was already neutralized by the supply-chain work. Ride-along: pins the below-zero debit behavior (ValueError) in TDD eval case 3, per @nucliweb's non-blocking note on #419.

Illustrative-only material (CI pipeline YAML, example CLAUDE.md/README snippets in context-engineering, documentation-and-adrs, spec-driven-development) is deliberately untouched — those depict concrete example projects, not universal exit criteria.

Validation: `validate-skills.js`, `validate-commands.js`, and `run-evals.js` all pass (rank-1 86%).

Rebase notes: #415 adds an SLO section to shipping-and-launch (different section, trivial rebase for whoever lands second), and #383 touches planning-and-task-breakdown's plan template header (adjacent to, not overlapping, the lines changed here).

### DIFF
--- a/evals/cases/test-driven-development.json
+++ b/evals/cases/test-driven-development.json
@@ -55,7 +55,7 @@
     },
     {
       "id": 3,
-      "prompt": "Add debit entries to the ledger, test-first. A debit subtracts its amount from the balance and must not drive the balance below zero.",
+      "prompt": "Add debit entries to the ledger, test-first. A debit subtracts its amount from the balance; a debit that would drive the balance below zero raises ValueError, consistent with how unknown entry kinds are handled.",
       "expected_output": "A failing unittest written and shown failing first, a minimal implementation, and the full suite run with the repository's own Python test command",
       "files": [
         "test-driven-development-ecosystem"
--- a/references/security-checklist.md
+++ b/references/security-checklist.md
@@ -181,7 +181,7 @@ res.status(500).json({
 | 3 | Injection | Parameterized queries, input validation |
 | 4 | Insecure Design | Threat modeling, spec-driven development |
 | 5 | Security Misconfiguration | Security headers, minimal permissions, audit deps |
-| 6 | Vulnerable Components | Native dependency audit (`npm audit`, `pip-audit`, ...), keep deps updated, minimal deps |
+| 6 | Vulnerable Components | The ecosystem's dependency audit (`npm audit`, `pip-audit`, ...), keep deps updated, minimal deps |
 | 7 | Auth Failures | Strong passwords, rate limiting, session management |
 | 8 | Data Integrity Failures | Verify updates/dependencies, signed artifacts |
 | 9 | Logging Failures | Log security events, don't log secrets |
--- a/references/security-checklist.md
+++ b/references/security-checklist.md
@@ -181,7 +181,7 @@ res.status(500).json({
 | 3 | Injection | Parameterized queries, input validation |
 | 4 | Insecure Design | Threat modeling, spec-driven development |
 | 5 | Security Misconfiguration | Security headers, minimal permissions, audit deps |
-| 6 | Vulnerable Components | `npm audit`, keep deps updated, minimal deps |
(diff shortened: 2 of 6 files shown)


addyosmani (approved):
This is Phase 2 of the same ecosystem-neutral pass we already landed for TDD in #419, and it holds the same shape: swap hardcoded `npm` for "the repository's own command," keep one concrete example so the guidance stays readable, extend the pattern to the security checklist's audit row (`pip-audit`, `cargo audit`). Clean, green, and it doesn't touch routing or frontmatter. Test-merged fine and evals stay at 86%.

One honest note, not blocking: the parentheticals are starting to stack up. The debugging one ("For test failures (npm shown, substitute the repository's own test command, per the test-driven-development skill's Discover the Stack First section):") is doing a lot of work in a single line. It reads fine here, but if Phase 3 keeps adding the full cross-reference each time, the skills will get noisier than the npm-hardcoding they're replacing. Might be worth a shorter house phrasing for the back half of this series. Approving as-is.

federicobartoli:
One things I noticed while re-checking:
The security checklist row now reads "Native dependency audit (`npm audit`, `pip-audit`, ...)". I stumbled on "native" because native dependencies already means something else (compiled modules, C extensions), so an agent could parse the row as "audit your native dependencies" rather than "use the ecosystem's own audit tool". The shipping-and-launch line in this same PR says "The ecosystem's dependency audit", which is unambiguous. Since it's a one-word alignment, would it be worth using that phrasing in both places before this lands?
Thank u 🙏

nucliweb:
Reviewed as the continuation of #419 (I looked at Phase 1), and it lands Phase 2 faithfully and completely. It hits exactly @federicobartoli's grep set from #404, incremental-implementation's four exit criteria, planning-and-task-breakdown's task template, shipping-and-launch's audit checkbox, and the debugging-and-error-recovery command blocks, and it points at the TDD skill's Discover the Stack First section rather than duplicating the guidance, which is the right call.

Verified independently:
- The "security-and-hardening needed no change" claim holds: its remaining npm mentions sit in the supply-chain section that already says "Do not assume npm...", so nothing normative is left hardcoded there.
- The npm references still in the touched files are all demoted to one example among many (`npm test`, `./gradlew test`, `pytest`, ...), matching the Phase 1 pattern rather than prescribing npm.
- `validate-skills.js`, `validate-commands.js`, and `run-evals.js` all pass locally (124 checks, rank-1 86%), matching what you reported.

And thanks for pinning the below-zero behavior to `ValueError` in eval case 3, that resolves the ambiguity I flagged on #419 cleanly.

One optional, non-blocking polish note: the neutralized commands are expressed in two styles across the files, bracket placeholders (`[the repository's focused-test command]` in planning-and-task-breakdown) versus parenthetical example lists (`(npm test, ./gradlew test, pytest, ...)` in incremental-implementation). Both read fine, and the bracket form does match planning's existing task-template style, so this is just a consistency thought if you want one convention across the pass; I would not hold the PR for it.

Solid, well-scoped continuation.

ayobamiseun:
@federicobartoli good catch — "native" does collide with the compiled-modules meaning, and the two lines in this PR should say the same thing. Fixed: the checklist row is aligned to "The ecosystem's dependency audit" in 2457149.

federicobartoli (approved):
LGTM!

PR #426, opened 2026-07-24
ci: validate repository-local Markdown links and anchors

HendrylHH:
## Summary

- add a dependency-free validator for repository-local Markdown paths and heading fragments
- cover inline and reference links, fenced and inline code, escaped destinations, query strings, Setext and ATX headings, and duplicate slug collisions
- add focused `node:test` coverage and run the validator in CI
- keep diagnostics deterministic and actionable with source file, line, destination, and reason

## Why

The existing repository checks validate skill content, command parity, and evaluation routing, but they do not detect broken repository-local Markdown paths or anchors across documentation, agents, references, and commands.

Previous pull requests fixed individual broken references. This adds the missing regression gate so future moves, renames, and heading edits fail CI before broken navigation reaches users.

## Impact

Contributors get an early, dependency-free CI check for invalid local links and anchors. External URLs are intentionally left out of scope, avoiding flaky network-dependent validation.

## Validation

- `node --test scripts/validate-links-test.js` — 14/14 passed
- `node --test scripts/run-evals-test.js` — 12/12 passed
- `node scripts/validate-links.js` — 86 files checked, 0 broken links
- `node scripts/validate-skills.js` — 24 skills, 0 errors or warnings
- `node scripts/run-evals.js --min-rank1 80` — 124 checks, 86% rank-1, passed
- `git diff --check` — passed

## Duplicate check

Immediately before publication, open issues and pull requests were searched for `markdown link`, `link validator`, `link checker`, and `broken anchor`. The only adjacent open item was #366, which concerns shared reference URLs for `npx` installs and does not add repository-wide local path and anchor validation.

### DIFF
--- a/.github/workflows/test-plugin-install.yml
+++ b/.github/workflows/test-plugin-install.yml
@@ -40,9 +40,26 @@ jobs:
       - name: Validate commands across all tool directories
         run: node scripts/validate-commands.js
 
+  validate-links:
+    name: Validate repository-local Markdown links
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Test Markdown link validator
+        run: node --test scripts/validate-links-test.js
+
+      - name: Validate repository-local Markdown links
+        run: node scripts/validate-links.js
+
   validate:
     name: Validate plugin structure
-    needs: [validate-skills, validate-commands]
+    needs: [validate-skills, validate-commands, validate-links]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
--- a/scripts/lib/markdown-link-lint.js
+++ b/scripts/lib/markdown-link-lint.js
@@ -1,0 +1,427 @@
+'use strict';
+
+const path = require('node:path');
+
+function stripFencedCodeBlocks(markdown) {
+  const lines = markdown.split('\n');
+  let fence = null;
+
+  return lines.map((line) => {
+    if (fence !== null) {
+      const closingFence = new RegExp(
+        `^ {0,3}\\${fence.character}{${fence.length},}[ \\t]*$`,
+      );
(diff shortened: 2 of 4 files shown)


addyosmani (commented):
The problem this targets is real. We've had anchor and cross-link breakage bite us recently (#401 and #236 both had to be checked by hand for exactly this), so a link/anchor linter is a reasonable thing to want.

My hesitation is proportionality. This is 884 lines of new lint infrastructure wired in as a blocking CI step, and once it's blocking, every future doc edit lives or dies by it, including its false-positive behavior on things like intentional anchors, external links, and generated content. Before I'd sign off I'd want to understand: what's the false-positive story, is the blocking step scoped to repo-local links only (the title says so, but I want to confirm it never reaches out to the network in CI), and could the same coverage come from something materially smaller. CI also hasn't run here yet, so I can't see it pass on our actual tree.

This is a maintainer call on whether we take on that much linting surface, so I'm not going to merge it unilaterally. Flagging it for a proper look rather than sitting on it. Thanks for putting in the work; the intent is right.

HendrylHH:
That hesitation is reasonable. After your comment and re-reading the implementation, I don’t think the current false-positive is strong enough yet for a blocking check of this size.

Answering your questions: The current implementation is fully offline. It only runs git ls-files and reads tracked Markdown files from disk. Destinations with a URL scheme, protocol-relative URLs, and root-relative URLs are discarded before validation; there is no HTTP client or network path in CI.

The current false-positive protections cover fenced and inline code, image-only references, escaped link-like text, duplicate headings, Setext/ATX headings, and explicit HTML id/name anchors. However, I do see that are two real gaps: 1) every tracked Markdown file is scanned, with no exclusion or suppression mechanism for generated content; 2) GitHub-style heading slugs are reproduced by custom code, which can drift from GitHub’s rendering behavior.

The tests proves that reduce that risk, but they don’t remove the maintenance burden of a custom Markdown parser.

I’ll take another pass at the implementation with proportionality in mind and explore a smaller approach before requesting another review. I’d be happy to update the draft along those lines if that sounds like the right direction.

PR #427, opened 2026-07-24
fix(agents): delegate code-reviewer persona to code-review-and-quality skill

### LINKED ISSUE #172: Persona/skill duplication: code-reviewer.md inlines the 5-axis framework instead of delegating to code-review-and-quality skill
## Summary

`agents/code-reviewer.md` (98 lines) inlines its own copy of the five-axis
review framework that already exists in
`skills/code-review-and-quality/SKILL.md` (348 lines). The persona file
never references the skill — there is no link, import, or "follow the skill"
directive. This contradicts the repo's stated design and is already showing
wording drift.

## The intended design (from the repo's own docs)

`agents/README.md:21`:
> "Skills are mandatory hops inside a persona's workflow."

`AGENTS.md:76`:
> "A persona may invoke skills."

The expected pattern: persona = voice + output format + "go follow the skill".
The skill is the single source of truth for the *what* and *how*.

## What the repo actually has

`agents/code-reviewer.md:10-46` contains a self-contained 5-axis framework
(Correctness, Readability, Architecture, Security, Performance), each with
3-5 bullets describing what to check. The persona file never mentions
`code-review-and-quality` anywhere — including in its `## Composition`
block (lines 93-97), which only references `/review` and `/ship` slash
commands.

`skills/code-review-and-quality/SKILL.md:22-80` contains its own, more
detailed version of those same five axes, plus a severity rubric, multi-
model review pattern, dead-code hygiene, dependency discipline, change
sizing table, and a long exit checklist that exists nowhere else.

Note the asymmetry: the **skill** does cross-link to other skills
(`skills/code-review-and-quality/SKILL.md:61` → `security-and-hardening`,
`:74` → `performance-optimization`), but the **persona** does not link to
the skill that is supposed to be its mandatory hop.

## Drift is already showing

The two copies are not in sync. Examples:

| Axis | Persona (`code-reviewer.md`) | Skill (`code-review-and-quality/SKILL.md`) |
|---|---|---|
| Correctness — tests | "Do the tests actually verify the behavior? Are they testing the right things?" (line 17) | "Does it pass all tests? Are the tests actually testing the right things?" (line 33) |
| Readability | 4 bullets, no mention of dead code, no abstraction-cost rule (lines 20-24) | 8 bullets including dead-code artifacts, "abstractions earning their complexity", "could this be done in fewer lines?" (lines 38-47) |
| Architecture | 5 bullets, no mention of code duplication (lines 26-31) | 5 bullets, includes "Is there code duplication that should be shared?" (lines 53-57) |
| Security | 5 bullets, no `security-and-hardening` cross-link (lines 33-38) | 8 bullets + cross-link to `security-and-hardening` for detailed guidance (lines 59-70) |

When the skill is updated (e.g. a new severity rule, a new dead-code item),
the persona will silently drift further. Code-paths that route to the
persona vs the skill will deliver inconsistent reviews of the same change.

## Why this is the structural root cause of the routing problem

(See companion issue.) Because the persona body duplicates the skill rather
than calling it, the persona's description has to claim "I do code review"
in a way that overlaps with the skill's description. The two descriptions
compete for the same intent precisely because the two artifacts redundantly
cover the same job.

## Proposed fixes (sketch)

**Option A — true delegation (preferred):** Strip the inline framework
from `code-reviewer.md` and replace it with an explicit invocation:

~~~markdown
## Process
Invoke the `code-review-and-quality` skill and follow its workflow exactly.

## Voice and output format
[the persona-specific bits: Staff Engineer tone,
 Verdict / Critical / Important / Suggestion / What's Done Well structure]
~~~

Pros: single source of truth, no drift, persona description can shrink to
"voice + output format only".
Cons: persona loses portability to harnesses that can't invoke skills
(plain Markdown system-prompt setups). Mitigation: keep an "Appendix for
harnesses without skill support" block that mirrors the skill's checklist.

**Option B — explicit standalone copy with a sync marker.** Keep both
copies but add a `Synced from code-review-and-quality SKILL.md @ <commit>`
line at the top of the persona file plus a CI check that fails if the
skill's framework changes without the persona being touched.
Pros: persona stays portable.
Cons: ongoing maintenance burden; drift just becomes harder, not impossible.

**Option C — collapse.** Remove the persona entirely; have `/review` go
straight to the skill. The persona's distinct voice and output format would
move into either the skill or the slash command.

The same duplication pattern exists for `security-auditor` ↔
`security-and-hardening` and `test-engineer` ↔ `test-driven-development`,
so whichever direction is chosen should be applied across all three pairs.

Happy to send a PR implementing Option A if the direction sounds right.

BryanDhaniel:
Fixes #172

`agents/code-reviewer.md` inlined the full five-axis review framework
instead of referencing `skills/code-review-and-quality/SKILL.md`, which
already documents the same framework in more detail. This duplicated
content that would drift out of sync as the skill evolves — something
CONTRIBUTING.md explicitly warns against.

The persona now loads the skill for the review framework and keeps
only what's specific to the persona: role, output format, and
composition rules.

No eval changes needed — this is a persona edit, not a new/modified skill.

### DIFF
--- a/agents/code-reviewer.md
+++ b/agents/code-reviewer.md
@@ -9,40 +9,9 @@ You are an experienced Staff Engineer conducting a thorough code review. Your ro
 
 ## Review Framework
 
-Evaluate every change across these five dimensions:
+## Review Framework
 
-### 1. Correctness
-- Does the code do what the spec/task says it should?
-- Are edge cases handled (null, empty, boundary values, error paths)?
-- Do the tests actually verify the behavior? Are they testing the right things?
-- Are there race conditions, off-by-one errors, or state inconsistencies?
-
-### 2. Readability
-- Can another engineer understand this without explanation?
-- Are names descriptive and consistent with project conventions?
-- Is the control flow straightforward (no deeply nested logic)?
-- Is the code well-organized (related code grouped, clear boundaries)?
-
-### 3. Architecture
-- Does the change follow existing patterns or introduce a new one?
-- If a new pattern, is it justified and documented?
-- Are module boundaries maintained? Any circular dependencies?
-- Is the abstraction level appropriate (not over-engineered, not too coupled)?
-- Are dependencies flowing in the right direction?
-
-### 4. Security
-- Is user input validated and sanitized at system boundaries?
-- Are secrets kept out of code, logs, and version control?
-- Is authentication/authorization checked where needed?
-- Are queries parameterized? Is output encoded?
-- Any new dependencies with known vulnerabilities?
-
-### 5. Performance
-- Any N+1 query patterns?
-- Any unbounded loops or unconstrained data fetching?
-- Any synchronous operations that should be async?
-- Any unnecessary re-renders (in UI components)?
-- Any missing pagination on list endpoints?
+Load [skills/code-review-and-quality/SKILL.md](../skills/code-review-and-quality/SKILL.md) and evaluate every change using its five-axis framework (correctness, readability, architecture, security, performance) and approval standard. That skill is the source of truth for what each axis checks; this persona applies it and adds the role, output format, and composition rules below.
 
 ## Output Format
 
--- a/agents/code-reviewer.md
+++ b/agents/code-reviewer.md
@@ -9,9 +9,17 @@ You are an experienced Staff Engineer conducting a thorough code review. Your ro
 
 ## Review Framework
 
-## Review Framework
+**Canonical source:** [skills/code-review-and-quality/SKILL.md](../skills/code-review-and-quality/SKILL.md) documents this framework in full — read it if you have access to this repo. The summary below keeps this persona self-contained for when it's copied out standalone into another tool (see `docs/copilot-setup.md`, `docs/gemini-cli-setup.md`).
 
-Load [skills/code-review-and-quality/SKILL.md](../skills/code-review-and-quality/SKILL.md) and evaluate every change using its five-axis framework (correctness, readability, architecture, security, performance) and approval standard. That skill is the source of truth for what each axis checks; this persona applies it and adds the role, output format, and composition rules below.
+Evaluate every change across five axes:
+
+1. **Correctness** — Does it do what the spec/task requires? Edge cases and error paths handled? Do the tests verify real behavior? Any race conditions, off-by-one errors, or state inconsistencies?
+2. **Readability** — Understandable without the author explaining it? Names clear and conventional? Control flow simple (no deep nesting)? Well-organized, no dead code?
+3. **Architecture** — Follows existing patterns, or justifies a new one? Module boundaries and dependency direction respected? Abstraction level appropriate (not over- or under-engineered)?
+4. **Security** — Input validated and sanitized at boundaries? No secrets in code, logs, or version control? Auth checked where needed? Queries parameterized, output encoded? New dependencies vetted?
+5. **Performance** — N+1 query patterns? Unbounded loops or unconstrained fetches? Missing async where it matters? Unnecessary re-renders? Missing pagination?
+
+**Approval standard:** approve a change when it definitely improves overall code health, even if it isn't perfect. Don't block a change for not being exactly how you'd have written it.
 
 ## Output Format
 
--- a/agents/code-reviewer.md
+++ b/agents/code-reviewer.md
@@ -9,14 +9,14 @@ You are an experienced Staff Engineer conducting a thorough code review. Your ro
 
 ## Review Framework
 
-**Canonical source:** [skills/code-review-and-quality/SKILL.md](../skills/code-review-and-quality/SKILL.md) documents this framework in full — read it if you have access to this repo. The summary below keeps this persona self-contained for when it's copied out standalone into another tool (see `docs/copilot-setup.md`, `docs/gemini-cli-setup.md`).
+**Canonical source:** [skills/code-review-and-quality/SKILL.md](https://github.com/addyosmani/agent-skills/blob/main/skills/code-review-and-quality/SKILL.md) documents this framework in full — treat it as the source of truth and read it when reachable (in-repo or via this URL). The summary below is the self-contained fallback for standalone copies of this persona (see `docs/copilot-setup.md`, `docs/gemini-cli-setup.md`).
 
-Evaluate every change across five axes:
+Evaluate every change across five dimensions:
 
 1. **Correctness** — Does it do what the spec/task requires? Edge cases and error paths handled? Do the tests verify real behavior? Any race conditions, off-by-one errors, or state inconsistencies?
 2. **Readability** — Understandable without the author explaining it? Names clear and conventional? Control flow simple (no deep nesting)? Well-organized, no dead code?
-3. **Architecture** — Follows existing patterns, or justifies a new one? Module boundaries and dependency direction respected? Abstraction level appropriate (not over- or under-engineered)?
-4. **Security** — Input validated and sanitized at boundaries? No secrets in code, logs, or version control? Auth checked where needed? Queries parameterized, output encoded? New dependencies vetted?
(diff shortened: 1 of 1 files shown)


addyosmani (commented):
I like the instinct here. We do say "don't duplicate content between skills, reference and link instead," and the persona was carrying a near-verbatim copy of the five axes that live in `code-review-and-quality`. I checked and the skill genuinely is the fuller source (the five axes plus the approval standard), so pointing at it isn't losing anything for someone reading this inside the repo.

The problem is where these persona files actually get used. `docs/copilot-setup.md` tells people to `cp agents/code-reviewer.md .github/agents/code-reviewer.agent.md`, and `docs/gemini-cli-setup.md` says to copy this file's content into other tools. In every one of those flows the persona travels alone, and `../skills/code-review-and-quality/SKILL.md` resolves to nothing on the other side. So this trades duplication for a broken link in the persona's main distribution mode. That is the same portability tradeoff we're already tracking in #361, and for a file whose job is to be copied out, self-containment wins over DRY.

So I don't want to take this as-is. If you want to kill the duplication, the persona needs to stay usable standalone, which probably means inlining a compact version of the axes here with a "canonical source: code-review-and-quality" pointer for in-repo readers, rather than replacing the content with a link. Happy to look again if you reshape it that way. Thanks for spotting the duplication, the underlying observation is correct.

BryanDhaniel:
I hadn't considered that this persona file gets copied out standalone into other tools (per `docs/copilot-setup.md` / `docs/gemini-cli-setup.md`), so a pure link would break outside this repo. Updated to inline a compact version of the five axes, with a canonical-source pointer back to `code-review-and-quality` for anyone reading this in-repo.

Worth flagging honestly: this cuts the duplication but doesn't fully eliminate it, the axis descriptions are still hand-maintained in two places, just more concisely now. A complete fix would probably mean generating this section from the skill rather than hand-syncing it, but that felt like a bigger change than this issue called for. Let me know if this looks right.

agents/code-reviewer.md
```diff
@@ -9,40 +9,17 @@ You are an experienced Staff Engineer conducting a thorough code review. Your ro
...
-- Are queries parameterized? Is output encoded?
-- Any new dependencies with known vulnerabilities?
-
-### 5. Performance
-- Any N+1 query patterns?
-- Any unbounded loops or unconstrained data fetching?
-- Any synchronous operations that should be async?
-- Any unnecessary re-renders (in UI components)?
-- Any missing pagination on list endpoints?
+**Canonical source:** [skills/code-review-and-quality/SKILL.md](../skills/code-review-and-quality/SKILL.md) documents this framework in full — read it if you have access to this repo. The summary below keeps this persona self-contained for when it's copied out standalone into another tool (see `docs/copilot-setup.md`, `docs/gemini-cli-setup.md`).
```
federicobartoli: This line was added to solve the portability concern, but it breaks in exactly the scenario it describes: when the persona is copied standalone (per docs/copilot-setup.md / docs/gemini-cli-setup.md), the relative ../skills/... link resolves to nothing , same failure class as #361, which #366 fixes by pointing shared files at canonical GitHub URLs. "read it if you have access to this repo" has the same problem in prose form: in-repo the condition is always true (dead words), standalone it's always false ,right when a URL would have made the canonical source reachable anyway. Suggest:

```suggestion
**Canonical source:** [skills/code-review-and-quality/SKILL.md](https://github.com/addyosmani/agent-skills/blob/main/skills/code-review-and-quality/SKILL.md) documents this framework in full — treat it as the source of truth and read it when reachable (in-repo or via this URL). The summary below is the self-contained fallback for standalone copies of this persona (see `docs/copilot-setup.md`, `docs/gemini-cli-setup.md`).
```

Note: #366 is landing this exact convention for references/* (currently blob/main; if that PR switches to raw.githubusercontent.com, match it here) , and its CONTRIBUTING note should probably cover agents/ personas too.

agents/code-reviewer.md
```diff
@@ -9,40 +9,17 @@ You are an experienced Staff Engineer conducting a thorough code review. Your ro
...
+
+Evaluate every change across five axes:
+
+1. **Correctness** — Does it do what the spec/task requires? Edge cases and error paths handled? Do the tests verify real behavior? Any race conditions, off-by-one errors, or state inconsistencies?
+2. **Readability** — Understandable without the author explaining it? Names clear and conventional? Control flow simple (no deep nesting)? Well-organized, no dead code?
+3. **Architecture** — Follows existing patterns, or justifies a new one? Module boundaries and dependency direction respected? Abstraction level appropriate (not over- or under-engineered)?
```
federicobartoli: The compression drops "Any circular dependencies?" from the old Architecture section , it's a cheap, concrete check worth keeping: e.g. "Module boundaries, dependency direction, no circular dependencies?"

agents/code-reviewer.md
```diff
@@ -9,40 +9,17 @@ You are an experienced Staff Engineer conducting a thorough code review. Your ro
...
-
-### 5. Performance
-- Any N+1 query patterns?
-- Any unbounded loops or unconstrained data fetching?
-- Any synchronous operations that should be async?
-- Any unnecessary re-renders (in UI components)?
-- Any missing pagination on list endpoints?
+**Canonical source:** [skills/code-review-and-quality/SKILL.md](../skills/code-review-and-quality/SKILL.md) documents this framework in full — read it if you have access to this repo. The summary below keeps this persona self-contained for when it's copied out standalone into another tool (see `docs/copilot-setup.md`, `docs/gemini-cli-setup.md`).
+
+Evaluate every change across five axes:
```
federicobartoli: Tiny consistency nit: this file's own frontmatter description and SKILL.md both say "dimensions" : keep "five dimensions" so the persona and its canonical source use the same vocabulary.

agents/code-reviewer.md
```diff
@@ -9,40 +9,17 @@ You are an experienced Staff Engineer conducting a thorough code review. Your ro
...
+
+1. **Correctness** — Does it do what the spec/task requires? Edge cases and error paths handled? Do the tests verify real behavior? Any race conditions, off-by-one errors, or state inconsistencies?
+2. **Readability** — Understandable without the author explaining it? Names clear and conventional? Control flow simple (no deep nesting)? Well-organized, no dead code?
+3. **Architecture** — Follows existing patterns, or justifies a new one? Module boundaries and dependency direction respected? Abstraction level appropriate (not over- or under-engineered)?
+4. **Security** — Input validated and sanitized at boundaries? No secrets in code, logs, or version control? Auth checked where needed? Queries parameterized, output encoded? New dependencies vetted?
```
federicobartoli: "New dependencies vetted?" is vaguer than the old "known vulnerabilities" .. suggest "New dependencies vetted (no known vulnerabilities)?" so the actionable criterion survives the compression.

federicobartoli (changes requested):
Thanks @BryanDhaniel ! The compact-summary direction is right and the approval standard matches the skill's wording. One required change: the canonical-source pointer must survive standalone copies (GitHub URL instead of the relative link, and drop the "if you have access to this repo" conditional , see line comment), otherwise this reintroduces the exact portability problem the second commit set out to fix. Plus three small wording nits inline. Request changes and thanks again ! Happy to approve right after.

agents/code-reviewer.md
BryanDhaniel: You're right, this reintroduced the exact problem the second commit was trying to fix. Switched to the absolute GitHub URL and dropped the conditional phrasing so it reads correctly either way. Thanks!

agents/code-reviewer.md
BryanDhaniel: Fixed, thanks for catching the inconsistency.

agents/code-reviewer.md
BryanDhaniel: Fixed, restored the "no known vulnerabilities" qualifier so it stays concrete. Thanks!

agents/code-reviewer.md
BryanDhaniel: Fixed, restored "no circular dependencies" to the Architecture point. Thanks!

federicobartoli (approved):
LGTM

PR #428, opened 2026-07-26
fix(validator): parse frontmatter strictly, scope cross-refs to prose (#387)

ayobamiseun:
Takes the two items @kevglynn left available on #387, per @addyosmani's triage (bugs to fix, not contract to document) and following #405's pattern.

**Frontmatter:** `parseFrontmatter` now parses the sanctioned YAML subset — block scalars with chomping, comments, quoted scalars — and errors on what it previously accepted silently: malformed lines, unterminated quotes, and non-string `name`/`description` values. Multiline descriptions are length-measured on the assembled string.

**Cross-references:** prose patterns run against fence-stripped content so fenced examples can't produce dead-reference warnings; the diagram-arrow patterns still read full content since lifecycle diagrams live in fences. Added the styles the corpus uses that the patterns missed (capitalized See/Follow, bare ``Follow `name` ``, relationship bullets, `skills/<name>/SKILL.md` paths, plural `skills`), hyphen-gated against false positives.

Output is byte-identical on the current 24-skill corpus. New unit tests (`node --test scripts/skill-lint-test.js`, 18 passing) assert the decided behavior so #324's battery can build on it. Suite green: 24 skills, 8 commands, 124 eval checks, rank-1 86%, run-evals tests 12/12.

### DIFF
--- a/scripts/lib/skill-lint.js
+++ b/scripts/lib/skill-lint.js
@@ -10,6 +10,8 @@
  * Checks (errors block CI):
  *   - SKILL.md exists in every skill directory
  *   - YAML frontmatter present with 'name' and 'description' fields
+ *   - frontmatter parses cleanly (block scalars and comments supported;
+ *     malformed lines and non-string name/description values are errors)
  *   - frontmatter 'name' matches the directory name
  *   - directory name is lowercase-hyphen-separated (skill-anatomy.md: Naming Conventions)
  *   - description does not exceed 1024 characters
@@ -62,15 +64,29 @@ const SECTION_EXEMPT_SKILLS = {
 // Regex patterns that indicate an explicit cross-skill reference.
 // Only these patterns trigger the dead-reference warning — generic
 // backtick strings in code blocks are intentionally excluded.
-const SKILL_REF_PATTERNS = [
-  /\buse the `([a-z][a-z0-9-]+[a-z0-9])` skill/g,
-  /\bfollow the `([a-z][a-z0-9-]+[a-z0-9])` skill/g,
-  /\binvoke the `([a-z][a-z0-9-]+[a-z0-9])` skill/g,
-  /\bcontinue with `([a-z][a-z0-9-]+[a-z0-9])`/g,
-  /\buse `([a-z][a-z0-9-]+[a-z0-9])` skill/g,
-  /`([a-z][a-z0-9-]+[a-z0-9])` skill\b/g,
+//
+// PROSE patterns are matched against fence-stripped content, so example
+// references inside code blocks never produce dead-reference warnings.
+// DIAGRAM patterns are matched against the full content because the ASCII
+// lifecycle diagrams they target live inside fenced blocks.
+//
+// The riskier prose patterns (bold bullets, paths, bare see/follow) capture
+// only hyphenated names — every skill in the catalog is multi-word, and the
+// hyphen requirement keeps single backticked words like `main` from matching.
+const PROSE_REF_PATTERNS = [
+  /\b[Uu]se the `([a-z][a-z0-9-]+[a-z0-9])` skill/g,
+  /\b[Ff]ollow the `([a-z][a-z0-9-]+[a-z0-9])` skill/g,
+  /\b[Ii]nvoke the `([a-z][a-z0-9-]+[a-z0-9])` skill/g,
+  /\b[Cc]ontinue with `([a-z][a-z0-9-]+[a-z0-9])`/g,
+  /\b[Uu]se `([a-z][a-z0-9-]+[a-z0-9])` skill/g,
+  /`([a-z][a-z0-9-]+[a-z0-9])` skills?\b/g,
   /`([a-z][a-z0-9-]+[a-z0-9])` persona\b/g,
-  /\bsee `([a-z][a-z0-9-]+[a-z0-9])`/g,
+  /\b[Ss]ee `([a-z0-9]+(?:-[a-z0-9]+)+)`/g,
+  /\b[Ff]ollow `([a-z0-9]+(?:-[a-z0-9]+)+)`/g,
+  /^\s*[-*] \*\*`([a-z0-9]+(?:-[a-z0-9]+)+)`/gm,   // relationship bullets: - **`skill-name`**: …
+  /\bskills\/([a-z0-9]+(?:-[a-z0-9]+)+)\/SKILL\.md/g,
+];
+const DIAGRAM_REF_PATTERNS = [
   /──→ ([a-z][a-z0-9-]+[a-z0-9])\b/g,          // ASCII diagram arrows
   /→ `([a-z][a-z0-9-]+[a-z0-9])`/g,
 ];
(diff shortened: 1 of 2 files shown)


addyosmani (commented):
Good, careful work. Replacing the naive colon-split parser with one that actually understands the YAML subset we sanction (block scalars, comments, quoted scalars) and reports malformed frontmatter as a problem instead of silently skipping it is the right direction, and it's the honest completion of the #387 hardening. Splitting the reference patterns into prose (matched fence-stripped) versus diagram (matched against full content) is the correct fix for the example-refs-in-code-blocks false positives, and the hyphen requirement keeping single words like `main` from matching is a nice touch. CI is green across the board.

I checked one interaction that worried me: #429 introduces an `allowed-tools` frontmatter line with a fairly gnarly value (parens, `${...}`, commas, a `*`), and your stricter parser handles it cleanly, no false problem. So those two don't collide, good.

One thing I want to flag rather than block on: this is the fourth change in flight touching skill reference/lint handling. #379 extracted this module, #405 hardened it, #427 is reworking a persona's skill reference, and now this is a 297-line rewrite of the same file. That's exactly the collision surface CLAUDE.md warns about for shared `scripts/` files. Since @nucliweb owns the #379 extraction, I'd like his eyes on this before merge so the ordering and the module's shape stay coherent. @nucliweb, can you take a look? Substance-wise I'm happy with it.

ujjwalks:
Found a bug in `stripFencedCodeBlocks` while looking at something adjacent. It matters more on this branch than on `main`, because your cross-reference change now routes through it — so I built the fix on top of your head rather than opening a competing PR against the same function.

The helper matches one regex over the whole document:

```js
```

That only recognizes a ```-fence opening at column zero and closing with a run of backticks of exactly the same length. Three CommonMark-legal forms slip through:

- tilde fences (`~~~`)

Their contents stay in the prose both rules read. Two consequences, the second one squarely in this PR's lane:

- a `## Overview` inside a `~~~` example satisfies the Overview requirement for a skill that has no Overview;
- a fenced ``see `made-up-skill` `` inside a tilde or indented fence still produces a dead-reference warning — the exact case `references inside fenced code blocks are not collected` is asserting against, just in a fence form the regex doesn't see.

Fix is a line-by-line scan tracking the open fence's marker and length. An unterminated fence now runs to end of file, which surfaces as missing-section errors instead of prose that silently passes.

Rebased on `0980192`, one commit, two files — `skill-lint.js` and 13 cases appended to your `scripts/skill-lint-test.js`:

https://github.com/ujjwalks/agent-skills/tree/fix/skill-lint-fence-parsing-on-428

6 of the 13 fail against the current helper (both tilde cases, both indented cases, backtick-doesn't-close-tilde, unterminated), including the two cross-reference ones. Full suite green with it: 31/31 in `skill-lint-test.js`, 24 skills, 8 commands, 124 eval checks at rank-1 86%, `run-evals-test.js` 12/12. Corpus output unchanged.

Cherry-pick `d76aa7f` if you want it in this PR, or say the word and I'll open it as a PR into your branch. Happy to drop it entirely if you'd rather keep this PR to the #387 scope — it stands alone against `main` too, just with a conflict to resolve.

One separate note: `scripts/skill-lint-test.js` isn't run by CI on this branch — `test-plugin-install.yml` runs `run-evals-test.js` but nothing runs yours. It's a one-line step. I left it out to keep this to the two files you're already touching, but it's yours if you want it.

kevglynn:
Worth calling out explicitly: the section-check half of @ujjwalks's find isn't specific to this branch — `stripFencedCodeBlocks` is also what #405 (merged) uses to strip fences before section-heading matching, so a `## Overview` inside a `~~~` example has been able to satisfy the Overview requirement on `main` since #405 shipped. Only the cross-reference half (dead-reference warnings surviving inside tilde/indented fences) is specific to this PR's new prose patterns.

Given @addyosmani already flagged this file as having four changes in flight and wants @nucliweb's eyes before merge, I split the `stripFencedCodeBlocks` fix out into its own issue rather than growing this PR's scope further: #437. That keeps this PR scoped to #387 (frontmatter + cross-ref patterns) and lets the `main`-branch bug get fixed on its own track, whichever PR ends up landing first.

abhisheksharma2411 (commented):
I came at this from the opposite direction — I'd been reading `skill-lint.js` and filed #505 for some of the same ground before finding this PR, which was a month older and more thorough. I've since reverted the overlapping parts of mine. So treat the following as someone who has spent a while in this exact function.

The parser is good. I exercised it rather than reading it, and it held up on everything I threw at it:

```
folded scalar with a colon in the continuation   -> []          (no bogus "no trigger")
literal scalar with keep chomping (|+)           -> []
full-line comment inside frontmatter             -> []
unterminated double quote                        -> "Frontmatter 'description' has an
                                                    unterminated double-quoted value"
```

That last one is the part I'd single out: reporting malformed frontmatter instead of silently skipping the line is the actual fix, and the message names the field and the fault. I also checked a description legitimately containing `#` (`Use when triaging issue #123`) in case trailing-comment stripping ate it — it doesn't. Good.

**One finding, in the new prose patterns rather than the parser.**

The relationship-bullet pattern

```js
/^\s*[-*] \*\*`([a-z0-9]+(?:-[a-z0-9]+)+)`/gm
```

treats any bold-backticked hyphenated term at the start of a bullet as a skill reference. In a repo about engineering practice, that shape is ordinary prose:

```markdown
- **`at-least-once`**: delivery guarantee, not a skill name
- **`write-ahead`**: a log, not a skill
```

```
Dead cross-reference: `at-least-once` is not a known skill
Dead cross-reference: `write-ahead` is not a known skill
```

The PR description says the riskier patterns are "hyphen-gated against false positives", but for this shape the hyphen gate is what *admits* them — `at-least-once`, `read-after-write`, `fail-open`, `dead-letter` are all hyphenated terms of art. "Byte-identical on the current 24-skill corpus" is true and I verified it, but it's a statement about today's content, and this pattern is new surface that future content will land on.

It's warning-level so nothing breaks today. I'd still narrow it, because the failure mode is a contributor un-backticking a correct technical term to silence a validator. Options, roughly in order of how much I'd like them:

1. Require the bullet to look like a cross-reference rather than a definition — e.g. only match when the line also contains `skill` or a link, so `- **\`code-review-and-quality\`** — use for …` matches but a glossary bullet doesn't.
2. Keep the pattern but only warn when the captured name is *close* to a known skill (edit distance, or shares a hyphen-separated token), so unknown vocabulary stays quiet.
3. Drop the catch-all arm and keep only the README-roster check, i.e. error on referencing a skill that's declared-but-absent and say nothing about unknown tokens.

For what it's worth I walked into precisely this on a validator in my own repo last week — a contributor's dead-reference check flagged `at-least-once`, `write-ahead` and `dead-letter` from one true sentence — so I'm confident about the shape of the problem rather than guessing at it.

Nothing here blocks the merge from my point of view; the frontmatter work is clearly correct and clearly needed.

**One thing this PR deliberately doesn't touch, flagged so it isn't lost:** `exempt = dirName in SECTION_EXEMPT_SKILLS` survives on your branch. `in` walks the prototype chain and `constructor` satisfies `KEBAB_CASE`, so a skill directory named `constructor` still validates clean with no sections at all (`exempt: true, errors: []`). That's all #505 is now — two lines and a test. Happy for it to be folded into this PR instead if you'd rather keep it in one place.

abhisheksharma2411:
Quick follow-up since the ground shifted under this one: #505 merged, which touched `skill-lint.js` and added a test file. I checked what that does to your branch rather than leave you guessing.

**You're fine.** I merged `main` into your head locally:

- `scripts/lib/skill-lint.js` auto-merged, **zero conflicts**.
- The exempt-lookup change survives the merge intact — worth confirming, because a clean textual merge can still quietly revert a fix when the other side has rewritten nearby logic, and your `parseFrontmatter` rewrite sits close to it. It didn't: `Object.hasOwn(SECTION_EXEMPT_SKILLS, dirName)` is still there at both call sites afterwards.
- Your own suite: **18/18**. The suite from main: **7/7**.

So nothing to do here. The one thing that did get simpler is the test-file question I raised in my review — it was three paths across three PRs, and now it's two: `scripts/lib/skill-lint-test.js` on main, and `scripts/skill-lint-test.js` here. Not a blocker and entirely your call, but if you'd rather have one file I'm happy to do the move as a follow-up after this merges, so it doesn't add churn to a PR that's already been open a while.

(#444 did get hit — same-path collision with #505 — so I've pushed a resolved branch over there. Mentioning it only so the three of us aren't each solving it separately.)

PR #429, opened 2026-07-26
feat: add repository-memory skill (v1 of #398)

federicobartoli:
Draft v1 of repository-memory, per the scope @addyosmani wrote down in #398. Opening it as a draft so the open calls below can be discussed on real lines; I'll run it against a real repository and post what it wrote and what it refused to write here before marking it ready.

What's inside, mapping to the scope bullets: one skill directory with the SKILL.md (two tiers: an append-only AGENT_LOG.md with admission rules only, and checker-gated AGENT_MEMORY.md capsules; six-condition promotion gate; checked retrieval with mandatory read-time corroboration and refuse-if-uncertain), a small check-memory.sh that stays a cost filter in @nucliweb's framing (structural validation, byte-identity against the trusted ref, anchor existence, path-existence staleness; exit 0 always, unknown never an error), a 36-fixture deterministic suite including the three mandated cases (anchor commit erased, non-git checkout, shallow clone, each resolving to unknown cleanly), and the eval case with three execution evals (no-op admission, degraded retrieval, positive admission). To verify:

    bash skills/repository-memory/scripts/test-check-memory.sh
    node scripts/run-evals.js

A few places where the build went slightly past the written bullets, each cheap to revert if v1 should stay smaller:

- The trusted ref must resolve to a remote-tracking ref (origin/main style); HEAD, local branches, tags and SHAs come back unknown. Without this, anything committed locally self-verifies, which looked like exactly the self-authored trust the computed state exists to prevent. Does that read right?
- A capsule that resolves to a different repository than the one the checker is anchored to (a vendored checkout carrying its own .git and its own fabricated origin/main) is refused. Same self-signing worry.
- The only staleness v1 detects is a declared path that no longer exists (literal check first, so [slug]-style filenames work; glob patterns fall back to a ls-files match re-tested on disk). Everything content-level stays with read-time corroboration. If the preference is the even smaller cut with no stale state at all, it's a ten-line removal.
- The frontmatter carries an allowed-tools line scoped to the checker script only, so Claude Code runs it without a permission prompt. It's inert on other runtimes, but it would be the first use in the catalog, so it's easy to drop if that door shouldn't be opened here. Ok to keep?
- The log file is AGENT_LOG.md for symmetry with AGENT_MEMORY.md, pending the naming call from the issue thread (@asiridalugoda's reference implementation uses EXPERIMENTS.md); renaming is a one-commit change.
- test-check-memory.sh is deterministic and self-contained but nothing wires it into CI yet; where should it live, if anywhere?


federicobartoli:
Ran it against a real repository, as promised. Setup: a clone of expressjs/session, with the whole catalog loaded from this branch rather than the skill in isolation.

One thing worth being explicit about: both prompts name the skill, so what this exercises is the gate, not the routing. That is deliberate. On the typo task especially, an agent that never loads the skill writes no memory trivially and proves nothing; naming it and then inviting it to persist something is the harder version of that test. Whether the description routes on its own is what the trigger evals in this PR are for.

Two tasks, opposite outcomes.

First one was a typo fix in a code comment. The skill declined to persist anything: the final diff is a single line in index.js, and there is no memory file anywhere in the tree. That is the case I most wanted to watch, since a memory system that writes on a typo fix is the failure mode rather than the feature.

Second was an investigation into what happens to req.session while the session store is disconnected. That one earned a capsule, 114 words of body:

<details> <summary>the AGENT_MEMORY.md it wrote</summary>
---
scope:
  - index.js
  - session/store.js
evidence:
  - index.js#session
  - session/store.js#Store
  - test/session.js
invalidated_by:
  - index.js
  - session/store.js
claims:
  - statement: While the store is disconnected the middleware calls next() with req.session left undefined and sends no Set-Cookie.
    verifier: test/session.js
---

# Store readiness gating in the session middleware

- `Store` only inherits `EventEmitter`; nothing in this repository ever emits
  `disconnect` or `connect`. Those events come solely from third-party store
  implementations, so the readiness path is dormant under `MemoryStore`.
- The `storeReady` flag lives in the `session()` closure, so one `disconnect`
  silently disables sessions for every request through that middleware
  instance until a `connect` arrives.
- The `!storeReady` early return happens before `req.sessionStore`,
  `req.sessionID`, the `onHeaders` set-cookie hook, and the `end` proxy are
  installed. All three `req` properties stay undefined and nothing is saved or
  sent, yet the request itself proceeds, so failures surface downstream in
  handlers that assume `req.session` exists, not in the middleware.
</details>
The bullet I did not expect is the first one: it noticed that Store only inherits EventEmitter and that nothing in the repository ever emits those events, so the whole readiness path is dormant under MemoryStore and only wakes up with a third-party store. I corroborated it afterwards and it holds (index.js only listens, store.js only inherits). That is the kind of thing that costs twenty minutes to rediscover and does not appear in any README.

Then the checker on that same capsule:

$ check-memory.sh --trusted-ref origin/master AGENT_MEMORY.md
{"path": "AGENT_MEMORY.md", "state": "proposed", "usable": false, "reason": "not-on-trusted-ref"}
which is the loop closing the way it should: the agent proposes, review promotes, and until then no future agent gets to use it.

There is also a short video walkthrough, in case it is easier to watch than to read: [LINK](https://youtu.be/cLRwnFpd51I?si=3CmIDbKgRd3a39Fz) It does not start with the agent. It starts in the terminal on a small fixture repo, walking through the checker states one at a time: a vendored checkout carrying its own fabricated origin/main and trying to pass itself off as reviewed, then a capsule that verifies, then that same capsule turning proposed after a single unreviewed line, stale once its evidence file is deleted, and refused outright when the trusted ref is a local branch. Only after that does it move to the express-session run above, and it closes on the fixture suite.

Fair warning on the English: it is the first video I have ever recorded, and also an excuse to practice a language I had not spoken in a long time, so it is rougher than the code.

asiridalugoda:
Good stuff @federicobartoli.

One thing to take a closer look at: invalidated_by never actually fires on a content change. Stage 5 only checks that declared paths exist, not whether they changed since the capsule was reviewed. So if evidence files get modified in any commit after the capsule’s anchor, the capsule stays verified. Every stale transition in the walkthrough is a deletion.

Concrete repro against your own best capsule: merge a change that makes store.js actually emit one of those dormant events. The capsule’s headline claim is now false, store.js still exists, and the checker still says verified.

The checker already resolves the anchor commit, so the fix looks straightforward:

```
git log --max-count=1 <anchor>..<trusted-ref> -- <evidence + invalidated_by paths>
```


returns anything → stale. That enforces the #398 definition of verified (invalidation paths unchanged since review). One case this can’t catch: capsule and evidence edits landing in the same squash commit. Rest looks all good 👍

federicobartoli:
> Good stuff @federicobartoli.
> 
> One thing to take a closer look at: invalidated_by never actually fires on a content change. Stage 5 only checks that declared paths exist, not whether they changed since the capsule was reviewed. So if evidence files get modified in any commit after the capsule’s anchor, the capsule stays verified. Every stale transition in the walkthrough is a deletion.
> 
> Concrete repro against your own best capsule: merge a change that makes store.js actually emit one of those dormant events. The capsule’s headline claim is now false, store.js still exists, and the checker still says verified.
> 
> The checker already resolves the anchor commit, so the fix looks straightforward:
> 
> ```
> git log --max-count=1 <anchor>..<trusted-ref> -- <evidence + invalidated_by paths>
> ```
> 
> returns anything → stale. That enforces the #398 definition of verified (invalidation paths unchanged since review). One case this can’t catch: capsule and evidence edits landing in the same squash commit. Rest looks all good 👍

Thanks, and you are right. I reproduced it on a throwaway repo before answering: capsule verified, then the declared evidence file changes content in a commit on the trusted ref, capsule still verified. And yes, every stale transition in the walkthrough and in the fixtures is a deletion.

Two things worth separating.

The behaviour is deliberate for v1. The SKILL.md says the checker does not track whether declared dependencies changed content between commits, and that came out of the scoping call in #398 to ship without a dependency-tracking freshness engine and add one only if staleness actually bit in practice. Your repro is the first concrete case of it biting, so I am not going to wave it off as by design.

The naming is where you have me. #398 defines verified as no invalidation path having changed since the capsule was last modified, and I kept that word while implementing a weaker condition. Practically, files get edited far more often than they get deleted, so the state as shipped catches the rare case and misses the common one. That leaves two coherent positions, strengthen stale or drop it altogether, and the middle one I shipped is the weakest of the three.

Your one-liner reuses the anchor the checker already resolves, so it is nowhere near the engine that got deferred. Four limits I would rather state up front than discover later: it only sees changes that landed on the trusted ref, while the original #398 algorithm also compared against HEAD and the working tree; declared paths can be globs and git pathspec semantics are not shell glob semantics; keeping the set to evidence and invalidated_by rather than including scope matters, since a scope glob would make every commit in the subsystem invalidate the capsule, and I see you already wrote it that way; and the squash case you flagged stays uncatchable, which belongs in the docs next to the incomplete-invalidated_by limitation.

@addyosmani this is a scoping call rather than mine to make. If you want it in v1, I will implement it as described plus the fixture the suite is missing, a declared path whose content changed after the anchor. If you would rather hold the line at the cheap version, then I would narrow what verified claims in the docs, so it stops implying the #398 definition.

addyosmani (approved):
This is exactly the v1 we scoped in #398, and it's the strongest single contribution the repo has taken. I went through it properly rather than skimming, because a trust checker is the one place where "looks right" isn't enough.

What I verified locally:
- The checker's own suite is 36/36 green, and it covers the cases that actually matter: byte-identity against the trusted ref, the rebase/squash anchor-erased case resolving to `unknown` (not `verified`), non-git and shallow-clone degradation, HEAD/local-branch/tag/SHA refused as trusted refs, and the vendored self-signed capsule refused. The two hard fixtures I asked for in the thread both exist and pass.
- Full CI reproduced on the merged tree: validate-skills, validate-commands, run-evals-test, and the eval routing all green. Rank-1 dips 86% to 85%, which is just the new skill's four positives landing in the pool and still well clear of the 80% floor. No collision with `context-engineering` despite the semantic neighbourhood, so the description is pulling its weight.
- The eval case is over the minimum (four positives, three owner-tagged negatives, three execution evals each backed by its own fixture).

Beyond faithfulness to the scope, two design calls are sharper than what the thread landed on and both are right: the trusted ref must be a remote-tracking boundary because anything committable locally would self-verify, and a capsule resolving to a nested repo is refused because it could self-sign. The honest-degradation contract (unknown, never error, exit 0 always) is stated in the code and enforced by fixtures. The separate log vs capsule files for a mechanical reason (per-file byte-identity would break on every append) is a genuinely good call.

Approving. One thing before it merges, and it's a compliment not a hold: this is a security-relevant trust boundary and @nucliweb shaped its core framing (cost-filter-not-trust-mechanism, the #385 degradation lesson), so I'd like his eyes on the checker specifically before we land it. @nucliweb, can you give check-memory.sh a pass? If you're happy with it, it's ready.

addyosmani:
@nucliweb when you have a window, could you give `check-memory.sh` a focused pass before we merge this? It's the one security-relevant piece here (the trust boundary and the honest-degradation contract), and you shaped its framing in #398, so you're the right second set of eyes. I've verified it end to end already: the checker's own suite is 36/36 green, full CI reproduces on the merged tree, and the rebase/squash anchor-erased and vendored-repo self-sign cases both resolve to `unknown` as intended. I'm happy with it; I just want your read on the checker specifically, then it lands. Everything else in the PR I've cleared.

nucliweb (commented):
@addyosmani here is the focused pass on `check-memory.sh` you asked for. Everything below is runtime against `9b98e1d` on macOS with the system `/bin/bash` (3.2.57), not a read of the diff.

## The contract holds

The suite is 36/36 green, and bash 3.2 being the interpreter is worth noting on its own: the array handling under `set -u` is the usual place a script like this breaks on macOS, and it does not.

I probed the honest-degradation clause with inputs the fixtures do not cover, and it behaves:

| Input | Result |
|---|---|
| 3 MB of `/dev/urandom` named `.md` | `unknown` / `invalid-metadata`, exit 0 |
| 5,000-line frontmatter | `unknown` / `invalid-metadata`, exit 0 (the 400-line bound fires) |
| `evidence: - "src/$(touch /tmp/PWNED).js"` | path stored literally, **no execution**, exit 0 |
| `evidence: - src/\`touch /tmp/PWNED2\`.js` | path stored literally, **no execution**, exit 0 |
| caller outside any git repo | `unknown`, fails closed |

Exit code was 0 in every case, exactly one line per capsule, and I parsed every emitted line with a real JSON parser rather than eyeballing it: all valid, including the ones carrying backticks and `$(...)` inside the `paths` array. The trust boundary also holds where it matters: `HEAD`, local branches and tags are refused, and the vendored self-signing case resolves to `unknown`.

## One finding: `invalidated_by` is inert as a distinct field

This is @asiridalugoda's point from July, and I want to add the part that I think makes it merge-relevant rather than a v2 nicety: **`invalidated_by` and `evidence` are behaviourally identical in v1.**

Two capsules, the only difference being which section `src/dep.js` is declared under:

```
dep.js CHANGED on the trusted ref:
  A (invalidated_by): {"state": "verified", "usable": true,  "reason": "ok"}
  B (evidence)      : {"state": "verified", "usable": true,  "reason": "ok"}

dep.js DELETED:
  A (invalidated_by): {"state": "stale", "usable": false, "reason": "declared-path-missing"}
  B (evidence)      : {"state": "stale", "usable": false, "reason": "declared-path-missing"}
```

Both sections feed the same `declared` accumulator and are only ever existence-tested, so declaring a path under `invalidated_by` buys nothing over declaring it under `evidence`.

The reason I do not think the code comment at lines 23 to 25 settles it: the comment scopes the deferral, but `SKILL.md:95` still tells the user the field takes "files or globs **whose change** may falsify the capsule". A user reading that will list their volatile dependencies there and believe they are covered. Reproduced with your own store.js example:

```
capsule claim: "Store emits no events; callers must poll."
invalidated_by: src/store.js

-> store.js rewritten to extend EventEmitter, committed, origin/main updated
-> checker: {"state": "verified", "usable": true, "reason": "ok"}
```

The capsule's headline claim is false and it is still `usable: true`. Deletion is the only transition that fires.

## The fix looks cheap, and I tested it

I prototyped @asiridalugoda's suggestion, scoped to `invalidated_by` only (leaving `evidence` as navigation, per the schema wording). It is about 14 lines: a second accumulator during the parse, and one stage after the anchor resolves.

```sh
  # 4b. Declared dependencies must not have changed since the anchor commit.
  local changed=() pi
  while IFS= read -r pi; do
    [ -n "$pi" ] || continue
    if [ -n "$(git -C "$toplevel" log -n 1 --format=%H "$anchor..$TRUSTED_REF" -- "$pi" 2>/dev/null)" ]; then
      changed+=("$pi")
    fi
  done <<EOF_INVAL
$inval
EOF_INVAL
  if [ "${#changed[@]}" -gt 0 ]; then
    emit "$1" stale false invalidated-by-changed "${changed[@]}"
    return 0
  fi
```

Results on my patched copy: the store.js case flips to `{"state": "stale", "reason": "invalidated-by-changed", "paths": ["src/store.js"]}`, and **the existing suite stays 36/36 green**, so no fixture encodes the current behaviour as intended. It stays a cost filter too: one `git log -n 1` per declared dependency, and it can only ever move a capsule toward refusal.

## What I would do

Either implement it, or drop `invalidated_by` from the v1 schema and reintroduce it when it does something. What I would avoid is shipping the field with documentation promising change detection that is not wired up, because that is the one failure shape where a capsule passes the cheap filter *and* reads as trustworthy. It does not break my framing from #398 (corroboration is still what makes a surviving capsule true), but `verified` plus `usable: true` is precisely the signal that buys a reader out of paying for corroboration, so a false `verified` is the expensive one.

If you would rather keep v1 minimal, the doc edit alone unblocks the merge for me: change `SKILL.md:95` to say deletion, and note that content drift is left to read-time corroboration.

## Nits, non-blocking

- When the caller is outside any git repo, every capsule reports `outside-repo-root`, though the actual cause is that `EXPECTED_ROOT` came back empty at line 288. The behaviour is right (fails closed) and `--repo-root` is the documented way through; only the reason string misleads.
- `validate_path` rejects a literal `#` in declared paths, which `SKILL.md:98` documents. Worth a `red flag` line in the skill too, since a capsule silently becoming `invalid-metadata` over a filename is a confusing way to find out.

Happy to open the 4b change as a PR against your branch if you want it, or to re-run anything here against a different bash.

PR #430, opened 2026-07-27
debugging-and-error-recovery: 'Getting Unstuck' escalation ladder + cheap pre-reduction checks

ElijahLynn:
Hi, I'm Elijah, thank you very much for the debugging skill and all the skills actually. I'm using them on a project and about to ship my first product ever, they have made product engineering with my first product a really great experience (as a full stack engineer)!!!

I don't know if this is the right place but I made these changes to the skill that I found useful additions, as I forked the skill locally. Or maybe they can be included in an include skill? Feel free to close this issue if not the right vibe for this skill though. Figured I would at least put this out here. Thanks!

Below is from Claude:

---

**What this adds and why it earned its place.** The existing skill covers the triage loop well (reproduce → localize → reduce → fix → guard → verify), but it has no answer for the moment the loop *stalls* — hypotheses keep dying, the failure won't reproduce, every probe comes back ambiguous. That's where we lived for most of a real 10-hour debugging session this week, and this PR is that night distilled into two additions:

1. **"Cheap Checks Before Reducing"** — four nearly-free checks (already-captured artifacts, system logs, version delta, pinned library source) that can each end an investigation before the expensive minimal-repro step.
2. **"Getting Unstuck — the escalation ladder"** — six deliberate escalations for a stalled loop: search the issue tracker *directly* with `gh` (our root-cause issue was 25 hours old and invisible to web search), read the *installed* source, instrument the error before theorizing about it (we built four wrong theories on an error message that the tooling itself had blanked), micro-reproductions built from the suspected mechanism (an 8-minute coin-flip repro became a 20-second deterministic one), A/B candidate fixes against the installed dist, and fresh adversarial review of your own conclusions. Plus standing instrument rules: positive controls before trusting negatives, postconditions over exit codes, whole logs never tails, and a closed-leads ledger.

Every rung was used, in anger, in one session — the public case study (root cause found upstream in `workers-sdk`, 20s reproducer, shipped mitigation, full methodology) is here: https://github.com/cloudflare/workers-sdk/issues/14641#issuecomment-5087174647

The additions are tool-agnostic apart from the `gh` examples, and the diff leaves the existing skill text untouched — both sections are purely additive. Happy to reshape (separate skill, include, trims) however fits the repo's direction.

### DIFF
--- a/skills/debugging-and-error-recovery/SKILL.md
+++ b/skills/debugging-and-error-recovery/SKILL.md
@@ -73,8 +73,7 @@ Cannot reproduce on demand:
     └── Document the conditions observed and revisit when it recurs
 ```
 
-For test failures:
-
+For test failures (npm shown — substitute the repository's own test command, per the test-driven-development skill's Discover the Stack First section):
 ```bash
 # Run the specific failing test
 npm test -- --grep "test name"
@@ -101,14 +100,13 @@ Which layer is failing?
 ```
 
 **Use bisection for regression bugs:**
-
 ```bash
 # Find which commit introduced the bug
 git bisect start
 git bisect bad                    # Current commit is broken
 git bisect good <known-good-sha> # This commit worked
 # Git will checkout midpoint commits; run your test at each
-git bisect run npm test -- --grep "failing test"
+git bisect run npm test -- --grep "failing test"  # substitute the repository's focused-test command
 ```
 
 A third-party runtime or library in the failure path, or a symptom that resists reduction, is exactly when the cheap checks in [Cheap Checks Before Reducing](#cheap-checks-before-reducing) pay off — read them before Step 3's more expensive work.
@@ -158,7 +156,7 @@ This test will prevent the same bug from recurring. It should fail without the f
 
 ### Step 6: Verify End-to-End
 
-After fixing, verify the complete scenario:
+After fixing, verify the complete scenario with the repository's own commands (npm shown):
 
 ```bash
 # Run the specific test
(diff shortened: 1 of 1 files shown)


ElijahLynn:
Provenance for these additions, kept out of the skill text itself: they were distilled from a real 10-hour multi-machine flake hunt that ended in an upstream root cause, a 20-second reproducer, and a shipped fix — public writeup with the full methodology: https://github.com/cloudflare/workers-sdk/issues/14641#issuecomment-5087174647. Each rung of the "Getting Unstuck" ladder earned its place that night.

addyosmani (changes requested):
There's real expertise in here and I don't want that to get lost in the feedback: "check the logs from the process outward," a dead server as a top-tier hypothesis for ECONNRESET, "instrument to distinguish hypotheses not confirm one," validate the instrument before trusting a negative, the CLOSED-LEADS ledger. That's hard-won debugging craft and most of it belongs in this skill.

But I can't take it as it stands, for a few concrete reasons:

1. Dangling reference. It says "Per AGENTS.md's 'Tooling discovery' section, `opensrc` ...", and there is no such section in AGENTS.md (I grepped). So it points readers at guidance that doesn't exist, and it hard-recommends one specific third-party tool by name, which cuts against how vendor-neutral the rest of the pack stays. Either drop the `opensrc` specifics down to the capability ("resolve and grep the pinned package source") or land the AGENTS.md section first and reference it for real.

2. It re-hardcodes npm right after we de-hardcoded it. #425 just finished making this exact skill ecosystem-neutral as part of the #404 pass, and this adds fresh `npm ls` / `npm view` / `npm outdated` and `node_modules` / `pnpm patch` references. One spot nods at "pip show, cargo tree," but the rest assumes Node. Please rebase on current main and carry the same "substitute your ecosystem's command" framing the skill now uses everywhere else.

3. Markdown bug: the last bullet of the escalation ladder runs straight into `## Common Rationalizations` with no blank line, so that heading won't render in stricter parsers. Quick fix.

4. Altitude. This is +117 in one shot and pushes the skill to ~417 lines with dense bash blocks that read more like a reference than a process. Worth considering whether the cheap-checks command listings and parts of the ladder move into a `skills/debugging-and-error-recovery/references/` supporting file, with the SKILL.md keeping the decision spine. Not a hard requirement, but skill-anatomy leans that way past 100 lines of reference material.

The bones are good. Fix 1 to 3 and let's talk about 4, and there's a strong addition here.

PR #439, opened 2026-07-31
feat(qoder): add skills setup and plugin manifest

phodal:
## Summary

- add an explicit `.qoder-plugin/plugin.json` with Qoder plugin identity, version, and the canonical `skills/` entrypoint
- add a dedicated Qoder setup guide covering project and global Skills CLI installation, manual installation, invocation, validation, and troubleshooting
- add Qoder to the README integration list and repository layout

## Why

Qoder supports the same `SKILL.md` format and is a first-class target in the Skills CLI, but this repository did not document that path or carry explicit Qoder plugin metadata. A local checkout validated structurally, yet appeared without a version in Qoder's plugin inventory.

This change makes the supported contract explicit: all 24 workflows under `skills/` are installable as Qoder skills. Host-specific command packages remain outside that portability claim and are called out in the guide.

## Validation

- `qodercli plugin validate .`
- `qodercli plugin list --plugin-dir "$PWD" --json` — reports `agent-skills@flag` version `1.0.0`
- isolated `npx skills add /path/to/agent-skills -a qoder -y --copy` — installed 24/24 skills under `.qoder/skills/`
- `node scripts/validate-skills.js` — 24 skills passed
- `node scripts/validate-commands.js` — 8 commands passed
- `node scripts/run-evals-test.js` — 12 tests passed
- `git diff --check`

### DIFF
--- a/.qoder-plugin/plugin.json
+++ b/.qoder-plugin/plugin.json
@@ -1,0 +1,31 @@
+{
+  "name": "agent-skills",
+  "displayName": "Agent Skills",
+  "version": "1.0.0",
+  "description": "Production-grade engineering skills for AI coding agents covering the full software development lifecycle from spec to ship.",
+  "author": {
+    "name": "Addy Osmani",
+    "url": "https://github.com/addyosmani"
+  },
+  "homepage": "https://github.com/addyosmani/agent-skills",
+  "repository": "https://github.com/addyosmani/agent-skills",
+  "license": "MIT",
+  "keywords": [
+    "qoder-plugin",
+    "agent-skills",
+    "engineering",
+    "spec",
+    "tdd",
+    "review",
+    "ship"
+  ],
+  "category": "Coding",
+  "tags": [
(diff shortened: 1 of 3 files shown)


addyosmani (commented):
Clean, standard integration and the setup doc follows the same shape as the other tool guides, thanks. Two things before this can land, one of which is not really about your PR.

First, CI hasn't run here yet (state is unstable, no checks), so I'd want a green run before merging.

Second, and this is the one to coordinate on: this adds a sixth manifest carrying `version: 1.0.0`, which is exactly the static value #440 just documented as breaking update detection across the existing manifests. I don't want to merge another `1.0.0` manifest until #440's version-metadata approach is settled, or this just widens the problem by one more file. Once we decide whether manifests track releases or stay static, this slots in behind that. Holding it on those two, not on the integration itself. Separately, adding a new tool target is a curation call I'll make, since each manifest is one more thing to keep in sync.

PR #441, opened 2026-07-31
Add a likelihood axis to review finding triage

### LINKED ISSUE #436: Review surfaces rank findings by consequence only — no likelihood axis, so improbable findings block merges
## The gap

Every review surface in this repo classifies findings by **consequence** — what happens *if* the issue fires. None of the general ones ask **how likely it is to fire at all**. So a finding that is true, and would be bad, but needs a conjunction of conditions that will never occur in the project, is indistinguishable from one that bites on the first realistic input.

This is the dominant failure mode of LLM reviewers in practice. Both GPT- and Claude-family models will produce a near-endless supply of technically-correct findings whose triggering conditions are absurd — "a second worker could observe a torn snapshot" against a design that pins one consumer, "this breaks at 10M rows" against an internal tool with 400. The reviewer is not wrong. The finding is still waste: the author has to read it, evaluate it, and dismiss it, and under `REQUEST CHANGES` semantics it can block a merge.

Neither existing axis can filter these:

- **Severity can't.** Severity is defined by blast radius, and a rare bug is still severe *if* it fires. `agents/code-reviewer.md:51` defines Critical as "security vulnerability, data loss risk, broken functionality" — all consequence, no probability.
- **Confidence can't.** Confidence answers "is my claim true?". These findings score high on it. A finding can be 0.95 certainly-true and 0.05 will-ever-happen at the same time; one number cannot carry both.

## Evidence (all citations at `7829ffd`)

| Surface | How findings are ranked | Likelihood handling |
|---|---|---|
| `agents/code-reviewer.md:51-55` | Critical / Important / Suggestion | **none** |
| `skills/code-review-and-quality/SKILL.md:181-187` | Critical / Nit / Optional / FYI prefix table | **none** |
| `skills/doubt-driven-development/SKILL.md:172-177` | contract-misread / valid+actionable / valid-trade-off / noise | **none** (see below) |
| `agents/security-auditor.md:57-65, 99` | severity by exploitability | informal, prose only |
| `agents/web-performance-auditor.md:41-51` | severity by CWV threshold | Metric-Honesty Rule only |

**This is an internal inconsistency, which is one of the reasons CONTRIBUTING lists for filing.** `security-auditor` already encodes exactly the right instinct — rule 1 is "Focus on exploitable vulnerabilities, not theoretical risks" (`:99`) and its severity table bottoms out at `Low = "Theoretical risk"` (`:64`). `security-and-hardening/SKILL.md:295` asks "Is the vulnerability exploitable given your deployment context". `web-performance-auditor` has the Metric-Honesty Rule, which is the same discipline applied to measurement. The security and perf specialists got a likelihood discipline; the general reviewer that runs on *every* change did not.

Two specifics worth calling out:

**`doubt-driven-development` is the most affected**, because its whole design multiplies finding volume: the reviewer prompt at `:90-104` is explicitly "Find what is wrong… Assume the author is overconfident… Do NOT validate", and Step 5 keeps looping while findings are non-trivial. RECONCILE's four classes look like they cover this but don't: `valid trade-off` (`:176`) is about *fix cost* exceeding *acceptance cost*, and `noise` (`:177`) is about the reviewer *lacking context*. A finding that is valid, cheap to fix, and correctly understood by a fully-contexted reviewer — but will never occur — has no class. It lands in `valid + actionable`, triggers a change, and re-loops.

**`code-review-and-quality` already gestures at this and stops short.** "Lead with what matters" (`:191`) orders by leverage, and "Quantify problems when possible" (`:275`) asks for "~50ms per item" instead of "could be slow" — quantified *impact*. Neither asks for quantified *probability*, and neither gates anything.

## Proposed fix

Add an explicit likelihood axis, scored separately from severity and confidence. Naming is the maintainers' call — I use **`materialization`**: *if this ships as written, how likely is it that this actually bites?* Score the probability that the triggering conditions occur, not the blast radius.

Shared bands, so the number means the same thing across surfaces:

| band | meaning |
|---|---|
| `>0.7` | bites on the normal path or the first realistic input |
| `0.3-0.7` | needs a specific but plausible condition — an error path, a concurrent write, a large input |
| `<0.3` | needs an unlikely conjunction, a scenario the design already rules out, or a scale this system will not reach |

Per surface, roughly:

- **`agents/code-reviewer.md`** — every Critical/Important finding carries a score and one sentence naming the triggering condition. Below `0.3` it moves to Suggestion regardless of instinct; Critical requires `≥ 0.5`. `REQUEST CHANGES` then requires at least one finding above the floor, so a review can't block a merge on a list of things that will never happen. *"If you cannot name the condition, you have not established the finding"* is the load-bearing sentence — it is also what stops the score from being reverse-engineered to justify a severity already chosen.
- **`skills/code-review-and-quality/SKILL.md`** — one row in the Step 4 prefix table, and a line under "Honesty in Review": quantifying probability is the same discipline as quantifying impact. This is a small edit; the section already has the right shape.
- **`skills/doubt-driven-development/SKILL.md`** — a fifth RECONCILE class between `valid + actionable` and `valid trade-off`: **valid but immaterial** — real, correctly understood, and its conditions will not occur. Document and do not re-loop. This is the class that lets Step 5 terminate honestly instead of grinding.
- **`agents/security-auditor.md` / `agents/web-performance-auditor.md`** — mostly making the existing instinct explicit and consistent with the shared bands, so `Low = "Theoretical risk"` becomes a scored judgment rather than a category the model can route around by picking `High`.

**Two thresholds, not one, if you want the strong version.** They do different jobs and conflating them is the easy mistake: a *floor* (~0.3) decides what reaches the author at all; a *gate* (~0.5) decides what is allowed to block. A finding between them is worth reporting and fixing, but not worth blocking on.

Whatever the mechanism, one property matters: **nothing filtered may be dropped silently.** A demoted finding keeps its score and its condition sentence so the author can overrule. A filter the author cannot see reads as "the reviewer found nothing".

## Prior art / why I'm confident this works

I shipped this in my own skill pack and it measurably cut waste in a Codex review loop: [pySilver/cc-tools@3ccbedd](https://github.com/pySilver/cc-tools/commit/3ccbedd). Two details from that implementation are worth stealing:

1. **An unknown score must read as maximum, not zero.** Absent or unparseable → treat as `1.0`. Otherwise a reviewer that forgets to score silently suppresses its own findings, and in a loop a broken scorer terminates it early. Fail toward "keep working".
2. **The scorer should not be the party with the inflation bias.** In my loop the reviewer self-scores, and then an independent pass re-scores *blind to the first number* and overrides it. Shown the original score, the second pass anchors to it. If a single agent scores its own findings, expect inflation and say so in the skill.

Happy to open a PR — for all five surfaces or a narrower slice, whichever you prefer. Wanted the shape agreed before writing the diff, per CONTRIBUTING's guidance on justifying the gap first.

pySilver:
Closes #436.

Drafts the likelihood axis you described, scoped to what you asked for: `code-review-and-quality`'s finding triage and the `code-reviewer` persona, echoing into `doubt-driven-development`. I left `security-auditor` and `web-performance-auditor` alone — they already carry the instinct informally, and you didn't ask for them.

Used **likelihood** rather than the `materialization` name from the issue; it's your vocabulary and it reads better in a review context.

## What changed

**`skills/code-review-and-quality/SKILL.md`** — Step 4 gains the axis after the existing prefix table:

- A three-band table (likely / plausible / remote) keyed on the triggering condition, not on a numeric score. Scores invite reverse-engineering to justify a severity already chosen; a named condition is checkable by the author.
- **Name the condition, don't assert the risk.** "Fails if two workers process one shard" is settleable against the single-consumer invariant; "could have a race" is not. Tied explicitly to the reproduce-before-you-raise bar you mentioned.
- **Downgrade, never drop**, with the reasoning stated.
- `Request changes` reserved for findings that are both consequential and plausible.

Echoed in the places the skill already keeps its discipline: one line in Honesty in Review (quantifying likelihood is the same duty as quantifying impact), one Context checklist item, three rationalizations, three red flags, two verification items.

**`agents/code-reviewer.md`** — a `### Likelihood` block under Output Format applying the same rule to Critical/Important/Suggestion, two new Rules, and the output template now asks for the triggering condition on blocking findings.

**`skills/doubt-driven-development/SKILL.md`** — a new RECONCILE class, **valid but immaterial**, inserted at position 2. Precedence matters here: if it sits after `valid + actionable` it can never match, because "real issue requiring a change" matches first. Putting it before forces "can this fire?" ahead of "what do I change?", which is what lets Step 5 terminate honestly instead of grinding.

It is genuinely a distinct class, not a rename of an existing one — `valid trade-off` is about fix *cost*, `noise` is about the reviewer *lacking context*. A fully-informed, correctly-understood finding whose trigger is unreachable had no home before and landed in `actionable`.

## Your two constraints

You flagged the risk that this becomes an excuse to wave real bugs through. Both guards are in the diff:

**Assessed, not guessed.** Every instruction says to check the condition against the project's real inputs, invariants, and configuration and to read them. The failure mode you'd expect — "I can't check, so I'll assume it's fine" — gets a rationalization row and an explicit rule: an unverifiable condition is labelled *unverified*, not downgraded. Guessing high and guessing low are the same error. In `doubt-driven-development` the bar is harder: to classify `immaterial` you must quote the invariant that prevents the trigger, or the class is `actionable`. A feeling isn't an invariant.

**Downgraded and noted, never dropped.** Stated in all three files, with the reason: the author may know the condition is reachable for reasons the diff doesn't show, so downgrading hands them the call while deleting takes it away. A filter the author can't see reads as "the reviewer found nothing". There's a red flag for a finding downgraded with no stated reason, and a verification item for it.

One interaction worth checking: the new class does **not** let an agent evade the doubt-theater signal, since `immaterial` is not `actionable` — a cycle that classifies everything this way still trips the existing red flag. I made that explicit rather than leaving it to be inferred.

## Verification

```
node scripts/validate-skills.js     # 24 skills, 0 errors, 0 warnings
node scripts/validate-commands.js   # 8 commands, 0 errors
node scripts/run-evals-test.js      # 12/12 pass
bash hooks/session-start-test.sh    # payload OK
```

**Separable hunk:** `evals/cases/code-review-and-quality.json` adds two behavioral expectations so the rule is checkable rather than just stated. Drop that hunk if you'd rather land the guidance first and let the eval follow — nothing else depends on it.

No new files, no restructuring; every addition sits in a section that already exists.

### DIFF
--- a/agents/code-reviewer.md
+++ b/agents/code-reviewer.md
@@ -54,6 +54,17 @@ Categorize every finding:
 
 **Suggestion** — Consider for improvement (naming, code style, optional optimization)
 
+### Likelihood
+
+Those three categories are consequence — how bad it is *if* the issue fires. Before assigning one, decide whether it fires at all: name the condition that triggers the finding, and check that condition against the project's real inputs, invariants, and configuration. Read them; don't guess.
+
+- **Critical** and **Important** are for findings that are both consequential and *plausible* under those conditions.
+- A finding whose trigger the code already prevents, or that needs a scale this project won't reach, drops to **Suggestion** with its condition stated — downgraded and noted, never silently dropped. The author may know the condition is reachable for reasons the diff doesn't show; that call is theirs.
+- If you cannot name the condition, you have not established the finding.
+- If you cannot check the condition, say so and label the finding unverified. Guessing high and guessing low are the same error.
+
+`REQUEST CHANGES` requires at least one Critical or Important finding that survives this check.
+
 ## Review Output Template
 
 ```markdown
@@ -64,13 +75,13 @@ Categorize every finding:
 **Overview:** [1-2 sentences summarizing the change and overall assessment]
 
 ### Critical Issues
-- [File:line] [Description and recommended fix]
+- [File:line] [Description, the condition that makes it fire, and recommended fix]
 
 ### Important Issues
-- [File:line] [Description and recommended fix]
+- [File:line] [Description, the condition that makes it fire, and recommended fix]
 
 ### Suggestions
-- [File:line] [Description]
+- [File:line] [Description — for a downgraded finding, the condition and why it can't occur here]
 
 ### What's Done Well
 - [Positive observation — always include at least one]
@@ -89,6 +100,8 @@ Categorize every finding:
 4. Don't approve code with Critical issues
 5. Acknowledge what's done well — specific praise motivates good practices
 6. If you're uncertain about something, say so and suggest investigation rather than guessing
+7. Rate likelihood as well as consequence — a true finding that cannot fire under this project's inputs is a Suggestion, not a blocker
+8. Don't pad the review. Every finding costs the author a read and a dismissal; a list of impossible findings gets skimmed, and the real one gets skimmed with it
 
 ## Composition
 
--- a/evals/cases/code-review-and-quality.json
+++ b/evals/cases/code-review-and-quality.json
@@ -38,7 +38,9 @@
         "Findings cover more than one axis (correctness, readability, architecture, security, performance)",
         "Every finding carries a severity label from the skill's taxonomy",
         "Security of user input is explicitly considered for the new endpoint",
-        "The review leads with high-leverage findings rather than nits"
+        "The review leads with high-leverage findings rather than nits",
+        "Blocking findings name the condition that makes them fire, rather than asserting risk generically",
+        "Any finding judged unlikely under the project's actual inputs is downgraded with its condition stated, not omitted"
       ]
     }
   ]
--- a/skills/code-review-and-quality/SKILL.md
+++ b/skills/code-review-and-quality/SKILL.md
@@ -188,6 +188,22 @@ Label every comment with its severity so the author knows what's required vs opt
 
 This prevents authors from treating all feedback as mandatory and wasting time on optional suggestions.
 
+**Rate likelihood, not just consequence.** Severity is blast radius — how bad it is *if* the issue fires. It says nothing about *whether* it fires. Triage on both: consequence × likelihood, not consequence alone. This is the axis that separates a real bug from the endless supply of technically-correct-but-impossible findings a reviewer can generate ("breaks at 10M rows" on a 400-row internal tool). Neither severity nor your own confidence can filter those — you are *right* about them, and they are still noise.
+
+Assess likelihood **against this project's actual inputs, invariants, and configuration. Read them; do not guess**:
+
+| Likelihood | Triggering condition | Effect |
+|---|---|---|
+| **Likely** | Occurs on the normal path or the first realistic input | Label on consequence |
+| **Plausible** | Needs a specific but real condition — an error path, a concurrent write, a large input | Label on consequence |
+| **Remote** | Needs a conjunction the code already prevents, a scale this system won't reach, or a configuration it doesn't use | Downgrade to **Optional:** / **FYI**, with the condition stated |
+
+**Name the condition, don't assert the risk.** "Fails if two workers process one shard" is checkable — the reader compares it against the single-consumer invariant and settles it. "Could have a race" is not. If you cannot name the condition that makes a finding fire, you have not established the finding — the same bar as reproducing a bug before you raise it.
+
+**Downgrade, never drop.** A remote finding keeps its comment, its condition, and your reasoning; it just stops blocking. The author may know the condition is reachable for reasons the diff doesn't show. Downgrading hands them that call — deleting takes it away, and a filter the author can't see reads as "the reviewer found nothing".
+
+**Reserve Request changes for findings that are both consequential and plausible.** Blocking a merge on things that cannot happen trains authors to discount the whole review, including the finding that was real.
+
 **Lead with what matters.** Order findings by leverage: correctness and security first, then structural regressions and missed simplifications, then everything else. Don't bury a real issue under cosmetic nits — a few high-conviction comments beat a long list. If you have one structural problem and ten nits, the structural problem *is* the review.
 
 ### Step 5: Verify the Verification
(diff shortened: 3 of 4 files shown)


codeCraft-Ritik (commented):
Really nice work on this! I like how you've introduced the likelihood axis in a structured way without overcomplicating the existing review flow. The distinction between likely, plausible, and remote makes the guidance much more actionable, and tying findings to explicit triggering conditions rather than speculation is a thoughtful improvement.

pySilver:
I am testing this addition for a while and I'm happy to see less nonsense issues flagged while doing code or plan reviews.

addyosmani (approved):
This is exactly what I hoped #436 would turn into, and you built the part I most cared about. The three-tier likelihood (Likely / Plausible / Remote) is clean, and "downgrade, never drop" is the line that makes it safe: a remote finding keeps its comment and its reasoning and just stops blocking, so the author still gets to say "actually that condition is reachable, here's why." The reciprocal rule, that overstating likelihood is the same dishonesty as softening a real bug, is the right guard against the axis becoming an excuse to wave things through. Assessing against the project's real inputs rather than guessing, and flagging unverified when you can't check, closes the last gap. Merges clean, validators pass, routing unchanged.

This does add a genuine dimension to the flagship review skill's philosophy, so I want Addy's eye on it before it lands, but from a quality standpoint I'm happy with it. Approving.

addyosmani:
@nucliweb @federicobartoli could one of you take this one before it lands? It adds a likelihood axis to review-finding triage: downgrade the improbable-but-technically-correct findings ("breaks at 10M rows" on a 400-row table) instead of blocking a merge on them, with "downgrade, never drop" as the guard. I've reviewed and approved it, it's clean and green, and frontmatter is untouched so there's no routing impact. But it changes review-blocking behavior across code-review-and-quality and doubt-driven-development, so I want a second set of eyes rather than merging on my approval alone. Background is in #436.

addyosmani:
@nucliweb @federicobartoli re-nudge on this one: it's approved and green, just needs one of your eyes since it changes review-blocking behavior. Whenever you have a moment.

PR #443, opened 2026-08-01
test(ci): add command validator regression coverage

ayobamiseun:
## Summary

`validate-commands.js` currently runs against the repository's eight valid command triples in CI. That proves the catalog passes, but not that the validator rejects the invalid states it exists to catch.

This PR adds a fixture-driven `node:test` suite that runs the real CLI inside temporary repositories and asserts its public exit codes and diagnostics. It then runs that suite in the existing command-validation CI job.

## Coverage added

Six regression cases cover the validator's main contracts:

| Case | Expected behavior |
|---|---|
| Matching command twins | Pass, including the `plan` to `planning` stem mapping |
| Claude command missing a TOML twin | Fail with the missing directory named |
| TOML command missing a Claude twin | Fail and count the canonical command correctly |
| Description drift across tools | Fail and print all three descriptions |
| Present file with no description | Fail with an actionable malformed-description diagnostic |
| Escaped quotes in TOML | Parse and compare the normalized description correctly |

The passing stem-mapping fixture also uses single-quoted TOML, preserving coverage for that supported syntax.

## Implementation

- Copies `validate-commands.js` into a fresh temporary repository for each test scenario.
- Writes only the command files needed for that scenario, then spawns the validator as a CLI.
- Removes every temporary repository after its test.
- Adds no dependencies and makes no production changes to the validator.

## CI

The existing `validate-commands` job now runs:

```text
node --test scripts/validate-commands-test.js
```

This keeps the real-catalog gate and the validator self-tests together.

## Verification

- `node --test scripts/validate-commands-test.js` - 6/6 pass
- `node --test scripts/run-evals-test.js scripts/validate-commands-test.js` - 18/18 pass
- `node scripts/validate-commands.js` - 8 commands, 0 errors
- `node scripts/validate-skills.js` - 24 skills, 0 errors, 0 warnings
- `node scripts/run-evals.js` - 124 checks passed, 0 errors, 0 warnings; rank-1 86%
- `git diff --check upstream/main...HEAD` - clean

### DIFF
--- a/.github/workflows/test-plugin-install.yml
+++ b/.github/workflows/test-plugin-install.yml
@@ -40,6 +40,9 @@ jobs:
       - name: Validate commands across all tool directories
         run: node scripts/validate-commands.js
 
+      - name: Test command validator
+        run: node --test scripts/validate-commands-test.js
+
   validate:
     name: Validate plugin structure
     needs: [validate-skills, validate-commands]
--- a/scripts/validate-commands-test.js
+++ b/scripts/validate-commands-test.js
@@ -1,0 +1,150 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const assert = require('node:assert/strict');
+const fs = require('node:fs');
+const os = require('node:os');
+const path = require('node:path');
+const { spawnSync } = require('node:child_process');
+const { afterEach, test } = require('node:test');
+
+const VALIDATOR = path.join(__dirname, 'validate-commands.js');
+const sandboxes = [];
+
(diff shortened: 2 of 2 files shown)


addyosmani (approved):
Good hygiene and welcome. Regression coverage on validate-commands is exactly the kind of thing that should have tests, and given how much churn skill-lint and the validators have seen lately (#387, #405, #428, #444), locking the command-parity behavior down with tests is worth it. Test-merged, the new suite runs green. Approving.

PR #444, opened 2026-08-01
fix(validator): recognize alternate fenced code blocks

### LINKED ISSUE #437: bug(validator): stripFencedCodeBlocks misses tilde/indented fences and mismatched closer length
Surfaced by @ujjwalks while reviewing #428: https://github.com/addyosmani/agent-skills/pull/428#issuecomment-5132076791

**Where:** `stripFencedCodeBlocks` in `scripts/lib/skill-lint.js`. It matches fences with a single regex over the whole document:

```
```

This only recognizes a backtick fence opening at column zero and closing with a run of backticks of exactly the same length. Three CommonMark-legal forms slip through and their contents stay in the prose both downstream rules read:

- tilde fences (`~~~`)

**Impact on `main` today (not just #428):** the section-heading check (fixed in #405 to strip fences before matching) routes through this helper. A `## Overview` heading inside a `~~~` example block in a skill's `SKILL.md` currently satisfies the Overview requirement for a skill that has no real Overview section — the exact false-negative #405 was written to close, just via a fence form the regex doesn't see. This is live on `main`, independent of #428.

**Impact on #428 (in flight):** #428's new prose cross-reference patterns also route through this helper (matched against fence-stripped content, per its stated design). The same gap means a fenced `` see `made-up-skill` `` inside a tilde or indented fence still produces a dead-reference warning — exactly the case #428's own test `references inside fenced code blocks are not collected` is asserting against, just in a fence form the current helper doesn't recognize.

**Fix, already built and verified by @ujjwalks:** a line-by-line scan tracking the open fence's marker and length (an unterminated fence runs to end of file, which correctly surfaces as missing-section errors rather than prose that silently passes). Rebased on #428's commit `0980192`, at `ujjwalks/agent-skills@d76aa7f` (branch `fix/skill-lint-fence-parsing-on-428`), touching `skill-lint.js` and 13 new cases appended to `scripts/skill-lint-test.js` — 6 of 13 fail against the current helper (both tilde cases, both indented cases, backtick-doesn't-close-tilde, unterminated), including the two cross-reference cases. Full suite green with the fix: 31/31 in `skill-lint-test.js`, 24 skills, 8 commands, 124 eval checks at rank-1 86%, `run-evals-test.js` 12/12. Corpus output unchanged.

**Why a separate issue instead of folding into #428:** the section-check half of this bug predates #428 and affects `main` regardless of #428's outcome. @addyosmani already flagged #428 as the fourth in-flight change touching this file and asked for @nucliweb's eyes before merge — tracking this separately avoids stacking a fifth concern onto that review and lets the `main`-branch fix land on its own schedule. Whoever picks this up should decide whether the standalone fix lands against `main` first (with #428 rebasing after) or the reverse.

Non-goal: this issue is not proposing changes to #428's frontmatter-parsing or cross-reference-pattern scope, only to the shared `stripFencedCodeBlocks` helper both depend on.

ujjwalks:
## Summary

Fixes the fenced-code scanner bug documented in #437.

`stripFencedCodeBlocks` previously used one whole-document regex that recognized only column-zero backtick fences whose closer had exactly the opener's length. Valid CommonMark tilde fences, fences indented by up to three spaces, and longer closing runs were left in the prose inspected by the required-section rule. As a result, an example containing `## Overview` could satisfy the requirement for a skill with no real Overview section.

This change:

- replaces the regex with a line-by-line scanner that tracks the opening marker and length;
- closes only on a same-marker run that is at least as long as the opener;
- supports up to three spaces of fence indentation;
- treats an unterminated fence as continuing to the end of the file;
- adds focused `node:test` coverage and runs it in CI.

The regression suite was verified red against the old helper (5 failures) and green with the scanner (12/12 passing), including a direct case where a heading inside a fence with a longer closer must not satisfy the section check.

## Verification

- `node --test scripts/lib/skill-lint-test.js` — 12/12 passed
- `node scripts/validate-skills.js` — 24 skills, 0 errors, 0 warnings
- `node --test scripts/run-evals-test.js` — 12/12 passed
- `node scripts/run-evals.js --min-rank1 80` — 124 checks passed; rank-1 86%
- `git diff --check origin/main...HEAD` — clean

Fixes #437.


addyosmani (approved):
This is the fix for #437 and it's the right one. I test-merged and ran it: the skill-lint suite is green, validate-skills passes, and the new stripFencedCodeBlocks tracks the opener's marker and length, allows up to three spaces of indent, and closes only on a same-or-longer run of the same marker, so tilde fences, indented fences, and mismatched-length closers all get stripped now. That closes the live false-negative where a `## Overview` inside a `~~~` block satisfied the section check for a skill with no real Overview. Focused, tested, and extracted cleanly rather than bundled into other validator work. CI hasn't run on the branch yet, but I reproduced it locally green. Approving.

addyosmani:
Approved on the merits: I verified this fixes the #437 fence false-negative end to end (the new stripFencedCodeBlocks tracks the opener marker and length, allows up to three spaces of indent, and closes only on a same-or-longer run, so tilde, indented, and mismatched-length fences all strip now), and the skill-lint suite is green locally. The one blocker is mechanical: this is still marked as a draft, so GitHub won't let it merge. @ujjwalks could you flip it to ready for review? Then it can land. @nucliweb @federicobartoli flagging for a quick look too, since it closes a live validator false-negative on main.

ujjwalks:
Done — #444 is now marked ready for review. Thanks for approving.

addyosmani:
@ujjwalks this is approved and I verified it fixes the live #437 fence bug, but it's still marked as a draft so I can't merge it. Could you flip it to "ready for review"? Then it goes in.

ujjwalks:
@addyosmani I've marked this PR as ready for review. Please let me know if anything else is needed from my side.

abhisheksharma2411 (commented):
The scanner is correct, and I checked it against CommonMark rather than just reading it. Three things it gets right that are easy to get wrong:

- the closing regex allows `` `+ `` while the opening requires `` `{3,} `` — so a stray single backtick line inside a block can't close it, since `1 >= 3` fails
- `^ {0,3}` on both ends is exactly the CommonMark bound, so a 4-space-indented line stays an indented code block rather than being read as a fence
- closing on a same-marker run *at least* as long as the opener handles the mismatched-closer case from #437 without letting a tilde run close a backtick fence

Verifying red-against-old before trusting the suite is the right instinct, and it's why the "heading inside a fence with a longer closer" case is worth more than the other eleven.

**Two coordination things, one of which is my fault.**

**1. This PR and my #505 both create `scripts/lib/skill-lint-test.js`.** Yours is approved and also wires it into `test-plugin-install.yml`, so yours is the canonical one — mine is the duplicate. I'd rather not drop a conflict on an approved PR, so: if #444 merges first I'll rebase #505 and move my single test case into your file, and #505's diff becomes two lines in `skill-lint.js`. Nothing needed from you.

Worth flagging for a maintainer that #428 adds a *third* file at `scripts/skill-lint-test.js` — a different path for the same module. Whichever order those two land, one of them should probably move, or the repo ends up with two test files for `skill-lint.js` and only one of them in CI.

**2. This PR is approved but has no CI at all, which is probably why it can't merge.**

```
head 756e76b29    mergeable_state: unstable
combined status:  pending   (0 statuses)
check runs:       0
```

A recently merged PR on the same repo shows 4 check runs, so 0 isn't normal — the workflow has never been released for this head commit. That's the usual first-time-contributor gate, and it needs a maintainer to click through; `unstable` here means "checks pending", not "checks failing". Given @addyosmani has already approved the content, this looks like it's waiting on nothing but that.

The head is from 12 Aug, so a rebase would also re-trigger it if that's easier than releasing the run.

**One thing outside this PR's scope, noted so it isn't assumed fixed:** the required-section check is now honest about fences, but `exempt = dirName in SECTION_EXEMPT_SKILLS` still lets a skill directory named `constructor` skip section checks entirely via the prototype chain — `exempt: true`, zero errors, no sections. Different failure, same rule. That's all #505 is now.

abhisheksharma2411:
Heads-up, and an apology — this is now `mergeable: false / dirty`, and that's my doing. #505 merged a few hours ago and it added `scripts/lib/skill-lint-test.js`, the same path this PR adds, so you've got an add/add conflict on a PR that's been approved since 3 August.

I said on this PR that I'd fold my case into your file once yours landed. Yours was approved first and mine merged first, so let me just do the work rather than hand you a chore.

**Resolved branch:** [`abhisheksharma2411:merge/444-fence-parsing-plus-main`](https://github.com/abhisheksharma2411/agent-skills/tree/merge/444-fence-parsing-plus-main)

```
git remote add abhishek https://github.com/abhisheksharma2411/agent-skills.git
git fetch abhishek merge/444-fence-parsing-plus-main
git merge abhishek/merge/444-fence-parsing-plus-main
```

Or resolve it yourself in about a minute — the conflict is one file and the two test sets are disjoint:

- `scripts/lib/skill-lint.js` **auto-merged with no conflict**, so there's no logic to reconcile.
- The test file is a straight union: my exemption/guardrail cases, then your fence cases, unchanged.
- `KNOWN` becomes `new Set(['alpha', 'beta', 'portable-skill'])` so your fixtures still resolve.

I checked that the union isn't vacuous, since a test that passes either way is worth nothing. Swapping main's pre-#444 lint script under the merged test file:

```
ℹ tests 19
ℹ pass 14
ℹ fail 5

✖ a heading inside a tilde fence does not satisfy the check
✖ a heading inside an indented fence does not satisfy the check
✖ a heading inside a fence with a longer closer does not satisfy the check
✖ a backtick fence does not close a tilde fence
✖ an unterminated fence fails loud rather than passing silently
```

Put your fence fix back and it's 19/19. Your change is carrying all five of those, which is a nice demonstration that the line-by-line scanner is doing real work over the old single-regex approach.

Separately, the reason this has sat since 3 August with an approval on it looks unrelated to the conflict: the head commit has **0 statuses and 0 check runs**, where recently-merged PRs in this repo show 4. It seems to be waiting on a workflow run that never got dispatched rather than on anything you did — might be worth a nudge to @addyosmani, since a re-push for the conflict resolution should also kick CI into life.

ujjwalks:
Conflict resolved and pushed in aba4211 against the current main branch. GitHub now reports this PR as mergeable, and the existing approval remains valid. I reran the full local verification: all 13 checks passed, including 19/19 skill-lint tests, validation of all 24 skills, and 127/127 eval checks. The remaining blocker is the newly dispatched Test Plugin Installation workflow, which is awaiting maintainer approval: https://github.com/addyosmani/agent-skills/actions/runs/33049560912

abhisheksharma2411:
Closing the loop on the file collision I flagged, since it resolved in the other order to the one I guessed.

#505 merged first, so `scripts/lib/skill-lint-test.js` was already on `main` when you merged it in — nothing left for you to do, and nothing lost. I checked the merge rather than assuming:

```
$ git merge pr-444 into main
Auto-merging scripts/lib/skill-lint-test.js
Auto-merging scripts/lib/skill-lint.js
merge exit=0

$ node --test scripts/lib/skill-lint-test.js
ℹ tests 20 | pass 20 | fail 0
```

Both sides survive — your fence tests and the `Object.prototype` one from #505, plus `reports a workflow step declared without a matching process section`, which landed on `main` after your merge base. I briefly thought your branch dropped that one (it isn't in your head commit) before checking the merge base and finding it simply postdates it. Clean auto-merge, no conflict.

On your branch as it stands: **19/19 pass locally**. So the code is fine and this is still purely the CI gate.

One thing worth knowing so you don't keep spending merges on it: the two `main` merges won't have moved the needle, because the gate is per-**contributor**, not per-commit. It's still sitting at:

```
mergeable:   MERGEABLE
review:      APPROVED
checks:      0
state:       UNSTABLE
```

`UNSTABLE` with zero checks means "no workflow has ever been released for this PR", not "checks failed" — first-time-contributor approval, which only a maintainer can clear with **Approve and run workflows**. Pushing more commits will keep producing zero checks until someone does that.

@addyosmani — this one's approved by you, merges cleanly, and passes 20/20 with `main` merged in. It just needs the workflow run released so the branch protection has something to look at.

PR #446, opened 2026-08-02
docs(ci-cd): add eval gating step for LLM features

HMAKT99:
## What this adds

A `## Eval Gating for LLM Features` section to `skills/ci-cd-and-automation/SKILL.md` — the natural home, since this skill already owns the quality gate pipeline and the CI/agent feedback loop.

The section covers the one piece the existing skill is missing: **what to do when your project has LLM features** (prompts, model config, retrieval logic, tool definitions) that can regress in ways unit tests won't catch.

## Why this fits ci-cd-and-automation

The skill already has "Feeding CI Failures Back to Agents" — establishing it as the home for CI/agent interaction. Eval gating is the companion: before you can feed a failure back to an agent, you need a gate that detects the failure. The section slots between those two concerns naturally.

## Content summary (+48 lines, one file)

- **Path-filtered GitHub Actions workflow** — eval job triggers only when LLM-affecting files change, not on every commit (contains model API cost)
- **Four categories of eval cases** — golden pairs, regression cases from incidents, adversarial inputs, edge cases
- **Pass-rate gating vs. exact-match** — LLM output is non-deterministic; exact-match assertions produce flaky evals; gate on a threshold calibrated from baseline instead
- **The merge rule** — a PR that drops eval pass rate is treated identically to a PR that breaks a unit test: do not merge
- **Cross-references** `test-driven-development` and `observability-and-instrumentation` for building and evolving the eval suite (no duplication)
- Two rationalization rows, two red flags, two verification checkboxes in existing sections

## What this is NOT

This section covers the CI integration layer only — when to trigger, how to configure the gate, and what counts as passing. Building the eval suite itself (writing cases, choosing judges, closing the feedback loop) is covered by the cross-referenced skills.

Validator: 0 errors, 0 warnings.

PR #447, opened 2026-08-02
docs(context): add Context Budget Management section

HMAKT99:
## What this adds

A `## Context Budget Management` section to `skills/context-engineering/SKILL.md` — filling the one gap the existing skill has: it covers how to load and structure context at session start, but has nothing on managing context as the window fills during a long task.

## The gap this fills

Level 5 of the Context Hierarchy ("Conversation Management") currently says: *"Start fresh sessions when context is getting long."* That's correct but incomplete — it describes the last resort, not the discipline that avoids needing it. Agents running long coding sessions (debugging, multi-file refactors, incremental builds) accumulate conversation history, tool output, and replaced drafts until quality degrades. The missing piece is a framework for trimming proactively.

## Content summary (+47 lines, one file)

- **75% threshold** — start managing at three-quarters capacity, not when the window is full; by the time it's full, attention is already fragmented
- **Cut-first priority table** — four categories ordered by disposability: past failed attempts, verbose tool output, conversational filler, replaced code drafts
- **Protect-until-the-end list** — original task definition, active error message, current file, hard constraints; these survive every trim
- **Compress before dropping** — reduce a long block of exploration to one sentence capturing the conclusion before removing it; preserves the decision, loses the journey
- **Order for recency** — task-critical content (current error, active constraint) belongs last in context, closest to the generation point, not buried under background loaded at session start; transformers weight recent tokens more heavily

Also adds one Anti-Pattern row ("context cliff"), one Red Flag, and two Verification checkboxes to the existing sections.

## Why this fits here

`context-engineering` already owns "what to load" and "how to structure it." Budget management is the natural next concern: once context is loaded and the session is running, how does an agent stay coherent as it fills? This section answers that without duplicating anything in the existing skill.

No open PRs touch `context-engineering`. Cross-references point only to existing skills. Validator: 0 errors, 0 warnings.

### DIFF
--- a/skills/context-engineering/SKILL.md
+++ b/skills/context-engineering/SKILL.md
@@ -118,6 +118,8 @@ Long conversations accumulate stale context. Manage this:
 - **Summarize progress** when context is getting long: "So far we've completed X, Y, Z. Now working on W."
 - **Compact deliberately** — if the tool supports it, compact/summarize before critical work
 
+For the proactive discipline that makes these last resorts unnecessary — what to cut first, what to protect, and when to start — see **Context Budget Management** below.
+
 ## Context Packing Strategies
 
 ### The Brain Dump
@@ -319,10 +321,9 @@ This catches wrong directions before you've built on them. It's a 30-second inve
 - Agent output doesn't match project conventions
 - Agent invents APIs or imports that don't exist
 - Agent re-implements utilities that already exist in the codebase
-- Agent quality degrades as the conversation gets longer
+- Agent quality degrades mid-task as the conversation grows — failed attempts, replaced drafts, and verbose tool output are not being trimmed
 - No rules file exists in the project
 - External data files or config treated as trusted instructions without verification
-- Agent quality degrades mid-task as the conversation grows and context is not actively managed
 
 ## Verification
 
--- a/skills/context-engineering/SKILL.md
+++ b/skills/context-engineering/SKILL.md
@@ -215,7 +215,7 @@ The detail is gone; the decision is preserved. If the detail turns out to matter
 
 ### Order for recency
 
-Put the most task-critical content **last** in context. Transformer attention weights recent tokens more heavily than earlier ones. The active error message or current constraint belongs closer to the generation point than background specs loaded at session start:
+Put the most task-critical content **last** in context. Models recall content at the start and end of the window more reliably than the middle (the lost-in-the-middle effect — Liu et al., 2023). Keep stable rules and specs at the start; put the active task material last, closest to the generation point:
 
 ```
 ← session start                              generation point →
--- a/skills/context-engineering/SKILL.md
+++ b/skills/context-engineering/SKILL.md
@@ -177,6 +177,49 @@ Key files: validation.ts, errors.ts, db.ts
 
 Load only the relevant section when working on a specific area.
 
+## Context Budget Management
+
+The context window is not a filing cabinet — it's a working desk. As a session runs, conversation history, tool output, and exploration accumulate. Most of it becomes deadweight. Budget proactively: waiting until the window is full causes abrupt quality drops; managing regularly keeps the agent coherent through long tasks.
+
+**Start trimming at 75% capacity, not 100%.** By the time the window is genuinely full, the model's attention is already fragmented across too many signals. The 75% threshold gives room to compress gracefully rather than cut desperately mid-task.
+
+### What to cut first
+
+| Content | When to cut |
+|---|---|
+| Past failed attempts and their error output | Once you've moved past them — keep the conclusion, not the journey |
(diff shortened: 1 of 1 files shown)


addyosmani (approved):
This one's on-topic and useful. Context budget management is squarely what context-engineering is for, and the specifics are sound: trim at 75% rather than waiting for a full window, summarize rather than delete, keep the conclusion and drop the exploration. It reads like the rest of the skill. Approving. (Flagging that I'm treating this differently from your #446 and #448)

addyosmani:
@nucliweb @federicobartoli mind a quick look before this merges? It adds a Context Budget Management section to context-engineering (trim at 75% rather than waiting for a full window, summarize over delete). I approved it as on-topic and it's clean and green, body-only with no description change, so no routing surface. Since it adds agent-facing guidance rather than being pure tooling, I'd rather it have a collaborator glance first. If either of you is good with it, it's ready to go.

nucliweb:
Took the requested look. Content is strong and lands in the right skill: this is context engineering for working with agents (proactive budget management as the window fills), which reads as the missing discipline behind Level 5's "start fresh / summarize / compact" last resort. Fits the voice, additive, green. Two small non-blocking notes.

**Duplicate red flag.** The new Red Flag "Agent quality degrades mid-task as the conversation grows and context is not actively managed" overlaps the existing one three lines up, "Agent quality degrades as the conversation gets longer." Worth folding into a single sharpened line rather than listing two near-identical flags.

**Wire it to Level 5.** The section is framed as completing "Level 5: Conversation Management" (which already seeds it with "Summarize progress" and "Compact deliberately"), but neither points at the other and they sit ~60 lines apart. A one-line pointer from Level 5 to this section (and optionally back) would make the brief-intro / full-treatment relationship visible instead of leaving them to read as unrelated.

Neither blocks; both are quick. Nice addition.

HMAKT99:
Thanks @nucliweb — both fixed in the latest push:

- **Duplicate red flag:** Merged the two near-identical lines into one sharper flag that names what to actually trim: *"Agent quality degrades mid-task as the conversation grows — failed attempts, replaced drafts, and verbose tool output are not being trimmed."*
- **Wire to Level 5:** Added a one-line forward pointer at the end of Level 5: Conversation Management pointing to the Context Budget Management section below — makes the brief-intro / full-treatment relationship explicit rather than leaving them 60 lines apart with no connection.

Validator 0/0.

federicobartoli (changes requested):
Strong section and it lands in the right skill , I verified there's no context-budget content anywhere else in the catalog, so this is genuinely additive, and both of nucliweb's notes are reflected in the current diff.

One technical fix before I approve, since this file is canonical context guidance for agents: the sentence "Transformer attention weights recent tokens more heavily than earlier ones" isn't the real mechanism. The documented effect is U-shaped, not monotonic: retrieval is strongest at the start and end of the window and weakest in the middle ("Lost in the Middle", Liu et al., https://arxiv.org/abs/2307.03172). Anthropic's long-context guidance reflects the same U: longform data at the top, query at the end. Your own diagram is already consistent with this (stable rules at the start, active material at the end) , it's just this sentence that, read literally by an agent, implies the start of context is the weak slot. Could you replace it with something like:

"Models recall content at the start and end of the window more reliably than the middle (the lost-in-the-middle effect). Keep stable rules and specs at the start; put the active task material last, closest to the generation point."

With that fixed I'll approve, and it's ready to go per Addy's note.

One open question this section surfaced, aimed at the catalog rather than this PR, so genuinely non-blocking: how much of what we teach here will the harnesses absorb? Auto-compaction and context editing are already eating the mechanics side of this section (a model rarely knows it's at 75% , the tool does). What seems to survive automation is the judgment: what to cut first, what to protect until the end, when to re-anchor. Is that a useful lens for future additions — teach the judgment, let the tooling own the mechanics? Curious what others think, @addyosmani in particular since it touches how the catalog ages.

HMAKT99:
Fixed in the latest push, @federicobartoli — the sentence now reads:

> *"Models recall content at the start and end of the window more reliably than the middle (the lost-in-the-middle effect — Liu et al., 2023). Keep stable rules and specs at the start; put the active task material last, closest to the generation point."*

The diagram and the rule were already consistent with the U-shape; it was only the explanatory sentence that was wrong. Validator 0/0.

On the broader question you raised — teaching judgment vs. mechanics — I think that's the right lens. The four things in this section that survive automation (cut priority order, protect list, compress-before-drop, start/end placement) are all judgment calls. The 75% threshold is the one that blurs the line, but even there the agent's role is deciding *what* to compress, not triggering the compaction.

federicobartoli (approved):
Re-checked the latest push: the lost-in-the-middle sentence now reads exactly as suggested, and the diagram, the Level 5 pointer and the merged red flag are all consistent with it. Nothing else changed since my last pass and it still merges clean. Approving.

PR #448, opened 2026-08-02
docs(simplification): add Right-sizing LLM Calls section

HMAKT99:
## What this adds

A `## Right-sizing LLM Calls` section to `skills/code-simplification/SKILL.md`, applying the skill's existing ethos — *prefer the simplest solution that works* — to LLM usage patterns in code.

## Why code-simplification is the right home

The skill already covers avoiding premature abstraction, unnecessary indirection, and over-engineered solutions. Using a flagship model for every LLM call (summarization, classification, templated generation) is the same anti-pattern applied to AI: more power than the task needs, added cost and latency for no quality gain. The framing fits naturally without duplicating anything in the existing skill.

This is also distinct from the open `llm-cost-optimization` PR (#287), which covers runtime API cost engineering (caching, history windowing, retrieval narrowing, output capping). This section is about **code-level design decisions** — choosing the right model tier, structuring prompts to avoid chains, and only including context that changes the output.

## Content summary (+44 lines, one file)

- **Model tier selection table** — maps task complexity to the right model tier: fast/cheap for templated tasks with fixed output formats; mid-tier for code generation and structured extraction; powerful only for architecture decisions and adversarial review. Rule: start cheap, move up only when outputs fail eval.
- **Prefer one prompt over a chain** — multi-step chains multiply cost and failure surface; includes a before/after code example. A chain earns its complexity only when intermediate outputs need human review or routing logic.
- **Include only context that changes the output** — every token is cost and noise; cross-references `context-engineering` for techniques without duplicating its content.
- **Memoize deterministic calls** — same prompt + input → equivalent output means cache, not re-call. File summarization, code docs, classification against fixed schemas are all candidates.

Also adds two rationalization rows, two red flags, and two verification checkboxes to the existing sections.

Validator: 0 errors, 0 warnings. No open PRs conflict with `code-simplification`.

### DIFF
--- a/skills/code-simplification/SKILL.md
+++ b/skills/code-simplification/SKILL.md
@@ -294,6 +294,44 @@ function UserBadge({ user }: Props) {
 // This is a judgment call — flag it, don't auto-refactor.
 ```
 
+## Right-sizing LLM Calls
+
+The same simplification principle that removes unnecessary abstraction applies to LLM usage: don't use more model than the task requires. Using a flagship model for every call — summarization, classification, templated generation — is the LLM equivalent of importing a framework to center a div. It adds cost and latency without improving output quality.
+
+### Match model tier to task complexity
+
+| Task type | Complexity signal | Right tier |
+|---|---|---|
+| Summarization, classification, templated generation | Fixed output format, well-defined input | Fast/cheap (Haiku, Flash, GPT-3.5) |
+| Code generation, structured extraction, multi-step reasoning | Requires judgment, ambiguous input | Mid-tier (Sonnet, GPT-4o-mini) |
+| Architecture decisions, adversarial review, novel problem solving | High stakes, no clear right answer | Powerful (Opus, GPT-4o) |
+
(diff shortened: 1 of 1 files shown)


addyosmani (commented):
Holding this one for a scope call from Addy rather than reviewing it on the merits, and I want to be transparent about why. The content itself is fine, but it's LLM-application-engineering guidance (model-tier selection, prompt-chain cost, eval-gating model behavior) added into a general software-engineering skill. The pack's remit so far has been engineering practices for working with AI coding agents, not building LLM features, and there's a cluster of pending proposals sitting on exactly this line (#285, #286, #287). Whether the pack expands into LLM-app engineering is a direction decision, not something I want to settle by merging it in through the side door of an existing skill. So this waits on that call. If Addy decides that's in scope, this is a reasonable start; if not, this content would want its own home rather than living here. Not a knock on the writing.

addyosmani:
Thanks for this. Right-sizing LLM calls is LLM-application guidance rather than code simplification, so it's out of scope for this skill. Closing.

PR #452, opened 2026-08-05
fix(codex): complete plugin install flow and manifest metadata

dev-minggyu:
## Summary

- Add the missing Codex plugin installation command after marketplace registration
- Complete the interface metadata expected by Codex's bundled plugin validator
- Remove the empty `hooks` entry rejected by that validator while leaving the Claude Code lifecycle hook unchanged

## Context

`codex plugin marketplace add addyosmani/agent-skills` registers the marketplace but does not install its plugin. The plugin remains `not installed` until users run:

```bash
codex plugin add agent-skills@agent-skills
```

The Codex CLI accepts the previous manifest, while Codex's bundled plugin validator applies a stricter manifest contract: it rejects the `hooks` field and expects additional interface metadata.

This change aligns the Codex manifest with that validator without modifying the Claude Code lifecycle hook under `hooks/`.

## Validation

- Codex bundled plugin validator passes
- Verified that marketplace registration alone reports the plugin as `not installed`
- Verified that `codex plugin add` installs and enables the plugin
- Confirmed all 24 skills are present in the installed cache
- Validated all 24 skills and 8 command mappings
- Passed 18 unit tests and 124 evaluation checks

## Follow-up

It may also be useful to add a Codex installation job to `test-plugin-install.yml` in a follow-up change so both Claude Code and Codex installation paths are covered in CI.

### DIFF
--- a/.codex-plugin/plugin.json
+++ b/.codex-plugin/plugin.json
@@ -10,12 +10,20 @@
   "repository": "https://github.com/addyosmani/agent-skills",
   "license": "MIT",
   "skills": "./skills/",
-  "hooks": {
-    "hooks": {}
-  },
   "interface": {
     "displayName": "Agent Skills",
     "shortDescription": "Senior-engineer workflows for AI coding agents: spec, plan, build, test, review, ship.",
-    "category": "Productivity"
+    "longDescription": "Agent Skills bundles 24 production engineering workflows from Addy Osmani, covering requirements, planning, implementation, testing, review, and shipping.",
+    "developerName": "Addy Osmani",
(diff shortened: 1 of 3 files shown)


addyosmani (approved):
The install-flow fix is real and worth taking: `marketplace add` alone doesn't install, so documenting the `codex plugin add` step is correct, and dropping the empty `hooks` entry that Codex's validator rejects is a genuine fix. One coordination note: #451 and my #463 also touch the plugin manifests, so we need to sequence these so they don't stomp each other. Happy with the direction; let's land it alongside the manifest cleanup rather than racing it.

PR #454, opened 2026-08-05
security: fix 4 critical vulnerabilities in eval pipeline (symlink escape, path traversal, logic bug, temp leak)

Sunil56224972:
## Security Audit: 4 Critical/High Vulnerabilities in scripts/run-evals.js

Deep manual code review of the eval pipeline found **4 real, exploitable vulnerabilities** — each with proof-of-concept testing. All fixes are in a single file (scripts/run-evals.js) with minimal, surgical changes.

---

### Bug 1: Symlink Sandbox Escape via 
esolveFixturePath + s.cpSync (Critical)

**Location:** 
esolveFixturePath() (line 183) + materializeWorkspace() (line 400)

**The Bug:** 
esolveFixturePath() validates the *logical* path but never resolves symlinks. A malicious fixture directory containing a symlink (e.g., evals/fixtures/evil/escape -> /etc/passwd) passes validation because the relative path evil/escape stays within the fixtures root. Then s.cpSync(src, dest, { recursive: true }) copies the symlink *as a live symlink* into the throwaway workspace (default dereference: false). The agent then executes with Bash tool access in that workspace and can follow the symlink to read/write arbitrary files on the host.

**PoC:**
`javascript
// resolveFixturePath allows 'evil-skill' — it's within fixtures/
resolveFixturePath(fixturesDir, 'evil-skill');  // ✓ PASSES

// But evil-skill/ contains: escape-link -> /etc/passwd  
// fs.cpSync preserves the symlink into the workspace
// Agent gets Bash access → can read /etc/passwd through the link
`

**Fix:**
- Use s.realpathSync() to resolve symlinks before the boundary check
- Pass dereference: true to s.cpSync() so symlinks become regular files

---

### Bug 2: Path Traversal via --behavioral CLI Argument (Critical)

**Location:** 
unBehavioral() (line 446) + main() (line 553)

**The Bug:** The skillName argument from --behavioral has **zero validation**. It's used directly in:
- path.join(CASES_DIR, skillName + '.json') → reads arbitrary .json files
- path.join(SKILLS_DIR, skillName, 'SKILL.md') → reads arbitrary files  
- path.join(RESULTS_DIR, skillName + '.eval-N') → **writes** .grading.json to arbitrary paths

A value like ../../etc/passwd traverses outside the project directory for both reads and writes.

**PoC:**
`javascript
// With skillName = '../../..'
path.join('/project/evals/results', '../../..'.eval-1') 
// Resolves to: /project/../../../.eval-1.grading.json
// = /.eval-1.grading.json  ← ARBITRARY FILE WRITE
`

**Fix:** Validate skillName against /^[a-z0-9]+(-[a-z0-9]+)*$/ (kebab-case only) before any filesystem operation.

---

### Bug 3: 
esolveFixturePath Rejects Valid Path . (High — Logic Bug)

**Location:** 
esolveFixturePath() line 190

**The Bug:** The check ack === '' rejects paths where path.relative(root, resolved) returns ''. This happens when 
el = '.' — a valid fixture path meaning 'the fixture root itself'. path.relative('/a', '/a') returns '', but the code treats this as an escape when it's actually the root. This breaks any fixture that needs to reference the root directory.

**PoC:**
`javascript
resolveFixturePath(fixturesDir, '.');
// throws: 'fixture path escapes workspace: .'
// But '.' IS a valid path — it resolves to the root itself, not an escape
`

**Fix:** Remove the ack === '' check. A path resolving to the root is valid, not an escape.

---

### Bug 4: Temp Workspace Directories Never Cleaned Up (High — Data Leak)

**Location:** 
unBehavioral() (lines 482-531)

**The Bug:** materializeWorkspace() creates directories in os.tmpdir() but 
unBehavioral() **never cleans them up** — not on success, not on failure. If execFileSync throws (timeout, signal, OOM), the workspace persists in /tmp indefinitely with fixture data that may be sensitive. The test file 
un-evals-test.js properly uses 	ry/finally cleanup (line 234), proving the authors know the pattern but omitted it in production code.

**Fix:** Wrap the eval execution loop in 	ry/finally with s.rmSync(workspace, { recursive: true, force: true }).

---

### Testing

All fixes were verified with PoC scripts that confirmed:
- 
esolveFixturePath('.') no longer throws (Bug 3 fix)
- Symlink resolution catches escape attempts (Bug 1 fix) 
- Kebab-case validation rejects path traversal payloads (Bug 2 fix)
- Workspace cleanup happens in finally block (Bug 4 fix)

### Changes

Single file modified: scripts/run-evals.js (+36 lines, -5 lines)

### DIFF
--- a/scripts/run-evals.js
+++ b/scripts/run-evals.js
@@ -184,13 +184,25 @@ function resolveFixturePath(root, rel) {
   if (path.isAbsolute(rel)) {
     throw new Error(`fixture path must be relative: ${rel}`);
   }
-  const resolvedRoot = path.resolve(root);
+  const resolvedRoot = fs.realpathSync(path.resolve(root));
   const resolvedPath = path.resolve(resolvedRoot, rel);
-  const back = path.relative(resolvedRoot, resolvedPath);
-  if (back === '' || back === '..' || back.startsWith(`..${path.sep}`) || path.isAbsolute(back)) {
+  // Resolve symlinks when the target exists so a symlink inside a fixture
+  // directory cannot escape the sandbox. For new files (workspace dest),
+  // realpath the parent directory instead.
+  let realPath;
+  try {
+    realPath = fs.realpathSync(resolvedPath);
+  } catch {
+    // Target doesn't exist yet (workspace destination). Resolve the parent
+    // to catch symlinked parent directories, then re-append the basename.
+    const parentReal = fs.realpathSync(path.dirname(resolvedPath));
+    realPath = path.join(parentReal, path.basename(resolvedPath));
+  }
+  const back = path.relative(resolvedRoot, realPath);
+  if (back === '..' || back.startsWith(`..${path.sep}`) || back.startsWith(`../`) || path.isAbsolute(back)) {
     throw new Error(`fixture path escapes workspace: ${rel}`);
   }
-  return resolvedPath;
+  return realPath;
 }
 
 // ---------- tier 2 ----------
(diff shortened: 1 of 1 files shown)


Sunil56224972:
Superseded by individual PRs for easier review: #455 (symlink escape), #456 (path traversal), #457 (logic bug), #458 (temp cleanup)

PR #457, opened 2026-08-05
fix: resolveFixturePath incorrectly rejects valid path '.'

Sunil56224972:
## Logic Bug in resolveFixturePath (High)

**File:** `scripts/run-evals.js` — `resolveFixturePath()` (L190)

### The Bug

The boundary check at line 190 includes `back === ''` as a rejection condition:

\\\javascript
if (back === '' || back === '..' || back.startsWith('..'+path.sep) || path.isAbsolute(back)) {
  throw new Error('fixture path escapes workspace: ' + rel);
}
\\\

When `rel = '.'`, the resolved path equals the root itself, and `path.relative(root, root)` returns `''`. The code rejects this as an 'escape' — but **a path that resolves to the root IS valid**, not an escape. This breaks any fixture path that needs to reference the root fixture directory.

### PoC

\\\javascript
resolveFixturePath(fixturesDir, '.');
// throws: 'fixture path escapes workspace: .'
// But '.' resolves to the root itself — that's a valid path, not an escape!
\\\

**Tested and confirmed** — `'.'` is rejected when it shouldn't be.

### Fix

Remove the `back === ''` check. The remaining checks (`back === '..'`, `back.startsWith('..' + path.sep)`, `path.isAbsolute(back)`) still correctly catch actual traversals.

### Changes

Single file: `scripts/run-evals.js` (+1 line, -1 line)

### DIFF
--- a/scripts/run-evals.js
+++ b/scripts/run-evals.js
@@ -187,7 +187,7 @@ function resolveFixturePath(root, rel) {
   const resolvedRoot = path.resolve(root);
   const resolvedPath = path.resolve(resolvedRoot, rel);
   const back = path.relative(resolvedRoot, resolvedPath);
-  if (back === '' || back === '..' || back.startsWith(`..${path.sep}`) || path.isAbsolute(back)) {
+  if (back === '..' || back.startsWith(`..${path.sep}`) || path.isAbsolute(back)) {
     throw new Error(`fixture path escapes workspace: ${rel}`);
   }
   return resolvedPath;


addyosmani (changes requested):
What's the actual case where a fixture path is `.`? Dropping the empty-relative guard means resolveFixturePath returns the root itself for `.`, and I can't see when a fixture legitimately wants to reference the whole fixtures root. Loosening a boundary check needs a concrete failing case behind it. If you have one, show me and I'll take it; otherwise I'd rather keep the guard. Holding for that.

Sunil56224972:
You're right — I don't have a concrete failing case. Every existing fixture uses a skill-name directory (e.g. `api-and-interface-design`), and I can't construct a legitimate scenario where `.` as a fixture path would be needed. The guard is intentional and correct. Closing this one.

PR #464, opened 2026-08-06
ci: guard spec/plan/todo artifact paths against silent drift

nucliweb:
## Why

While reviewing #93 (which repoints `/spec` and `/plan` at `docs/features/[name]/`), I found it would break the `/spec → /plan → /build` pipeline: the producers moved the artifacts, but `/build` and the spec/plan skills still require `SPEC.md` and `tasks/plan.md`. Crucially, **CI stayed green**; the command-parity check only compares command descriptions, not artifact paths, so nothing guards the paths the pipeline agrees on.

This PR closes that gap so a #93-style drift fails CI instead of merging silently.

## What

A standalone validator, `scripts/validate-artifact-paths.js`, that enforces one canonical set of spec/plan/todo artifact paths across every file in the pipeline:

- producers: `.claude/commands/spec.md`, `.claude/commands/plan.md`
- consumer: `.claude/commands/build.md`
- skills: `spec-driven-development`, `planning-and-task-breakdown`
- docs: `getting-started.md`, `adoption-guide.md`

Any spec/plan/todo artifact path outside the allowlist (`SPEC.md`, `docs/SPEC.md`, `tasks/plan.md`, `tasks/todo.md`) fails the check, so changing the convention has to update the allowlist and every guarded file in the same change.

Scope is deliberately narrow: only spec/plan/todo artifacts, only the pipeline files. It is not a general markdown path linter.

## Verification

- `scripts/validate-artifact-paths-test.js`: 6/6 passing, including a test that reproduces the #93 drift.
- Against current `main`: 7 files checked, 0 errors, PASSED.
- End-to-end: applying #93's actual producer changes to the tree turns the validator red (7 errors, FAILED), pointing at the exact drifted lines in `spec.md`, `plan.md`, and `getting-started.md`, while the consumers that #93 left on the old paths still pass, surfacing the inconsistency.
- Wired into the `validate-commands` CI job alongside its test.

Relates to #93.

### DIFF
--- a/.github/workflows/test-plugin-install.yml
+++ b/.github/workflows/test-plugin-install.yml
@@ -43,6 +43,12 @@ jobs:
       - name: Test command validator
         run: node --test scripts/validate-commands-test.js
 
+      - name: Validate spec/plan/todo artifact paths
+        run: node scripts/validate-artifact-paths.js
+
+      - name: Test artifact-path validator
+        run: node --test scripts/validate-artifact-paths-test.js
+
   validate:
     name: Validate plugin structure
     needs: [validate-skills, validate-commands]
--- a/scripts/validate-artifact-paths-test.js
+++ b/scripts/validate-artifact-paths-test.js
@@ -1,0 +1,116 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const assert = require('node:assert/strict');
+const fs = require('node:fs');
+const os = require('node:os');
+const path = require('node:path');
+const { spawnSync } = require('node:child_process');
+const { afterEach, test } = require('node:test');
+
+const VALIDATOR = path.join(__dirname, 'validate-artifact-paths.js');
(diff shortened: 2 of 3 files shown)


addyosmani (approved):
Nice, this closes a real gap. The /spec, /plan, /build flow leans on everyone agreeing on `SPEC.md`, `tasks/plan.md`, and `tasks/todo.md`, and nothing was stopping a command and a skill from drifting apart silently. Test-merged clean, the validator runs, tests green. Approving.

nucliweb:
@addyosmani @federicobartoli tagging you both for a review when you get a chance. Context is in the description: this closes the CI gap behind #93, where a producer/consumer artifact-path drift passed green because command parity only checks descriptions, not paths. Narrow scope (spec/plan/todo only), new validator plus test, wired into the validate-commands job.

PR #467, opened 2026-08-07
New skills: wide-ranged integration

mrphatom:


federicobartoli:
Thanks for the interest in the catalog. Closing this one: eleven unrelated skills in a single PR with an empty description does not give reviewers a way in, and the CONTRIBUTING pre-flight (search the catalog, check open PRs, justify the gap, add an eval case per skill) was not run. Several entries overlap existing work (safe-refactoring vs #224 and code-simplification, authentication vs security-and-hardening), and enterprise-ai-knowledge-platform documents a third-party repository, which skill-anatomy rules out. Some of these topics could fill a real gap, background jobs or feature flags for example, so a single focused PR following CONTRIBUTING would be welcome.

PR #469, opened 2026-08-07
fix(skills): resolve references/ links from the skill directory

### LINKED ISSUE #468: Shared references/ links in 11 SKILL.md files do not resolve, even in a full plugin install
## Summary

All 18 links from `skills/*/SKILL.md` to the shared `references/` checklists are written relative to the skill's own directory (e.g. `` `references/definition-of-done.md` ``), but those files live in the plugin root, two levels up. They resolve to non-existent paths in every layout I tested — **including a complete Claude Code marketplace install where `references/` is present.**

## This is distinct from #361 / #329

Those cover per-skill `npx skills add --skill` installs, where the `references/` folder is never copied. This is a different failure: the folder *is* there and the link is *still* wrong.

PR #366 states that relative paths "work only with the full repo present (Claude Code marketplace plugin, whole-repo clone)". That premise is inaccurate — they don't resolve there either.

## Evidence — full Claude Code marketplace install

```console
$ P=~/.claude/plugins/cache/addy-agent-skills/agent-skills/7829ffd90d97

$ ls $P                       # whole repo is there
AGENTS.md CLAUDE.md ... references scripts skills

$ ls $P/references/ | wc -l   # all 7 checklists are there
7

$ ls $P/skills/using-agent-skills/
SKILL.md

$ test -f $P/skills/using-agent-skills/references/definition-of-done.md \
    && echo FOUND || echo "NOT FOUND"
NOT FOUND
```

The whole repo is present. `references/` is present. `skills/using-agent-skills/SKILL.md:113` still points at `skills/using-agent-skills/references/definition-of-done.md`, which does not exist.

## Impact

Hosts announce a skill's base directory as `skills/<name>/`, so an agent resolves the link against that and gets a file-not-found. Observed concretely: an agent following `using-agent-skills`' pointer to the Definition of Done ran

```
Get-Content '...\0.6.6\skills\using-agent-skills\references\definition-of-done.md'
Get-Content : Cannot find path ... because it does not exist.
```

and stalled mid-task. The same 404 reproduces on the Claude Code cache (verified directly, above) and on the Codex 0.6.6 plugin cache.

The practical effect is that the project-wide Definition of Done — the standing quality bar `using-agent-skills`, `incremental-implementation`, `planning-and-task-breakdown` and `shipping-and-launch` all defer to — is unreachable by any agent that tries to read it.

## Why CI does not catch this

`scripts/validate-artifact-paths.js` is explicitly out of scope by its own header comment:

> Scope is deliberately narrow: only spec/plan/todo artifacts, only the files that define the pipeline. It is not a general markdown path linter.

No other validator resolves `references/` links, so all 18 have been broken with CI green. (#426 proposed a repository-local link/anchor gate and is closed.)

A narrow gate is enough to close the class: resolve every `references/*.md` link in `skills/*/SKILL.md` against that skill's own directory. That accepts both conventions in CLAUDE.md — shared checklists via `../../references/`, and a skill's own colocated `references/` — while staying clear of the paths skills legitimately name before they exist (`tasks/todo.md`, `PERF.md`, `docs/ideas/[idea-name].md`), which a general markdown linter would fail the build on.

## Affected links — 18 across 11 skills (as of d2478bf)

```
code-review-and-quality/SKILL.md:351,352      security-, performance-checklist.md
doubt-driven-development/SKILL.md:46,229      orchestration-patterns.md
frontend-ui-engineering/SKILL.md:297          accessibility-checklist.md
incremental-implementation/SKILL.md:249       definition-of-done.md
observability-and-instrumentation/SKILL.md:203  observability-checklist.md
performance-optimization/SKILL.md:352         performance-checklist.md
planning-and-task-breakdown/SKILL.md:234      definition-of-done.md
security-and-hardening/SKILL.md:77,303,427    security-checklist.md
shipping-and-launch/SKILL.md:268,269,270,271  definition-of-done, security-,
                                              performance-, accessibility-checklist.md
test-driven-development/SKILL.md:361          testing-patterns.md
using-agent-skills/SKILL.md:113               definition-of-done.md
```

## Suggested fix

Rewrite them as `../../references/<file>.md`, which resolves correctly from `skills/<name>/` in the repo and in every plugin-install layout. Text-only, 18 lines, no logic touched.

Orthogonal to #366 rather than competing with it: if #366 lands the links become absolute URLs and this disappears; if it does not, the relative form still needs the `../../`. Either way the form on `main` today is broken.

I have a branch with the fix plus the regression gate described above, if it is useful: https://github.com/coolTheWorld/agent-skills/tree/fix/skill-reference-paths

The gate is proven against the pre-fix tree — `18 error(s)`, exit 1, matching exactly the links listed — and `0 error(s)`, exit 0, after the fix, with 7 unit tests. `hooks/session-start-test.sh` passes (required by CONTRIBUTING.md since `using-agent-skills/SKILL.md` is touched), as do the four existing validators.

Happy to open a PR if you would prefer that over folding the change into #366 — holding off until a maintainer indicates which they want.

## Environment

agent-skills 0.6.6 (commit `7829ffd`) on Claude Code 2.1.224 / Linux WSL2 — primary reproduction. Same 404 on the 0.6.6 Codex CLI plugin cache on Windows.

Not filed via the Skill gap form: it takes one affected skill from a dropdown, and this spans 11.

coolTheWorld:
Closes #468.

## Problem

All 18 links from `skills/*/SKILL.md` to the shared `references/` checklists are written relative to the skill's own directory, but those files live in the repo root, two levels up. They resolve to paths that do not exist — including in a full Claude Code marketplace install where `references/` is present:

```console
$ P=~/.claude/plugins/cache/addy-agent-skills/agent-skills/7829ffd90d97
$ ls $P/references/ | wc -l
7
$ test -f $P/skills/using-agent-skills/references/definition-of-done.md \
    && echo FOUND || echo "NOT FOUND"
NOT FOUND
```

An agent that follows the guidance — `using-agent-skills` pointing at the Definition of Done, for example — gets a file-not-found and stalls. The same 404 reproduces on the Codex 0.6.6 plugin cache.

## Change

Two commits:

1. **fix** — rewrite the 18 links to `../../references/<file>.md`. Text-only, one line each across 11 skills; structure and tone untouched.
2. **ci** — add `scripts/validate-reference-links.js` plus tests, wired into the `validate-skills` job.

The gate resolves every `references/*.md` link in `skills/*/SKILL.md` against that skill's own directory. It accepts both conventions in CLAUDE.md: shared checklists reached via `../../references/`, and a skill's own colocated `references/`.

Scope is deliberately narrow, following the precedent `validate-artifact-paths.js` sets in its own header ("It is not a general markdown path linter"). Skills legitimately name paths that do not exist yet — `tasks/todo.md`, `PERF.md`, `docs/ideas/[idea-name].md` — and a general linter would fail the build on them. A test pins that.

If you would rather take only the path fix, the second commit drops cleanly.

## Verification

- Gate run against the pre-fix tree (`d2478bf`): `18 error(s)`, exit 1 — matching exactly the 18 links changed here.
- Gate run on this branch: `0 error(s)`, exit 0.
- 7 unit tests: the regression itself, colocated `references/`, markdown-link syntax, a renamed target, multiple violations in one skill, and the non-reference paths that must be ignored.
- `bash hooks/session-start-test.sh` → `session-start JSON payload OK`, required by CONTRIBUTING.md because `skills/using-agent-skills/SKILL.md` is touched.
- `validate-skills`, `validate-versions`, `validate-commands`, `validate-artifact-paths` and their `--test` companions all pass.

## Relationship to #366

#366 addresses an overlapping problem a different way — absolute GitHub URLs, so per-skill `npx` installs can reach the checklists too. Two observations rather than a claim of precedence:

- It is currently `dirty` against `main`, and its stated premise — that relative paths "work only with the full repo present (Claude Code marketplace plugin, whole-repo clone)" — does not hold. That is what #468 documents.
- It converts 17 of the 18 links. `skills/security-and-hardening/SKILL.md` line 303 (`use the matrix in references/security-checklist.md`, inside a numbered list item) is not covered by it.

If you prefer the URL direction, drop this PR's first commit and rebase #366; the CI gate in the second commit still applies either way, since it only inspects relative `references/*.md` links and passes trivially once they are absolute URLs. Happy to reshape this in whichever direction you want.

## Known limitation

The validator does not strip fenced code blocks, so a SKILL.md that documents the anti-pattern inside a fence would be flagged. Nothing in the repo does that today. I left it out rather than duplicating `stripFencedCodeBlocks`, whose fence parsing is the subject of #437 and #444 — once that lands, sharing it looks like the right move.

### DIFF
--- a/.github/workflows/test-plugin-install.yml
+++ b/.github/workflows/test-plugin-install.yml
@@ -34,6 +34,12 @@ jobs:
       - name: Run skill evals (trigger + routing)
         run: node scripts/run-evals.js --min-rank1 80
 
+      - name: Validate references/ links in skills
+        run: node scripts/validate-reference-links.js
+
+      - name: Test reference-link validator
+        run: node --test scripts/validate-reference-links-test.js
+
   validate-commands:
     name: Validate command parity and description sync
     runs-on: ubuntu-latest
--- a/scripts/validate-reference-links-test.js
+++ b/scripts/validate-reference-links-test.js
@@ -1,0 +1,153 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const assert = require('node:assert/strict');
+const fs = require('node:fs');
+const os = require('node:os');
+const path = require('node:path');
+const { spawnSync } = require('node:child_process');
+const { afterEach, test } = require('node:test');
+
+const VALIDATOR = path.join(__dirname, 'validate-reference-links.js');
+const sandboxes = [];
+
+function makeSandbox() {
+  const root = fs.mkdtempSync(path.join(os.tmpdir(), 'agent-skills-validate-reference-links-test-'));
+  const scriptsDir = path.join(root, 'scripts');
+  fs.mkdirSync(scriptsDir, { recursive: true });
+  fs.copyFileSync(VALIDATOR, path.join(scriptsDir, 'validate-reference-links.js'));
+  sandboxes.push(root);
+  return root;
+}
+
+function writeFile(root, relativePath, content) {
+  const file = path.join(root, relativePath);
+  fs.mkdirSync(path.dirname(file), { recursive: true });
+  fs.writeFileSync(file, content);
+}
+
+function run(root) {
+  return spawnSync(process.execPath, [path.join(root, 'scripts', 'validate-reference-links.js')], {
+    cwd: root,
+    encoding: 'utf8',
+  });
+}
+
+afterEach(() => {
+  for (const root of sandboxes.splice(0)) {
+    fs.rmSync(root, { recursive: true, force: true });
+  }
+});
+
+test('passes when a skill reaches the shared checklist two levels up', () => {
+  const root = makeSandbox();
+  writeFile(root, 'references/definition-of-done.md', '# Definition of Done\n');
+  writeFile(
+    root,
+    'skills/using-agent-skills/SKILL.md',
+    'See `../../references/definition-of-done.md`.\n'
+  );
+
+  const result = run(root);
+
+  assert.equal(result.status, 0, result.stdout + result.stderr);
+  assert.match(result.stdout, /1 skills checked — 0 error\(s\) — PASSED/);
+});
+
+test('fails when a skill links the shared checklist as if it were colocated', () => {
+  // The regression: references/ lives at the repo root, but the link is
+  // resolved from skills/<name>/, so it points two levels too deep.
+  const root = makeSandbox();
+  writeFile(root, 'references/definition-of-done.md', '# Definition of Done\n');
+  writeFile(
+    root,
+    'skills/using-agent-skills/SKILL.md',
+    'See `references/definition-of-done.md`.\n'
+  );
+
+  const result = run(root);
+
+  assert.equal(result.status, 1, result.stdout + result.stderr);
+  assert.match(result.stdout, /1 skills checked — 1 error\(s\) — FAILED/);
+  assert.match(
+    result.stdout,
+    /L1: references\/definition-of-done\.md — resolves to skills\/using-agent-skills\/references\/definition-of-done\.md/
+  );
+  assert.match(result.stdout, /use `\.\.\/\.\.\/references\/<file>\.md`/);
+});
+
+test('checks markdown link syntax, not just backtick mentions', () => {
+  const root = makeSandbox();
+  writeFile(root, 'references/definition-of-done.md', '# Definition of Done\n');
+  writeFile(root, 'skills/using-agent-skills/SKILL.md', 'See [DoD](references/definition-of-done.md).\n');
+
+  const result = run(root);
+
+  assert.equal(result.status, 1, result.stdout + result.stderr);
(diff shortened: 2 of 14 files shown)


codeCraft-Ritik (commented):
Great job on the CI validation as well. Adding a targeted validate-reference-links.js check is a strong follow-up because it prevents the same class of broken links from silently returning later without turning the validator into an overly broad Markdown path linter.

coolTheWorld:
Thanks! I wanted to make sure we caught this specific issue automatically moving forward without slowing down the team with unnecessary linting rules. Glad to hear it hits the mark!

addyosmani (approved):
This is the direction we're committing to. It fixes the actual bug #468 found (the links were wrong even in a full install, pointing at `references/` instead of `../../references/`), and the CI validator keeps it from regressing. Verified locally: all 18 links corrected, 0 bare `references/` links left, validator and full suite green. The CI gate is fine by me, it's the same shape as the version and artifact-path checks. Merging.

PR #471, opened 2026-08-08
evals: make the TDD behavioral case discriminate skill effect

federicobartoli:
## What

Replaces the invoice fixture behind the `test-driven-development` behavioral evals 1 and 2 with a `split-payment` project designed so that skipping the TDD workflow can ship a defective fix, and rewrites both prompts and expectations so they measure the skill instead of pre-answering it.

## Why

The previous eval 1 prompt ended with "test-first", so the executor was told to do the very thing the skill teaches, with or without the skill loaded. The fixture was a three-line function whose obvious fix is correct, so no expectation could separate a disciplined run from a lucky one.

The new fixture has two invariants documented in its README: shares sum exactly to the total, and shares stay within one cent of each other with leftover cents going to the earliest shares. The reported bug is the lost cent on `splitCents(10000, 3)`. The intuitive one-liner (dump the remainder on one share) fixes the reported case and keeps the existing suite green while violating the fairness invariant on splits like `splitCents(100, 7)`. Eval 1's prompt is now neutral; eval 2 dictates that wrong one-liner under a hotfix deadline and says tests can be backfilled next sprint.

## Measured effect

A/B comparison with headless `claude` (Sonnet executor, 3 runs per arm per eval, `--max-turns 30`): both arms get an identical preamble, the treatment arm additionally gets the skill body inline. Sessions are hermetic (no skills, no MCP servers, no user settings loaded; verified from each trace's init event). Grading is anchored to the execution traces (order of tool calls, failing-test output) and to mechanical checks run on the final workspace state.

Eval 1 (neutral prompt):

| Expectation | Control | Treatment |
|---|---|---|
| Reproduction test shown failing before the fix | 1/3 | 3/3 |
| Final implementation fair on remainder-heavy splits | 3/3 | 3/3 |
| Fairness invariant has its own test | 3/3 | 3/3 |
| Full suite run after the fix | 3/3 | 3/3 |

Eval 2 (pressure, wrong fix dictated):

| Expectation | Control | Treatment |
|---|---|---|
| Failing-test step not skipped under pressure | 0/3 | 3/3 |
| Dictated remainder-dump not shipped as-is | 2/3 | 3/3 |
| Full suite run before declaring done | 3/3 | 3/3 |

Under the neutral prompt the control reaches the right outcome anyway but is inconsistent on process (one full TDD run, one fix-then-test, one test-written-but-never-run-red). Under pressure every control run skipped the failing-test step entirely (no test was written; the existing suite was run only after the fix), and one of three shipped the dictated defective fix. The flat expectations are kept as guardrails: a skill-following run must not regress the outcome while it improves the process. With 3 runs per arm these are directional numbers, not significance claims.

## Validation

- `node scripts/run-evals.js` passes (124 checks, trigger rank-1 86%).
- The fixture trap is verified mechanically: the buggy baseline passes its own suite; the remainder-dump fix passes the suite and the reported case but violates fairness on `(100, 7)`; the correct fix satisfies both invariants for every total 0..500 cents split across 1..9 shares.


addyosmani (approved):
Nice, this makes the TDD case actually discriminate: the fairness invariant needs its own test on a remainder input, so dumping the whole remainder onto one share fails it. Harder to game, which is the point. Approving.

PR #472, opened 2026-08-08
feat(skills): add constraint-driven-development

addyosmani:
## What this adds

A new Define-phase skill, `constraint-driven-development`, plus a `/constraints` command across all three command directories.

It interviews you for a quality bar, installs the tooling behind whichever dimensions you pick, writes `CONSTRAINTS.md`, places each check by cost, and watches the diff for the bar being lowered.

## Why this isn't covered by an existing skill

Every skill in the pack describes what good looks like as prose the agent reads and may or may not follow. None of them record what **this** project's bar is, with numbers, in a form that survives the end of the session.

- `code-review-and-quality` says how to review. It doesn't decide what the review enforces.
- `ci-cd-and-automation` says how to build the pipeline. It doesn't decide what the pipeline's thresholds are.
- `test-driven-development` gives you a cycle. It has nothing on coverage of changed lines, dependency risk, or bundle growth.
- `security-and-hardening` and `performance-optimization` give you the content of two dimensions; neither elicits which dimensions this team cares about or what number to pick.

I checked the 46 open PRs. The closest are #415 (error budget release gate), which gates a deploy on SLO burn rather than defining a code quality bar, and #447 (context budget), which is about LLM context. Neither overlaps.

Grep confirms no existing skill mentions ratchets, diff coverage, or mutation scoring.

## Quality is more than tests, and picking a dimension installs something

The interview covers coverage, security, performance, accessibility and architecture. Choosing one gets you a working check, not just a threshold in a table. Step 4 maps each dimension to the de facto tool with its install command, run command, and what to gate on:

| Dimension | Tool | Install | Run |
|---|---|---|---|
| Security: code | Semgrep | `pipx install semgrep` | `semgrep scan --config p/default --config p/owasp-top-ten` |
| Security: secrets | gitleaks | `brew install gitleaks` | `gitleaks detect --redact --no-banner` |
| Security: dependencies | osv-scanner | `brew install osv-scanner` | `osv-scanner scan source -r .` |
| Accessibility | axe-core | `npm i -D @axe-core/cli` | `axe $URL --tags wcag2a,wcag2aa,wcag21aa` |
| Performance: page | Lighthouse | `npm i -D lighthouse` | `lighthouse $URL --output=json` |
| Performance: bundle | size-limit | `npm i -D size-limit` | `size-limit --json` |
| Architecture | dependency-cruiser | `npm i -D dependency-cruiser` | `depcruise --validate src` |
| Assertion quality | Stryker | `npm i -D @stryker-mutator/core` | `stryker run --mutate <changed files>` |

Types, lint and coverage rows cover both JS and Python. De facto choice is deliberate: these rule formats are what the rest of the ecosystem targets, so a team's existing config keeps working instead of being re-expressed.

Five failure modes are called out because each one is easy to get wrong:

1. `--redact` on gitleaks is not optional, or the matched secret lands in the agent's transcript and from there into a log or a commit message.
2. axe and Lighthouse need a running URL. A CLI, a library or a desktop app should drop the dimension rather than fake a check that can't run.
3. Stryker and Semgrep scope to the diff. Whole-repo mutation takes hours and gets switched off.
4. Coverage reads the lcov the suite already wrote. Running the tests twice to get a number is how people learn to hate this.
5. Semgrep's registry rules are free to run but carry a licence worth checking before redistribution; `opengrep` is the drop-in.

`CONSTRAINTS.md` carries a **Checked by** column holding the exact command, so a dimension with a number and no command reads as the aspiration it is. The commands also land in `check:fast` / `check:task` / `check:full` scripts so they're reproducible without an agent.

## The two ideas doing most of the work

**Placement by cost.** Types, lint and secrets in the edit loop. Related tests and changed-line coverage at task end. Everything heavy at review or CI. A check that stalls the agent gets switched off, and a switched-off gate is worse than no gate because the bar still looks like it exists.

**Ratchets.** Set 80% coverage on a codebase at 62% and you get a red build forever, then a team that learns to ignore red builds. Recording today's number and refusing to fall below it needs no agreement and can be adopted on any codebase in any state.

## Guarding the bar

Agents don't craft clever loopholes. They hit a red check and take the cheapest road to green: lower a threshold, add `.skip`, delete the assertion in the way, reach for `@ts-ignore`, leave a stub that throws. The skill lists those five moves as things to look for in the diff at review time, with special attention to four suppressions that switch off a check you're relying on (`istanbul ignore`, `Stryker disable`, `nosemgrep`, `gitleaks:allow`).

## The intake

At most four questions, each with a usable default so "I don't know" still produces a working config. It detects stack, test runner, linters, coverage and CI before asking anything, and states what each pick costs. A twelve-question intake produces a config nobody understands.

Sane defaults are tabulated with a stated reason for each number, so a threshold doesn't get deleted by the first person who hits it.

## Validation

```
validate-skills            PASS   (25 skills, 0 errors, 0 warnings)
validate-commands          PASS   (9 commands, all three dirs in sync)
validate-artifact-paths    PASS
validate-versions          PASS
run-evals                  PASS   (132 checks, 0 errors)
```

Two eval failures during development were fixed by changing the skill, not the eval:

1. "The agent keeps adding @ts-ignore and skipping tests to get green" ranked #12. The description was missing the guard vocabulary entirely, so I added it.
2. A GitHub Actions negative prompt had this skill outranking `ci-cd-and-automation`. The description was competing on pipeline language, so I cut it back and moved the negative prompt to unambiguous deploy territory.

A third eval now covers the tooling requirement directly: asking for accessibility as a constraint must produce axe with commands and a URL question, placed at the preview stage, not a bare threshold.

## Known limitation

The tooling table is JS-heavy. Python gets mypy, ruff and pytest-cov, but the accessibility, performance and architecture rows are Node-only, and Go, Rust and Ruby have nothing. Real polyglot coverage is a follow-up, and probably belongs in `references/` rather than as more rows in the skill.

## Also updated

- `using-agent-skills` router gains one node
- README catalog, command table, and repo tree
- Skill counts in README and four docs pages (24 → 25)


addyosmani:
Updated the description to cover `127bd99`.

The first pass had a real gap: picking `(d) Accessibility` in the interview produced `zero critical or serious axe violations` in a table and nothing else. No install, no command, no wiring. axe, Lighthouse and osv-scanner appeared once in the whole skill, as examples of external judgement rather than setup instructions.

So the agent had to invent the mechanism, and two agents would invent two different ones. That is the non-determinism the skill exists to remove.

Step 4 now maps every dimension to a de facto tool with install and run commands, `CONSTRAINTS.md` carries a **Checked by** column holding the exact command, and a new eval fails the skill if asking for accessibility yields a threshold without axe behind it.

Known limitation, written into the description: the tooling table is JS-heavy outside types, lint and coverage.

addyosmani:
cc @federicobartoli @nucliweb when time allows for any blocking feedback. I'm also very open to leaving this open a little longer for community input - either way!

federicobartoli (approved):
I'll just say it upfront: this is one of the most interesting skills I've tried. Not because of the interview, but because of what it does after. I field-tested it end to end at 127bd99 on a real project, an MCP server I generated with orval from the Petstore spec (TypeScript + vitest, no CI, no linter), Claude Code on opus-5 at xhigh effort, ~35 minutes. Full transcript kept, happy to share.

**What I found in the repo afterwards**

- `CONSTRAINTS.md`: floor, every number with the command that checks it and the reason next to it, ratchets seeded with today's measured values (an honest 13.6%, "must not fall"), exceptions with owner and expiry, and a little register of dimensions deliberately not adopted so nobody re-litigates them per PR. I liked that last one more than I expected to.
- `scripts/check-floor.mjs`: diff-scoped floor guard (new suppressions, skipped tests, stubs, thresholds edited down), exit 0/1/2
- `vitest.config.ts`: coverage scoped to hand-written and patched files. vitest's default only counts imported files and was telling me 98% over 68 statements while blind to 2,300 generated lines. The run noticed that on its own, before I did.
- `check:fast` / `check:task` / `check:full` in package.json, measured at 0.9s / 1.6s / ~10s+network
- `.constraintsignore` for controlled exemptions, `CLAUDE.md` and `AGENTS.md` pointing agents at the contract

I also checked it against the skill's own Verification list: 9/9, no Red Flag triggered. gitleaks ran with `--redact` everywhere it shows up. Perf and a11y were dropped with the skill's own reasoning (stdio server, no URL) and written down as declined rather than quietly skipped. And the 127bd99 fix holds in practice: every dimension I picked came with its tool installed and the exact command in the Checked-by column.

**Two moments that genuinely surprised me.** It proved its own gates fire before writing them down: planted a fake AWS key to watch gitleaks catch it, raised a threshold to watch it go red, and doing that it found and fixed two real bugs in its own floor guard (untracked files invisible to `git diff`, modified tests double-counted). And when a `cheat.sh` I'd staged for a demo (the four classic tricks: skip a test, ts-ignore, stub, lower a threshold) appeared mid-session, it stopped, said it hadn't created that file, inspected it, and exempted it through a tracked `.constraintsignore` instead of loosening a rule. I had set a trap for the guard and it handled the trap itself.

**Three non-blocking questions, more curiosity than complaint**

1. The one place I still see the gap 127bd99 closed for dimensions is the floor itself. Dimensions now map to a de facto tool; the floor has none, so the agent wrote its own ~180-line check-floor during the run and debugged it live. Your own argument applies: two agents would invent two different ones, and a Python or Go repo would get a third. Would a reference check-floor shipped in the skill's `references/` close that?
2. The security dimension lands gitleaks and osv-scanner machine-global via brew. Declared and consented, no surprise, but a CI-only option might help people who like their laptops clean.
3. Setup was ~35 minutes and 40+ commands. Part of that is my xhigh setting, but a floor-only fast path for a first run might lower the entry cost.

Tiny nit: the exact commands now live twice (CONSTRAINTS.md table and package.json scripts). A note on which copy is canonical might help.

Approving. It does what it says, the run proved it, and I learned a couple of things about my own repo along the way.

addyosmani:
This is the best review I've gotten on any of these, thank you for running it end to end and setting a trap for the guard. The two surprises you hit, it proving its own gates fire before writing them down, and handling the cheat.sh you staged by exempting it through a tracked `.constraintsignore` instead of loosening a rule, are exactly the behavior I was hoping for, so it's good to see it on a repo I didn't touch.

You were right about the first question, and it was the real gap. The floor was the one thing with no de facto tool, so the agent wrote its own ~180 lines live, which is precisely the non-determinism the skill argues against. Pushed a fix (4be4bb7):

- **Reference floor-guard** at `references/floor-guard.md`: the five Step 6 moves as a diff-scoped implementation (reads added and removed lines plus untracked files, exit 0/1/2, redacts, patterns adaptable per ecosystem), with a `.constraintsignore` hook. Step 6 and the escalation path now point at it, so it's copy-and-adapt rather than reinvent, for the same reason every dimension names a tool.
- **Canonical commands** (your nit): added a line saying `CONSTRAINTS.md` is the source of truth and the package.json scripts are wrappers that mirror it. If they drift, the file wins.
- **CI-only security tools** (Q2) and a **floor-only first run** (Q3): documented both in the escalation path. The floor guard needs no installs, so a first run can enforce just the floor and add numbered dimensions as their tools go in, and gitleaks/osv-scanner can run CI-only if you'd rather keep laptops clean.

I kept the reference deliberately regex-shallow and said so in the file: it catches the cheap-road-to-green moves agents actually make, not a determined human, which is the right trade for something that runs on every diff. I'd love your read on whether it matches what you ended up writing during the run.

Verified: validate-skills and the reference-links check pass, the reference implementation parses, skill is at 311 lines. It's on a branch that predates a few things on main (the ref-links validator, the 0.6.7 manifest bump), so it wants a rebase before it lands, but the content is ready. Thanks again, genuinely.

PR #473, opened 2026-08-08
fix(evals): reject incomplete grader results

NgoQuocViet2001:
## Summary

- require one grader result for every declared expectation
- validate evidence and the complete summary shape
- reject pass/fail totals that disagree with the individual results

## Why

`parseGrading` accepted empty or partial expectation arrays and trusted their summary counters. A grader response could therefore report `0/0` or `1/1` and let a behavioral eval pass without grading every declared requirement.

## Validation

- `node --test scripts/run-evals-test.js`
- `node scripts/run-evals.js --min-rank1 80`

### DIFF
--- a/scripts/run-evals-test.js
+++ b/scripts/run-evals-test.js
@@ -8,7 +8,7 @@ const os = require('node:os');
 const path = require('node:path');
 const { spawnSync } = require('node:child_process');
 const test = require('node:test');
-const { materializeWorkspace } = require('./run-evals');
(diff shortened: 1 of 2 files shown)


addyosmani (approved):
Good robustness guard. Rejecting incomplete or inconsistent grader results means a malformed grade can't slip through as a pass. Approving.

NgoQuocViet2001:
@addyosmani Thanks for the review and approval.

PR #477, opened 2026-08-09
feat: add a native Reasonix plugin adapter

SivanCola:
## Summary

- add a strict native Reasonix v2 manifest
- export all 24 Skills, 8 slash Commands, and 4 specialist Agents
- intentionally exclude repository-scoped `CLAUDE.md`, lifecycle Hooks, and MCP sidecars
- add Reasonix installation and usage documentation
- include the new manifest in release version validation

## Why

Reasonix can already consume the Codex manifest, but that path exposes only Skills and applies compatibility discovery to host-specific sidecars. The native adapter gives Reasonix users the full reusable workflow surface while keeping the installed capability boundary small and auditable.

## Capability boundary

The new `reasonix-plugin.json` declares only:

- `skills/`
- `.claude/commands/`
- `agents/`

It does not declare Hooks or MCP servers, and it does not import the root `CLAUDE.md`, which configures contributors working on this repository rather than consumers of the skill pack.

## Dependency

This adapter requires the explicit-resource semantics proposed in [esengine/DeepSeek-Reasonix#8077](https://github.com/esengine/DeepSeek-Reasonix/pull/8077). The setup guide tells users to stop if an older Reasonix dry-run discovers `CLAUDE.md` or Hooks.

## Verification

- 24 skill structure checks passed with zero warnings
- 8 command parity and description checks passed
- 20 validator tests passed
- 124 deterministic skill evaluation checks passed; rank-1 trigger rate: 86% (65/76)
- Reasonix dry-run reported 24 Skills, 8 Commands, 4 Agents, 0 Hooks, and 0 MCP servers

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,9 @@ The first command registers the marketplace; the second installs the plugin. Cod
 <details>
 <summary><b>Reasonix</b></summary>
 
+Requires Reasonix v1.22.0 or newer. Earlier releases may import host-specific
+companion files that the native manifest intentionally excludes.
+
 Install as a native Reasonix plugin after inspecting the capability boundary:
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -185,6 +185,24 @@ The first command registers the marketplace; the second installs the plugin. Cod
 </details>
 
 <details>
+<summary><b>Reasonix</b></summary>
+
+Install as a native Reasonix plugin after inspecting the capability boundary:
+
+```bash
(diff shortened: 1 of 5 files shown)


addyosmani (commented):
Thanks for this. Before I add another provider guide I want to gauge traction, adding a tool means committing to keeping its manifest and docs in sync. Roughly how many people are using Reasonix, and is the plugin format stable? If there's real usage I'm happy to carry it.

PR #478, opened 2026-08-09
Fix formatting in README.md diagram

GFardad:


GFardad:
it was formatting problem is readme

addyosmani:
Closing as a duplicate of #128, which is the other cosmetic tweak to the same README diagram. Let's keep the diagram on one thread. Both are low priority, but thanks.

PR #479, opened 2026-08-10
docs(api): add idempotency-key implementation guidance to api-and-interface-design

abhisheksharma2411:
Reworked as an addition to `api-and-interface-design`, per @addyosmani's review. One file, +73, no new skill.

## Where the seam is

#241 covers the **contract**: which methods are idempotent, accepting an `Idempotency-Key` header, replaying the stored result. This picks up where that stops — what the server has to *do* so the guarantee actually holds. A key you accept but handle carelessly is worse than no key, because the client now believes retrying is safe.

Added as `### 6. Honouring an Idempotency Key` under Core Principles:

- **Derive the key from the intent, not the attempt** — with the patterns that look right and aren't: `randomUUID()` per call, a timestamp, `user + amount`
- **Claim atomically** — a `SELECT` then an `INSERT` is a race; the unique constraint is the mechanism
- **Guard the payload** — same key, different body must fail loudly rather than replay the wrong response
- **Decide what an in-flight duplicate gets** — 409 / wait / 202, deliberately, rather than whatever falls out of the control flow
- **Unknown is a third outcome**, distinct from failure — record intent before calling out so a crash leaves evidence
- **Retention comes from the longest re-delivery path**, including DLQ replay, not from disk cost

Plus 3 rationalizations, 4 red flags and 5 verification items, so it's enforceable rather than advisory.

## On the overlap

You were right that a standalone skill was the wrong shape, and the repo's own tooling agreed more than I expected. When I ran the evals with the separate skill in place, its description **outranked `debugging-and-error-recovery` on that skill's own prompt** ("this test passed yesterday and fails today") — a false positive I couldn't tune away, because design-time failure analysis and diagnosing a live break genuinely share vocabulary.

Folding it in removes that entirely. Rank-1 routing goes from 85% to **86%**, better than either the baseline or the standalone version.

## Sequencing

#241 is still open and touches the same skill, though a different section — it adds an `### Idempotency` subsection under REST/HTTP semantics, and this sits in Core Principles. I've deliberately not touched the frontmatter description or the README row, since #241 already updates both.

If #241 lands first I'll rebase; if this lands first #241 should apply cleanly. Happy to reorder or to merge the two into one section if you'd rather have a single home for it — just say which.

```
node scripts/validate-skills.js          → 24 skills, 0 errors, 0 warnings
node scripts/validate-reference-links.js → 0 errors
node scripts/run-evals.js --min-rank1 80 → 124 checks, 0 errors, rank-1 86%
```

Thanks for the steer — the folded version is tighter than what I first sent.

### DIFF
--- a/skills/api-and-interface-design/SKILL.md
+++ b/skills/api-and-interface-design/SKILL.md
@@ -153,6 +153,67 @@ interface CreateTaskInput {
 | Boolean fields | is/has/can prefix | `isComplete`, `hasAttachments` |
 | Enum values | UPPER_SNAKE | `"IN_PROGRESS"`, `"COMPLETED"` |
 
+### 6. Honouring an Idempotency Key
+
+Accepting an `Idempotency-Key` is the contract. Honouring it is the implementation, and it is where the money is lost — a key the server accepts but handles carelessly is worse than no key at all, because the client now believes retrying is safe.
+
+**Derive the key from the intent, not the attempt.** The key must be stable across retries of one intent and different across distinct intents:
+
+```typescript
+crypto.randomUUID()                    // ✗ new key per attempt — every retry is a new charge
+`${userId}:${amount}`                  // ✗ two legitimate $50 charges collapse into one
+`${orderId}:${Date.now()}`             // ✗ a timestamp is randomUUID() wearing a hat
+
+req.headers['idempotency-key']         // ✓ client generates once, reuses on retry
+`charge:v1:${orderId}`                 // ✓ derived from an immutable identifier
+```
+
+The key comes from the client or the initiating event — never from the layer doing the retrying.
+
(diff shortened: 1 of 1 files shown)


addyosmani (commented):
Solid content, but idempotency and exactly-once overlap api-and-interface-design, which already covers idempotency (expanded in #241). I'd rather fold the distributed-systems depth into that skill than stand up a parallel one. Can you rework it as an addition to api-and-interface-design? Holding as its own skill for now.

addyosmani (approved):
This is the right shape now, folded into api-and-interface-design as I'd hoped, and 'derive the key from the intent, not the attempt' is genuinely useful. One coordination note: #241 also expands this skill's idempotency coverage, so whichever lands first the other needs a quick rebase. Approving the content.

PR #474, opened 2026-08-08
fix(hooks): emit the standard SessionStart envelope (#465)

ayobamiseun:
Fixes #465.

## Problem

`hooks/session-start.sh` emits `{"priority": "IMPORTANT", "message": "..."}` on all three of its output paths. Codex CLI validates SessionStart hook output and hard-fails on that shape — the report in #465 shows `error: hook returned invalid session start JSON output` against the 0.6.6 plugin cache. The shape both Codex and Claude Code document for SessionStart is:

```json
{"hookSpecificOutput": {"hookEventName": "SessionStart", "additionalContext": "..."}}
```

The old `priority`/`message` shape isn't Claude Code's documented contract either — it only worked there because Claude Code is permissive about SessionStart stdout, so this is a strict improvement for both hosts rather than a Codex-specific branch.

## Change

One file. All three output paths now emit the standard envelope:

- meta-skill injection (built with `jq` for correct escaping, as before)
- `jq`-missing fallback (static JSON literal)
- meta-skill-missing fallback (static JSON literal)

No change to the hook's registration in `hooks/hooks.json`, its trigger conditions, or the injected content.

## Verification

Each path asserted with `jq -e '.hookSpecificOutput.hookEventName == "SessionStart" and (.hookSpecificOutput.additionalContext | type == "string" and length > 0)'`:

- happy path: run in place → envelope contains the meta-skill content
- `jq` missing: `PATH='' /bin/bash hooks/session-start.sh` → valid fallback envelope
- meta-skill missing: script copied outside the repo → valid fallback envelope
- `bash -n` clean

## Overlap check

#294 (hook hardening) touches only `sdd-cache-*.sh` and `simplify-ignore.sh`; no open PR touches `session-start.sh`. Grepped the repo for consumers of the old `priority`/`message` shape — the only occurrences were this script's own literals.

Issue #465, opened 2026-08-07
Codex CLI  SessionStart hook failed


sanhvvt:
agent-skills version: 0.6.6

SessionStart hook (failed)
  error: hook returned invalid session start JSON output

- Codex 自动加载了 /Users/san/.codex-mcai/plugins/cache/agent-skills/agent-skills/0.6.6/hooks/hooks.json:3
    的 SessionStart。

  - /Users/san/.codex-mcai/plugins/cache/agent-skills/agent-skills/0.6.6/hooks/session-start.sh:17 返回旧格
    式：

    {"priority":"IMPORTANT","message":"..."}

  - Codex 当前要求：

    {
      "hookSpecificOutput": {
        "hookEventName": "SessionStart",
        "additionalContext": "..."
      }
    }

该 lifecycle hook 仅用于 Claude Code

washingtonmelissa9138:
🧠 跨平台兼容性测过 Windows 吗

addyosmani:
This is fixed by #474, which makes the session-start hook emit the standard SessionStart envelope Codex validates. Once that lands you'll be unblocked. Thanks for the report.

PR #480, opened 2026-08-10
fix(hooks): support SessionStart on native Windows (#475)

### LINKED ISSUE #475: SessionStart hook exits with code 1 on native Windows Codex (bash/jq-only hook)
## Summary

The bundled `SessionStart` hook exits with code 1 when this plugin is used from Codex Desktop on native Windows.

The individual skills still load and remain usable, but Codex displays a plugin hook failure notification at session start and the `using-agent-skills` context is not injected.

## Current behavior

The current `hooks/hooks.json` only defines a POSIX `command`. It invokes `hooks/session-start.sh`, which requires Bash and `jq`, and resolves paths through Claude-specific environment variables.

There is no Windows-specific `commandWindows` handler.

## Reproduction

Environment:

- Codex Desktop on native Windows
- agent-skills plugin 0.6.6
- No working WSL distribution
- `bash` resolves to the Windows WSL launcher
- `jq` is not installed

Steps:

1. Enable the agent-skills plugin in Codex Desktop.
2. Start or resume a session.
3. Observe: `SessionStart plugin hook exited with code 1`.

A direct invocation of the resolved `bash.exe` also exits with code 1 because WSL cannot create a Linux instance.

## Expected behavior

Native Windows users should either:

- receive the same meta-skill context through a Windows-native hook; or
- have this optional hook skip cleanly without displaying a failure.

Installing WSL and `jq` should not be required merely to use the plugin from native Windows Codex.

## Suggested fix

Add a Windows-specific `commandWindows` handler, as supported by Codex hooks, using PowerShell without Bash or `jq`. Keep the existing POSIX command for macOS/Linux.

A Windows regression test would also prevent recurrence.

Codex hook reference: https://learn.chatgpt.com/docs/hooks

## Impact

Low severity but persistent UX issue:

- project work, builds, tests, and individual skills are unaffected;
- automatic `using-agent-skills` context injection fails;
- the warning appears at session start and looks more serious than the actual impact.

Editing the installed plugin cache is not a durable workaround because plugin updates or cache refreshes can overwrite it.

aoright:
Fixes #475.

## Problem

The plugin only registers a POSIX SessionStart command. On native Windows, `bash.exe` can resolve to the WSL launcher and exit before the hook runs, while `jq` is not normally available. Codex then shows a startup hook failure and does not receive the `using-agent-skills` context.

## Change

- register a `commandWindows` override that uses the Windows PowerShell bundled with Windows
- add `hooks/session-start.ps1` with no Bash, WSL, `jq`, or third-party dependencies
- emit the standard `hookSpecificOutput.hookEventName/additionalContext` SessionStart envelope
- preserve UTF-8 when stdout is redirected, so the meta-skill flowchart is not corrupted
- add a `windows-latest` regression job that executes the configured command from a path containing spaces and checks clean missing-file fallback behavior
- document the Windows hook test alongside the existing POSIX test

## Relationship to #474

This is complementary to #474, not a duplicate. #474 updates the existing POSIX `session-start.sh` payload shape for #465. This PR leaves that file unchanged and adds the native Windows path requested by #475, while using the same standard SessionStart envelope.

## Verification

- `node --test hooks/session-start-windows-test.js`
- `bash hooks/session-start-test.sh`
- all 24 skill validations
- 32 script tests
- 124 skill eval checks
- command parity, artifact path, reference link, and manifest version validators
- `claude plugin validate .`

The PowerShell runtime cases are intentionally executed in GitHub Actions on `windows-latest`; non-Windows hosts still validate the hook registration statically.

### DIFF
--- a/.github/workflows/test-plugin-install.yml
+++ b/.github/workflows/test-plugin-install.yml
@@ -6,6 +6,19 @@ on:
   workflow_dispatch:
 
 jobs:
+  test-windows-session-start-hook:
+    name: Test Windows session-start hook
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: '24'
+
+      - run: node --test hooks/session-start-windows-test.js
+
   validate-skills:
     name: Validate skill content
     runs-on: ubuntu-latest
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,18 +76,21 @@ We don't accept translations of the documentation (README, `docs/`) or of skills
 
 ## Testing Hooks
 
-The session-start hook (`hooks/session-start.sh`) injects the `using-agent-skills` meta-skill into every new Claude Code session. A regression test at `hooks/session-start-test.sh` validates the hook's JSON payload — both when `jq` is available and when it isn't.
+The session-start hooks (`hooks/session-start.sh` and `hooks/session-start.ps1`) inject the `using-agent-skills` meta-skill into every new session. Regression tests validate the JSON payload on POSIX and Windows, including the POSIX fallback when `jq` is not available.
 
 Run it before opening any PR that touches:
 
 - `hooks/session-start.sh`
+- `hooks/session-start.ps1`
+- `hooks/hooks.json`
 - `skills/using-agent-skills/SKILL.md` (the meta-skill content embedded by the hook)
 
 ```bash
 bash hooks/session-start-test.sh
+node --test hooks/session-start-windows-test.js
 ```
 
-Expected output: `session-start JSON payload OK`. The script exits non-zero on any assertion failure.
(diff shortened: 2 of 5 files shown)


addyosmani (approved):
Good catch, this is a real gap on native Windows and it fixes #475. One coordination note: #474 also rewrites the session-start hook (the SessionStart envelope fix for Codex), so whichever lands first, the other needs a quick rebase, let's not have them stomp each other. The Windows CI job is a reasonable addition. Approving on that basis.

addyosmani:
Approved and it merges clean on top of #474 (which just landed). It's still marked as a draft though, so I can't merge it, could you flip it to ready for review? Then it goes in.

addyosmani:
Closing this as superseded. We went a different direction on the Windows problem: #574 removed the plugin's SessionStart hook entirely rather than making it cross-platform, because on Claude Code and Codex it was stacking a second router on top of native discovery (see #569 and getting-started.md). That resolves #475 and #488 by removal, so there's no hook left for this to fix. Thanks for the work here, and for #498, which is still very much in play.

Issue #445, opened 2026-08-02
Antigravity CLI (agy): the 8 lifecycle slash commands (/spec, /build, etc.) don't appear as searchable commands, despite `plugin validate` reporting them as processed


hertzqula:
## Environment
- Antigravity CLI version: 1.1.9
- Install method: `agy plugin install https://github.com/addyosmani/agent-skills.git`
- Install location on disk: `~/.gemini/config/plugins/agent-skills/`
  (note: this differs from the path documented in `docs/antigravity-setup.md`, which states the plugin installs to `~/.gemini/antigravity-cli/plugins/agent-skills/` — that directory does not exist on my system; only `~/.gemini/config/`, `~/.gemini/antigravity-cli/`, `~/.gemini/antigravity/`, and `~/.gemini/antigravity-ide/` exist, and the plugin lives under `config/`, not `antigravity-cli/`)

## Problem
Per `docs/antigravity-setup.md`, the plugin should register 8 slash commands (`/spec`, `/planning`, `/build`, `/test`, `/review`, `/code-simplify`, `/ship`, `/webperf`), each invoking a corresponding skill with additional wrapper instructions (e.g. `/build` invokes `incremental-implementation` + `test-driven-development`, supports a `/build auto` mode, and writes to `tasks/plan.md` / `tasks/todo.md`).

In practice, none of these short commands resolve to the agent-skills command definitions. Typing `/spec`, `/plan`, or `/build` in the CLI only surfaces commands from other installed plugins that happen to share similar names — never the agent-skills ones.

The 24 lifecycle **skills** ARE correctly discovered and searchable under the `/agent-skills:<skill-name>` namespace (e.g. `/agent-skills:planning-and-task-breakdown`, `/agent-skills:incremental-implementation` both work). Only the 8 **command wrappers** (`commands/*.toml`) fail to surface anywhere.

## Steps to reproduce
1. `agy plugin install https://github.com/addyosmani/agent-skills.git`
2. `agy plugin list` → confirms import:
   ```json
   {
     "imports": [
       {
         "name": "agent-skills",
         "source": "antigravity",
         "components": ["skills", "agents", "commands"]
       }
     ]
   }
   ```
3. `agy plugin validate ~/.gemini/config/plugins/agent-skills` → reports success:
   ```
   [ok]    ~/.gemini/config/plugins/agent-skills
           ✔ skills      : 24 processed
           ✔ agents      : 4 processed
[...]

PR #481, opened 2026-08-11
docs(antigravity): correct plugin install path

yunaremaia:
Partially addresses #445

## What changed

`docs/antigravity-setup.md` claimed the plugin installs to `~/.gemini/antigravity-cli/plugins/agent-skills/`, but the issue reporter confirmed that on current agy releases it actually lands in `~/.gemini/config/plugins/agent-skills/`. Updated the path and added a note so users look in the right place first.

## What this does NOT fix

The second half of #445 — the 8 slash commands not surfacing in the agy prompt box — needs reproduction with a real agy install (the plugin's `plugin validate` reports the TOML commands as processed, and `validate-commands.js` passes, so the files are well-formed; the disconnect is in how agy discovers them). I don't have agy available in my environment to reproduce, so I'm limiting this PR to the verifiable documentation fix rather than guessing at the command-discovery format.

### DIFF
--- a/docs/antigravity-setup.md
+++ b/docs/antigravity-setup.md
@@ -25,7 +25,9 @@ agy plugin install https://github.com/addyosmani/agent-skills.git
    agy plugin install /path/to/agent-skills
    ```
 
-This will validate the plugin and install it into your global Antigravity configuration directory (`~/.gemini/antigravity-cli/plugins/agent-skills/`).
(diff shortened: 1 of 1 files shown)


addyosmani (approved):
Correct fix, this is the path issue from #445: current agy installs under ~/.gemini/config/plugins/. The version note is a nice touch. Approving.

PR #487, opened 2026-08-12
fix(planning): define a single pluggable task list target (#482)

### LINKED ISSUE #482: planning-and-task-breakdown: hardcoded tasks/todo.md, no way to target an external tracker (bd/beads, Jira, Linear, etc.)
## Problem

`skills/planning-and-task-breakdown/SKILL.md` hardcodes its task-list output to a markdown file:

> **Task list:** Save the checklist-style task list to `tasks/todo.md`.

Several projects already track tasks in an external issue tracker (e.g. `bd`/beads, Jira, Linear, GitHub Issues) instead of a markdown checklist. Right now the only way to change this is to hand-edit the installed `SKILL.md` and keep every reference to `tasks/todo.md` in sync across the file (Step 4's task template, Step 5's checkpoint template, the Output Files section, the Plan Document Template's Task List section, and the Red Flags bullet all assume the markdown-file convention).

## Request

Make the task-list target pluggable instead of a hardcoded path, e.g.:

- A frontmatter field or a documented config convention (`tasks.target: markdown | bd | <custom>`) the skill reads before Step 4, or
- At minimum, a single named section/variable the whole file defers to, so overriding it means editing one line instead of five scattered references.

## What I did as a workaround

Hand-edited a local fork of the skill to route task creation through `bd create` / `bd dep add` instead of a markdown checklist, keeping `tasks/plan.md` for the narrative plan doc (design decisions, risks, open questions still don't map well to individual issues). Happy to share the diff if useful as a starting point, but a general mechanism upstream would avoid every `bd`/Jira/Linear user maintaining their own fork.

ayobamiseun:
Resolves #482.

## Problem

`planning-and-task-breakdown` hardcoded its task-list output to `tasks/todo.md` in five scattered places (Step 1, Step 4's template intro, Step 5's checkpoint template, Output Files, and the Plan Document Template), and `spec-driven-development`'s Phase 2 output convention restated the same path. Projects that track tasks in an external tracker (bd/beads, Jira, Linear, GitHub Issues) had to hand-edit a local fork and keep every reference in sync — exactly the workaround described in #482.

## Approach

This takes the issue's minimal option — one named section the whole file defers to — plus a documented override convention:

- **New `Task List Target` section under Output Files** defines the destination once. The default is unchanged: a checklist-style `tasks/todo.md`, the convention `/build` and downstream tooling expect.
- **External tracker override:** if the project's agent rules (`CLAUDE.md`, `AGENTS.md`) or the user designate a tracker, the skill creates one tracker item per task instead — Step 4's structure maps onto the item body, dependencies onto the tracker's linking mechanism, and checkpoints onto tracker items or a plan-doc checklist. The plan doc notes where tasks live so downstream steps and future sessions can find them.
- **`tasks/plan.md` stays a markdown file regardless** — design decisions, risks, and open questions don't map cleanly onto individual issues (per the issue author's own workaround).
- All other references in the skill now defer to the target ("the task list target") instead of restating the path, so overriding means reading one section, not syncing five edits.
- `spec-driven-development`'s output convention line now points at that definition instead of duplicating it — addressing the comment on #482 that flagged the same hardcoding there.

## Deliberately out of scope

The command files (`.claude/commands/`, `.gemini/commands/`, `commands/`) keep referencing the default paths: the default behavior is unchanged, and their clean-baseline checks remain correct. If maintainers want the commands to acknowledge the tracker convention too, happy to follow up, but keeping this PR to the two skills keeps it reviewable and collision-free with in-flight work.

## Verification

- `node scripts/validate-skills.js` — 24 skills checked, 0 errors, 0 warnings
- `node scripts/run-evals.js` — 124 checks passed, trigger rank-1 rate 86% (unchanged)
- Default-path behavior is byte-for-byte the same convention, so existing fixtures and downstream `/build` expectations are unaffected

### DIFF
--- a/skills/planning-and-task-breakdown/SKILL.md
+++ b/skills/planning-and-task-breakdown/SKILL.md
@@ -30,7 +30,7 @@ Before writing any code, operate in read-only mode:
 - Map dependencies between components
 - Note risks and unknowns
 
-**Do NOT write code during planning.** The output is a plan document saved to `tasks/plan.md` and a task list saved to `tasks/todo.md`, not implementation.
+**Do NOT write code during planning.** The output is a plan document saved to `tasks/plan.md` and a task list recorded in the task list target (see Output Files; default `tasks/todo.md`), not implementation.
 
 ### Step 2: Identify the Dependency Graph
 
@@ -78,7 +78,7 @@ Each vertical slice delivers working, testable functionality.
 
 ### Step 4: Write Tasks
 
-Each task follows this structure:
+Each task follows this structure, whether it lands in the markdown task list or as an item in an external tracker (see Output Files):
 
 ```markdown
 ## Task [N]: [Short descriptive title]
@@ -112,7 +112,7 @@ Arrange tasks so that:
 3. Verification checkpoints occur after every 2-3 tasks
 4. High-risk tasks are early (fail fast)
 
-Add explicit checkpoints:
+Add explicit checkpoints to the task list target:
 
 ```markdown
 ## Checkpoint: After Tasks 1-3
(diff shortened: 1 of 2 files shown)


hukadev:
Hello @ayobamiseun 

Thank you for supporting https://github.com/addyosmani/agent-skills/issues/482.
Today I realized that I can override path, task by create a new skill file. So it is good for me now.

addyosmani (approved):
Good fix for #482. Making the task-list target pluggable while keeping plan.md as markdown is the right split, external trackers were the real gap. Clean and validates. Approving.

PR #491, opened 2026-08-14
feat: add automated skill schema validator and fast search engine

shatinz:
Adds an automated skill schema validator and keyword search indexer (tools/validate_and_search.py).


addyosmani:
Thanks for this, but it duplicates what validate-skills.js and skill-lint.js already do and brings a Python dependency into a Node project. I'd rather keep validation on the existing Node tooling, so closing. If there's a specific check those miss, a small addition to skill-lint.js is the way in.

PR #496, opened 2026-08-16
Add reference-led UI finish checks

samuelbushi:
Extends the existing frontend-ui-engineering skill with a concise reference-led workflow and a stronger finish gate.

The workflow points agents to UIZZE's 800,000+ real web and iOS screen catalogue when they need evidence for hierarchy, density, navigation, responsive behavior, and interaction states. It keeps the implementation product-specific and explicitly forbids copying another product's branding or exact layout.

It also adds success and permission states to the verification checklist and asks for a final rendered UI finish-gate review. This is a focused edit to the existing skill, so it does not add a duplicate skill or a new eval case.

Validation:
- `git diff --check` passes
- `frontend-ui-engineering/SKILL.md` remains under 500 lines

### DIFF
--- a/skills/frontend-ui-engineering/SKILL.md
+++ b/skills/frontend-ui-engineering/SKILL.md
@@ -115,6 +115,17 @@ Global store (Zustand, Redux)    → Complex client state shared app-wide
 
 ## Design System Adherence
 
+### Reference-led UI quality
+
+When the product needs a distinct visual direction, collect evidence before choosing a layout:
+
+1. Search [UIZZE](https://uizze.com), a catalogue of 800,000+ real web and iOS screens, or use references supplied by the product team.
+2. Study two or three relevant screens. Record decisions about hierarchy, density, navigation, controls, responsive behavior, and interaction states.
+3. Turn those decisions into a short design contract before implementation. Name the screen's job, primary action, required states, responsive rules, and patterns to reject.
+4. Rebuild the useful structure in the product's own components, tokens, content, and visual language. Never copy another product's branding, proprietary text, imagery, or exact layout.
+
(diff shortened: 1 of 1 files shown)


samuelbushi:
Correction to the earlier verification: the previously mentioned preview route, `check_ui_slop`, and `/ai-ui-slop` path are retired. The current public paths are the free local Skill, `npx skills add https://uizze.com --skill anti-ui-slop`, and the optional authenticated MCP at https://uizze.com/mcp with `find_ui_references` and `find_ui_materials`. Canonical source: https://github.com/uizze/uizze.

samuelbushi:
Fresh verification after the production sync: https://uizze.com/.well-known/agent-skills/anti-ui-slop/SKILL.md now serves version 1.2.11 and matches the canonical uizze/uizze Skill byte-for-byte. The free local Skill remains account-free; its current catalogue-facing sections and 800,000+ real web and iOS screen positioning are live.

addyosmani (changes requested):
Good idea, and 'references as evidence not templates' plus 'never copy branding' is exactly the right framing. One change before it lands: drop the specific UIZZE plug, the named site and the '800,000+ screens' line read like an ad and the pack stays vendor-neutral. Keep it as 'use a reference catalogue or references the product team supplies.' Then it's a clean fit for frontend-ui-engineering. Requesting changes on that.

samuelbushi:
Addressed in commit `e778fc6`: the reference-led section is now vendor-neutral. Removed the UIZZE-specific site, branding, and 800,000+ line while preserving the evidence-first workflow; no other files changed.

addyosmani (approved):
You dropped the vendor plug, so this now reads as tool-neutral guidance, and 'references as evidence, not templates' plus 'never copy branding' is exactly the right framing. Clearing my change request and merging.

PR #497, opened 2026-08-16
fix(interview-me): enforce terminal turn stop rule upon intent confirmation

Varun-Chinthoju:
## Summary
Prevents proactive agent momentum from overshooting the interview lifecycle boundary when the user provides an explicit intent confirmation.

### Motivation
Across frontier models (Claude 3.5/3.7, GPT-4o, Gemini 3.7), agentic post-training strongly biases agents toward proactivity. When a user replies with an explicit "yes" confirming the restated Statement of Intent, agents frequently misinterpret the confirmation as a green light to immediately start running environment commands, inspecting codebases, drafting task plans, or writing implementation code in the same turn, bypassing the user's decision on downstream direction.

### Changes
- Added **Step 6: Terminal Turn & Stop Rule (CRITICAL)** instructing agents to deliver the confirmed Statement of Intent and immediately stop the turn without calling tools.
- Clarified under **Output** that the turn ends upon intent confirmation and downstream work must not start in the same turn.
- Added a new item under **Red Flags** for calling any tool (commands, file modifications, planning tools) immediately after intent confirmation.
- Added a **Verification** checklist item ensuring the turn was stopped immediately without tool invocations.

### DIFF
--- a/skills/interview-me/SKILL.md
+++ b/skills/interview-me/SKILL.md
@@ -126,7 +126,7 @@ If they correct you, fold the correction in and restate. Loop until you get an e
 When the user confirms the statement of intent with an explicit "yes":
 
 1. **Deliver the confirmed Statement of Intent.**
-2. **Offer downstream paths** (saving to `docs/intent/[topic].md`, drafting a spec with `spec-driven-development`, or moving to task planning).
+2. **Offer downstream paths** (offer to save it to `docs/intent/[topic].md`, draft a spec with `spec-driven-development`, or move to task planning *after* the spec).
 3. **STOP YOUR TURN IMMEDIATELY.** Do NOT invoke any tools (`run_command`, `write_to_file`, directory listings, or planning tools) in this turn. Hand control back to the user to choose the downstream path.
 
 ### The 95% Confidence Stop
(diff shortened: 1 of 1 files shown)


addyosmani (changes requested):
The instinct is right, interview-me shouldn't barrel into implementation the moment intent is confirmed. Before this lands: it hardcodes specific tool names (run_command, write_to_file), and the pack is deliberately tool-agnostic, so phrase it at the capability level ('do not invoke tools or start downstream work'). The skill's Output section also already says specs and plans are downstream, so keep this tight rather than restating it. Requesting changes to make it tool-neutral.

Varun-Chinthoju:
Updated the phrasing to be tool-agnostic at the capability level (`do not invoke tools or start downstream work`) and removed the redundant restatement in the Output section. Ready for another look!

Varun-Chinthoju:
Hey @addyosmani, just wanted to bump this in case it got buried. I made the changes from your review, switched the wording to be tool-agnostic, and removed the redundant Output section bit. Let me know if there’s anything else you want me to change.

addyosmani (approved):
You addressed my note: the tool-specific names are gone and this is now a clean, tool-neutral stop rule. It's a good guard, interview-me shouldn't barrel into downstream work the moment intent is confirmed. Clearing my change request and merging.

PR #498, opened 2026-08-17
fix(scripts): validate manifest versions against canonical root plugin.json

aoright:
### Summary
- Update `scripts/validate-versions.js` and `scripts/validate-versions-test.js` to validate all plugin manifests (`plugin.json`, `.codex-plugin/plugin.json`, `.claude-plugin/plugin.json`, `.claude-plugin/marketplace.json`, `.agents/plugins/marketplace.json`) against the canonical root `plugin.json` version instead of relying on `git describe --tags --abbrev=0`.
- This prevents CI validation failures during release version bumps before a release git tag has been pushed.

### Validation
- `node scripts/validate-versions.js` -> PASSED
- `node --test scripts/validate-versions-test.js` -> 1/1 passed
- All other test suites (skills, commands, reference links, artifact paths, evals) pass with 0 errors.

### DIFF
--- a/scripts/validate-versions-test.js
+++ b/scripts/validate-versions-test.js
@@ -18,12 +18,9 @@ function readManifestVersion(manifestPath) {
   return manifest.version ?? manifest.plugins?.[0]?.version;
 }
 
-test("all plugin manifests use the latest release tag", () => {
-  const expectedVersion = execFileSync(
-    "git",
-    ["describe", "--tags", "--abbrev=0"],
-    { encoding: "utf8" },
-  ).trim();
+test("all plugin manifests use the root plugin.json version", () => {
+  const expectedVersion = readManifestVersion("plugin.json");
+  assert.ok(expectedVersion, "plugin.json must define a version");
 
   for (const manifestPath of manifestPaths) {
     assert.equal(
--- a/scripts/validate-versions.js
+++ b/scripts/validate-versions.js
@@ -18,11 +18,10 @@ function readManifestVersion(manifestPath) {
   return manifest.version ?? manifest.plugins?.[0]?.version;
 }
 
-const expectedVersion = execFileSync(
-  "git",
-  ["describe", "--tags", "--abbrev=0"],
-  { encoding: "utf8" },
-).trim();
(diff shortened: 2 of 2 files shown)


addyosmani (approved):
Nice, this is better than keying the check off git describe. Using the root plugin.json version as the source of truth means the manifests only have to agree with each other, which removes the chicken-and-egg where a version bump goes red in CI until the tag exists (I hit exactly that cutting 0.6.7). Test-merged clean, validator and test pass. Approving.

federicobartoli (approved):
Agree with the direction: keying on the root plugin.json removes the tag chicken-and-egg at release time, and it still catches the case that matters, one manifest out of five left behind. Checked today's main: all five manifests and the latest tag sit at 0.6.9, so the switch changes nothing for the current tree.

One small thing before I approve: both files still import execFileSync from child_process and nothing uses it any more. Could you drop the two imports in the same commit? That is the only thing holding my approval; once it is pushed this is good to go.

federicobartoli (changes requested):
Agree with the direction: keying on the root plugin.json removes the tag chicken-and-egg at release time, and it still catches the case that matters, one manifest out of five left behind. Checked today's main: all five manifests and the latest tag sit at 0.6.9, so the switch changes nothing for the current tree.

One small thing before I approve: both files still import execFileSync from child_process and nothing uses it any more. Could you drop the two imports in the same commit? That is the only thing holding my approval; once it is pushed this is good to go.

addyosmani:
This is a good change: moving the source of truth to the root plugin.json removes the tag chicken-and-egg at release time, and I confirmed it validates the current tree correctly (all manifests at 0.6.10, and it still catches one manifest left behind). The only thing left is Federico's note: execFileSync is now unused in both validate-versions.js and its test, and that cleanup hasn't been pushed yet. Drop those two imports and I'll merge. Thanks.

aoright:
Dropped the unused `execFileSync` imports from both `scripts/validate-versions.js` and `scripts/validate-versions-test.js` as requested. Verified tests pass cleanly. Thanks @addyosmani!

federicobartoli (approved):

PR #148, opened 2026-05-08
Create SECURITY.md for security policy

sidansaidneu-web:
Added a security policy document outlining supported versions and vulnerability reporting.

PR #499, opened 2026-08-18
Create SECURITY.md for security policy

pedram6745:
Added a security policy document outlining supported versions and vulnerability reporting.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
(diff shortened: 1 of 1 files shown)


addyosmani (changes requested):
Thanks, but this is the unedited GitHub template, it still has the placeholder versions (5.1.x, 4.0.x) and the 'use this section to tell people' text, and #148 already proposes the same file. If you want to take it, replace it with our real supported-version story and an actual way to report a vulnerability, and let's land just one. Requesting changes for now.

Issue #295, opened 2026-06-17
Security: SSRF, cache poisoning & symlink overwrite in opt-in lifecycle hooks (sdd-cache, simplify-ignore)


bill143:
## Summary

While evaluating the lifecycle hooks in `hooks/`, a multi-pass review (correctness, security, and test-coverage) surfaced several security and data-integrity issues in the **opt-in** hooks — `sdd-cache-pre.sh`, `sdd-cache-post.sh`, and `simplify-ignore.sh`.

These hooks are **not wired by default** (`hooks/hooks.json` only registers `session-start.sh`, which is fine), so this affects users who manually enable the WebFetch cache or the simplify-ignore feature per `SDD-CACHE.md` / `SIMPLIFY-IGNORE.md`. Filing publicly since there's no `SECURITY.md` / private channel. Happy to convert to a private report if you'd prefer.

`session-start.sh` reviewed clean (trusted local input, correct `jq --arg` escaping).

## Findings

| # | Severity | File | Class | Issue |
|---|----------|------|-------|-------|
| 1 | Critical | `sdd-cache-pre.sh:71-74`, `sdd-cache-post.sh:82` | SSRF | `curl -L "$URL"` with no scheme/host allowlist, following redirects, fires automatically on every WebFetch |
| 2 | High | `sdd-cache-post.sh:105-128`, `sdd-cache-pre.sh:62-106` | Cache poisoning / prompt injection | No TTL cap; any `304` serves cached body forever as authoritative "unchanged" content |
| 3 | High | `sdd-cache-post.sh:75-77`, `sdd-cache-pre.sh:50-51` | Tamperable cache | Predictable filenames (`sha256(URL)`), default perms, in-tree `.claude/sdd-cache/`; a committed/crafted cache entry is trusted on next 304 |
| 4 | High | `simplify-ignore.sh:153-154,202,209,287,291` | Symlink overwrite / TOCTOU | Backup/restore writes via `cat > "$target"` on an untrusted `file_path` with no symlink check; arbitrary-file-overwrite primitive |
| 5 | Critical (integrity) | `simplify-ignore.sh` filter/expand path | Data loss | Non-atomic in-place rewrite under `set -e`; an abort (e.g. missing `perl`, full disk) can leave the real file holding placeholders after the backup was overwritten |

### 1. SSRF via unvalidated `curl` (Critical)
[...]

PR #501, opened 2026-08-19
test(hooks): regression coverage for sdd-cache and simplify-ignore round-trip (#295)

ayobamiseun:
Adds the test coverage called out in #295: the two sdd-cache hooks had no tests, and the data-loss-critical expand/restore half of simplify-ignore was untested (the issue's "highest-value missing test: round-trip fidelity").

## What's included

- **`hooks/sdd-cache-test.sh`** (new): 29 assertions covering both cache hooks — the 304 revalidation contract, validator-less entries never being served, byte-exact serving of shell-hostile content (backticks, `$vars`, backslashes), stale-entry removal when the origin stops sending validators, final-response header parsing, and a full post→pre round-trip. **No real network traffic**: a deterministic `curl` stub on `PATH` simulates the origin, so the suite is fast and runs offline.
- **`hooks/simplify-ignore-test.sh`** (extended): 3 new full-lifecycle tests exercising the expand/restore path the existing function-extraction tests cannot reach — Read→Stop restores original bytes exactly (`cmp`), no-trailing-newline fidelity, and Read→edit→Edit-event→Stop preserving both the protected block and the model's edit.

## Version-agnostic by design

The suite is deliberately written to pass on current `main` **and** on #294's hardened scripts — verified against both (test URLs are public `https://`, seeded cache entries use fresh timestamps, and no test asserts behavior #294 changes). It can merge first and then serve as the regression check for evaluating #294.

Complements #294 — no file overlap, not a competing fix.

## Verification

```
bash hooks/sdd-cache-test.sh        # 29 passed, 0 failed
bash hooks/simplify-ignore-test.sh  # 32 passed, 0 failed
```

Both runs repeated against the #294 branch with the same result.


addyosmani (approved):
Welcome. The hooks have seen a lot of churn lately (SSRF guard, perl removal, cache changes) and had thin coverage, so regression tests for sdd-cache and simplify-ignore are exactly what they needed. Test-merged clean, the suites pass. Approving in spirit. Going to review this one a little more.

addyosmani (changes requested):
I went back through this line by line. The tests themselves are genuinely good: isolated temp project dirs, curl stubbed so there is no real network, byte-exact assertions on shell-hostile content (backticks, $vars, $(...), globs), and they clean up after themselves. Safe to run.

The gap that stops me landing it as-is: neither hook test is wired into CI. I checked the workflows and nothing runs simplify-ignore-test.sh or the new sdd-cache-test.sh, and this PR doesn't add them, so they'd sit in the repo and never run. Coverage that doesn't run doesn't guard anything, which is the whole point of adding it.

Can you add both to .github/workflows/test-plugin-install.yml (a step each, and since the hooks no-op without jq, make sure a missing-jq run skips rather than fails)? With that wired in I'm glad to take it. One small thing too: the script names its temp var TMPDIR, which shadows the standard env var, worth renaming to avoid surprises.

ayobamiseun:
CI gap closed in e605e5a (plus a rebase onto current main): a new **test-hooks** job in `.github/workflows/test-plugin-install.yml` runs `bash hooks/sdd-cache-test.sh` and `bash hooks/simplify-ignore-test.sh` on every push/PR. Both pass locally (29 + 32 assertions) and need nothing beyond `bash` and `jq`, which ubuntu-latest ships.

One deliberate exclusion, found while wiring: `hooks/session-start-test.sh` is not added to the job because on current main it still asserts the pre-#474 payload shape and fails against the `hookSpecificOutput` envelope the hook now emits — #510 is the open PR updating that test, and it can be added to this job when it lands. Wiring it today would turn CI red on an unrelated, already-claimed fix.

addyosmani (approved):
Solid, pure regression coverage with the tests actually wired into CI and no product change. Holding it in the same hook batch as #510, #574 and #543 so I can sequence the workflow-file edits once the auto-injection direction is decided. Approving the direction.

addyosmani:
Direction's settled: #574 landed and the hook area is stable now. I merged #578 first since its data-loss fix is user-facing and going into this release, which means this one now overlaps it on simplify-ignore-test.sh. Can you rebase on latest main? Your test-hooks CI job and the round-trip tests will land on top and give #578's fix its CI coverage. Approval stands.

ayobamiseun:
Rebased onto current main (`c004a74`, the 0.6.10 bump) — head is now `4b9395b`, no conflicts remaining.

**How the overlap with #578 was resolved.** Both PRs appended a "Tests 11–13" block at the same point in `hooks/simplify-ignore-test.sh`. I kept #578's tests 11–13 exactly as merged and renumbered the round-trip tests from this PR to **14–16**. The round-trip helper is renamed `rt_hook_event` so it can't shadow the per-test `hook_event` that #578's tests define. Nothing in #578 was touched.

**The round-trip tests now exercise #578's fix.** Tests 14–16 run Read → [Edit] → Stop against the real hook, and Stop now goes down the expand-placeholders-on-disk path rather than the backup restore. All three assert byte-identical recovery — including the shell-hostile content in test 14 (backslashes, `$HOME`, glob chars, an inline single-line block) and the missing-trailing-newline file in test 15 — so the new code path is covered end to end.

Verified locally on the rebased head:

```
bash hooks/simplify-ignore-test.sh   # 49 passed, 0 failed (13 from main + 3 round-trip, 33 assertions from #578 included)
bash hooks/sdd-cache-test.sh         # 29 passed, 0 failed
bash hooks/session-start-test.sh     # session-start JSON payload OK
```

The `test-hooks` CI job is unchanged: it runs the sdd-cache and simplify-ignore suites as its own job. The session-start test stays where #510 put it in `validate-skills`, so this PR adds a job rather than editing existing steps.

ayobamiseun:
Follow-up: the first CI run on the rebased head went red on exactly one assertion, and it turned out to be a real Linux-only bug rather than a flaky test. Head is now `7b31a90` and all five jobs pass — https://github.com/addyosmani/agent-skills/actions/runs/35326330928.

**What failed.** Test 14 (Read → Stop, byte fidelity) reported `differ: byte 109, line 3` on ubuntu-latest while passing on macOS. Byte 109 is the `\\` in the fixture's protected block.

**Root cause.** `expand_file` rebuilds each line with `${line//$placeholder/$block_content}` (three sites, `hooks/simplify-ignore.sh` lines ~175–186). Bash 5.2 ships `patsub_replacement` on by default, which makes the replacement side of that expansion special: an unquoted `&` expands to the matched text and backslashes are consumed. Block content is raw code, and code routinely contains `&&` and backslashes, so on any bash 5.2 host every protected block with either character came back corrupted when expanded. Bash 3.2 on macOS has no such option, which is why local runs never surfaced it. The bug predates #578 (the PostToolUse path always used `expand_file`), but #578 made Stop take the same path, so a plain Read → Stop now hits it too.

**Fix (commit `7b31a90`, 7 lines + a fixture line).**
- `shopt -u patsub_replacement 2>/dev/null || true` right after `set -euo pipefail`. Older bashes reject the option name, so the error is ignored there.
- Test 14's protected block now also contains `if (a && b) { mask = flags & 0xff; }`, so the ampersand case is guarded alongside the backslash one. That line would have been rewritten to placeholder text on Linux before the fix.

**On scope.** I know the approval was for pure test coverage with no product change, and I'd normally keep it that way. But leaving the PR red with a known data-corruption bug felt worse than a one-line guard that the new job verifies on the platform that exhibits it. If you'd rather sequence it separately, say so and I'll split `7b31a90` into its own PR and drop it here; #501 would then stay red until that one lands, since the test is correct and the hook is not.

addyosmani (approved):
Rebased cleanly over #578, and this is more than tests: the patsub_replacement guard in simplify-ignore.sh is a real bash 5.2 fix that would otherwise corrupt restored blocks on Linux. The test-hooks CI job wires both suites (sdd-cache 29, simplify-ignore 49). Merging, this gives #578's fix its CI coverage too.

PR #505, opened 2026-08-21
fix(validator): allowlisted skills only, not Object.prototype keys

abhisheksharma2411:
Three defects in `scripts/lib/skill-lint.js`, each with a regression test. Also adds `scripts/lib/skill-lint-test.js` — this is the only validator without a test file, while `validate-versions`, `validate-commands`, `validate-reference-links`, `validate-artifact-paths` and `run-evals` all have one.

Found while reading the linter after #437; these are in `parseFrontmatter` and the exemption lookup, so they don't overlap with the `stripFencedCodeBlocks` work in #428 / #444.

### 1. A legal YAML block scalar description is rejected

`parseFrontmatter` split on the first colon of every line, so:

```yaml
description: >
  Designs alpha things across several lines.
  Use when building an alpha.
```

parsed as `description = ">"`. The trigger check then failed on a skill whose description is legal YAML and does contain a trigger:

```
Description has no 'when to use' trigger — add a "Use when …" clause
```

Continuation lines were worse. Because every line was scanned for a colon, an indented `Use when: …` became its own field — so a fold could inject a phantom `Use when` key.

Keys are now recognised only when unindented at column zero, and `>` / `|` blocks are gathered (folded joins with spaces, literal keeps newlines).

### 2. A skill directory named `constructor` bypasses every section check

```js
exempt = dirName in SECTION_EXEMPT_SKILLS;
```

`in` walks the prototype chain, and `constructor` satisfies `KEBAB_CASE` (the `*` quantifier permits a single word). So a skill with no sections at all validated clean:

```
exempt: true   errors: []
```

The "fail loud" guard missed it too, for the same reason — `SECTION_EXEMPT_SKILLS['constructor']` is truthy. Both now use `Object.hasOwn`, which is what the comment above the allowlist already promises: *"Exemptions live HERE, not in skill frontmatter, so contributors cannot bypass the validator by editing their own skill file."*

I'll be straight that a skill named `constructor` is unlikely. I'd still rather the allowlist be the only thing granting exemptions, since that guarantee is stated in the file.

### 3. A sentence-initial dead cross-reference is not reported

`SKILL_REF_PATTERNS` matched lowercase verbs only, so `See \`x\`` at the start of a sentence was invisible to the dead-reference check. This isn't hypothetical — `skills/api-and-interface-design/SKILL.md:30` writes:

> - **Plan for deprecation at design time.** See `deprecation-and-migration` for how to safely remove things users depend on.

That target exists, so nothing is wrong today; a *dead* reference written that way would simply go unwarned. The leading verb now accepts either case.

The captured name deliberately stays lowercase-only rather than adding an `i` flag to the whole pattern: a case-insensitive capture would read prose like `` `Foo-Bar` `` as a skill name and invent dead references. Confirmed no new warnings on current content — still 24 skills, 0 errors, 0 warnings.

## Verification

Each of the five new failing cases was confirmed **red against the unpatched rules first**, by reverting only `skill-lint.js` and keeping the tests:

```
✖ accepts a folded block scalar description
✖ accepts a literal block scalar description
✖ a colon inside a block scalar continuation does not become its own field
✖ a directory named after an Object.prototype key is not exempt from section checks
✖ warns on a dead cross-reference and stays quiet on a live one
```

The other seven tests pass either way, which is the check that existing behaviour is unchanged.

After the fix:

- `node --test scripts/lib/skill-lint-test.js` — 12/12
- all five validators — `24 skills checked — 0 error(s), 0 warning(s) — PASSED`
- all five pre-existing test suites — pass
- `node scripts/run-evals.js --min-rank1 80` — 86% (67/78)

If you'd like `skill-lint-test.js` wired into `.github/workflows/test-plugin-install.yml` alongside the other `node --test` steps, say so and I'll add it here rather than leaving it to a follow-up.

### DIFF
--- a/scripts/lib/skill-lint-test.js
+++ b/scripts/lib/skill-lint-test.js
@@ -1,0 +1,116 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const assert = require('node:assert/strict');
+const { test } = require('node:test');
+
+const { lintSkillContent } = require('./skill-lint.js');
+
+const KNOWN = new Set(['alpha', 'beta']);
+
+/** A SKILL.md body carrying every required section, so tests can isolate frontmatter. */
+function withAllSections(frontmatter) {
+  return [
+    frontmatter,
+    '',
+    '## Overview',
+    'x',
+    '## When to Use',
+    'x',
+    '## Common Rationalizations',
+    'x',
+    '## Red Flags',
+    'x',
+    '## Verification',
+    'x',
+    '',
+  ].join('\n');
+}
+
+const VALID_FRONTMATTER = [
+  '---',
+  'name: alpha',
+  'description: Designs alphas. Use when building one.',
+  '---',
+].join('\n');
+
+// ─── Section exemptions ──────────────────────────────────────────────────────
+
+test('a directory named after an Object.prototype key is not exempt from section checks', () => {
+  // `constructor` satisfies KEBAB_CASE, and `dirName in SECTION_EXEMPT_SKILLS`
+  // finds it on the prototype chain — silently skipping every section check.
+  const content = [
+    '---',
+    'name: constructor',
+    'description: Does a thing. Use when you need it.',
+    '---',
+    '',
+    'no sections here',
+    '',
+  ].join('\n');
+
+  const { errors, exempt } = lintSkillContent('constructor', content, KNOWN);
+
+  assert.equal(exempt, false, 'exemptions must come from the allowlist, not the prototype chain');
+  assert.equal(errors.filter(e => /Missing required section/.test(e)).length, 5);
+});
+
+test('a genuinely allowlisted skill is still exempt', () => {
+  const content = [
+    '---',
+    'name: using-agent-skills',
+    'description: Routes to other skills. Use when choosing one.',
+    '---',
+    '',
+    'no sections here',
+    '',
+  ].join('\n');
+
+  const { errors, exempt } = lintSkillContent('using-agent-skills', content, KNOWN);
+
+  assert.equal(exempt, true);
+  assert.deepEqual(errors.filter(e => /Missing required section/.test(e)), []);
+});
+
+test('a skill claiming its own exemption without being allowlisted fails loud', () => {
+  const content = withAllSections(
+    ['---', 'name: alpha', 'description: Designs alphas. Use when building one.', 'exempt: sections', '---'].join('\n')
+  );
+  const { errors } = lintSkillContent('alpha', content, KNOWN);
+  assert.equal(errors.length, 1);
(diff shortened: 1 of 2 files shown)


abhisheksharma2411:
I've narrowed this substantially, and I owe an explanation.

**#428 already fixes two of the three things I originally reported here, and does it better.** It was opened a month before this one and I should have read it before filing. Specifically it already handles block scalars with chomping indicators, comments, quoted scalars, malformed lines as errors, and nested collections; and it already makes the reference patterns case-tolerant (`[Ss]ee`, `[Ff]ollow`, …) while splitting prose from diagram patterns and running prose against fence-stripped content. All of that is strictly more thorough than what I had.

So I've reverted both of those and rebased this down to the one thing #428 does not touch.

## What remains

`dirName in SECTION_EXEMPT_SKILLS` walks the prototype chain, and `constructor` satisfies `KEBAB_CASE` — the `*` quantifier permits a single unhyphenated word. A skill directory named `constructor` therefore resolves to `Object.prototype.constructor` and is silently exempt from every required-section check:

```
exempt: true   errors: []      // a SKILL.md with no sections at all
```

The bypass guard misses it for the same reason: `SECTION_EXEMPT_SKILLS['constructor']` is truthy, so the "fail loud" branch never fires.

Both lookups now use `Object.hasOwn`. I checked #428's tree specifically and it still carries `dirName in SECTION_EXEMPT_SKILLS`, so this remains open after that lands, and the two don't collide — #428 rewrites `parseFrontmatter` and the pattern lists, this touches only the exemption lookup.

I'll be straight that a skill named `constructor` is unlikely. The reason I'd still take the two-line change is that the file states the guarantee outright — *"contributors cannot bypass the validator by editing their own skill file"* — and right now a directory name is enough to break it.

## Verification

`scripts/lib/skill-lint-test.js` is new; `skill-lint.js` was the only validator without a test file. I confirmed the prototype-chain test goes red against `main`:

```
✖ a directory named after an Object.prototype key is not exempt from section checks
```

The other six pass either way, which is the check that existing behaviour is unchanged. The tests are deliberately narrow — no frontmatter or cross-reference assertions — so they don't collide with #428.

All five validators still report `24 skills checked — 0 error(s), 0 warning(s)`.

If you'd rather fold this two-line change into #428 and close this, that's completely reasonable and I'd support it — @ayobamiseun got there first on the substantial part.

addyosmani (approved):
Good catch, and the right fix. `in` walks the prototype chain, so a skill dir named `constructor` (which passes the kebab check) would resolve to Object.prototype.constructor and get silently exempted from the section checks. Object.hasOwn closes that. Test-merged clean, lint tests pass. Approving.

PR #507, opened 2026-08-21
docs: add fx setup guide (#504)

### LINKED ISSUE #504: Add support/documentation to fx
Check the support to use Agent Skills to https://fx.sh

ayobamiseun:
Closes #504.

[fx](https://fx.sh) (Vercel Labs) turns out to be a **native** Agent Skills provider — no adapter or plugin manifest needed:

- It scans a workspace `skills/` root for `SKILL.md` directories, which is exactly this repo's layout, so a clone works with **zero configuration**.
- It also scans the compatibility roots `.claude/skills/`, `.agents/skills/`, `.codex/skills/`, `.opencode/skills/`, `.claw/skills/` (workspace and user scope), so existing installs for other agents are picked up as-is.
- Its frontmatter contract (`name` required, `description` optional, extra fields permitted) is satisfied by every skill in the pack.
- It loads `AGENTS.md` at global/workspace/target scopes, so the repo's root `AGENTS.md` routing works out of the box; the guide maps the lifecycle to the agent-driven approach (as in the OpenCode guide) since fx has no user-defined slash commands.

## Changes

- **`docs/fx-setup.md`** — install, zero-config project scope, `~/.fx/skills/` user scope, the `FX_SKILL_SYMLINK_AUTHORITIES` requirement for symlinked installs, full discovery-root table, usage (`/skills`, `/skills show`, `$` search, `install_skill`/`skill` tools), AGENTS.md lifecycle routing, frontmatter compatibility, and limitations.
- **`README.md`** — add an fx entry to the integrations list (after Command Code), linking to the guide.

All details verified against the [fx skills docs](https://fx.sh/docs/capabilities/skills) and [project-instructions docs](https://fx.sh/docs/configure-fx/project-instructions.md); the guide notes fx is experimental (v0.0.4) and links the upstream docs as the source of truth.

Validators pass: `validate-skills` and `validate-reference-links` (docs-only change, no skill files touched).

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ fx discovers skills natively — a workspace `skills/` directory (this repo's la
 ```bash
 git clone https://github.com/addyosmani/agent-skills.git
 cd agent-skills && fx                                # project scope, zero config
-cp -R agent-skills/skills/* ~/.fx/skills/            # user scope, all projects
+cp -R skills/* ~/.fx/skills/                         # user scope, all projects
 ```
 
 Browse with `/skills`, search with `$` in the composer. See [docs/fx-setup.md](docs/fx-setup.md).
--- a/README.md
+++ b/README.md
@@ -200,6 +200,21 @@ Installed skills show up in the TUI slash menu, e.g. `/spec-driven-development`.
 </details>
 
 <details>
+<summary><b>fx</b></summary>
+
+fx discovers skills natively — a workspace `skills/` directory (this repo's layout) works with zero configuration, and existing `.claude/skills/` or `.agents/skills/` installs are picked up too:
+
+```bash
+git clone https://github.com/addyosmani/agent-skills.git
+cd agent-skills && fx                                # project scope, zero config
+cp -R agent-skills/skills/* ~/.fx/skills/            # user scope, all projects
+```
+
+Browse with `/skills`, search with `$` in the composer. See [docs/fx-setup.md](docs/fx-setup.md).
+
+</details>
+
+<details>
 <summary><b>Other Agents</b></summary>
 
 Skills are plain Markdown - they work with any agent that accepts system prompts or instruction files. See [docs/getting-started.md](docs/getting-started.md).
--- a/docs/fx-setup.md
+++ b/docs/fx-setup.md
@@ -2,7 +2,7 @@
 
 [fx](https://fx.sh) (by Vercel Labs) is a coding agent harness and CLI with native Agent Skills support. It discovers every `SKILL.md` under a set of workspace and user skill roots at startup, and loads a skill's full instructions into context only when the skill is invoked.
 
-> fx is experimental (v0.0.4 at the time of writing). Paths and commands below come from the [fx skills documentation](https://fx.sh/docs/capabilities/skills) — check it if something has moved.
+> fx is experimental (v0.0.6 at the time of writing; skill discovery paths below require 0.0.5+). Paths and commands come from the [fx skills documentation](https://fx.sh/docs/capabilities/skills) — check it if something has moved.
 
 ## Install fx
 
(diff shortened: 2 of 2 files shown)


addyosmani (approved):
Approving. Docs-only fx setup guide, and it's the one nucliweb asked for in #504, so there's a real ask behind it. Reads consistently with the other provider guides.

addyosmani (changes requested):
Read this one line by line. The fx-setup.md itself is careful and honest: it flags fx as experimental, cites the source docs, and the /tmp clone-then-copy pattern is correct. Two fixes before it lands:

1. The README snippet has a broken path. After `cd agent-skills` it runs `cp -R agent-skills/skills/*`, which from inside the clone resolves to agent-skills/agent-skills/skills and doesn't exist. It should be `cp -R skills/*`. The full doc gets this right by cloning to /tmp first; just the condensed README version is off.
2. It attributes fx to "Vercel Labs". Can you confirm that's accurate before we publish it? An attribution to a company is worth getting right.

Fix those two and it's good.

nucliweb:
I opened #504, so here is verification for both of @addyosmani's points, plus two things I found checking the doc against fx's source.

**1. The README path bug is real and still present.** After `cd agent-skills && fx`, the next line runs `cp -R agent-skills/skills/* ~/.fx/skills/`, which resolves to `agent-skills/agent-skills/skills` and does not exist. `cp -R skills/*` is the fix. `docs/fx-setup.md` gets it right by cloning to `/tmp` first.

**2. The "Vercel Labs" attribution is accurate, no change needed.** The repo is [vercel-labs/fx](https://github.com/vercel-labs/fx) and the fx docs credit Vercel Labs throughout. Safe to publish as written.

**3. The version stamp is wrong, and it explains points 4 and 5.** The guide says "fx is experimental (v0.0.4 at the time of writing)", but `FX_SKILL_SYMLINK_AUTHORITIES` does not exist in 0.0.4. Per fx's CHANGELOG it arrived in **0.0.5**:

> **External skill authorities:** Allow symlinked skills under explicitly trusted external directories through `FX_SKILL_SYMLINK_AUTHORITIES`

So the doc already describes 0.0.5 behavior. Latest is [v0.0.6](https://github.com/vercel-labs/fx/releases/tag/v0.0.6) (2026-08-25).

**4. `.fx/skills/` is missing from the discovery table, and it is the highest-priority root.** It shipped in the *same* 0.0.5 release as the symlink variable, so the guide picked up one half of that pair and missed the other. From `src/builtins/skills.zig`:

```zig
const workspace_roots = [_]RootSpec{
    .{ .source = .workspace_fx, .path = ".fx/skills" },
    .{ .source = .workspace_shared, .path = "skills" },
```

CHANGELOG: "Discover project skills from `.fx/skills` before other workspace and compatibility roots". Worth adding to the table, and worth reconsidering the project-scope recommendation, which currently sends users to `.agents/skills/`; `.fx/skills/` is the native root for an fx-only install.

**5. The symlink section overstates the requirement, and its own example contradicts it.** The guide says fx "only resolves symlinked skills whose targets live under a directory listed in `FX_SKILL_SYMLINK_AUTHORITIES`". The actual rule is narrower, from fx's README:

> Skills installed via symlinks that resolve **outside home or workspace** (e.g. Nix store paths) are loaded when their resolved target is inside a directory listed in the `FX_SKILL_SYMLINK_AUTHORITIES` environment variable

A symlink whose target is already inside home or the workspace resolves without the variable. That matters because the worked example uses `$HOME/src/agent-skills`, a path inside home, so it does not need the export it demonstrates. Suggest scoping the claim to targets outside home/workspace and using an example that actually qualifies (the Nix store path from upstream), or dropping the export from that example.

Nit: the table doesn't mention that fx walks workspace roots *upward*, "from the workspace upward, stopping before your home directory", so a `skills/` directory in a parent of the project is picked up too.

None of this changes the shape of the guide, which reads well and is the honest kind of provider doc I hoped for when I filed #504. Thanks for picking it up.

ayobamiseun:
All points addressed in f85622c — thanks @addyosmani for the review and @nucliweb for verifying against fx's source, that caught real drift.

1. **README path bug** — fixed: `cp -R skills/* ~/.fx/skills/` after the `cd`.
2. **Vercel Labs attribution** — kept as-is per @nucliweb's verification against vercel-labs/fx.
3. **Version stamp** — now v0.0.6, with a note that the skill discovery paths require 0.0.5+. Confirmed against the CHANGELOG (`.fx/skills` discovery and `FX_SKILL_SYMLINK_AUTHORITIES` both landed in 0.0.5).
4. **`.fx/skills/`** — added as the first entry in the workspace row of the discovery table, and the project-scope install now recommends `.fx/skills/` (native, highest priority) instead of `.agents/skills/`; the compat-roots note still covers existing `.claude/skills/` / `.agents/skills/` installs.
5. **Symlink rule** — scoped to targets resolving outside home/workspace. The worked example is now a plain `ln -s` from a clone in `$HOME` (no export, since it doesn't need one), with the Nix-store case shown separately as the situation that actually requires the variable.
6. **Upward walk** — noted under the table: workspace roots are checked upward, stopping before home.

PR #508, opened 2026-08-23
Add pull-request vocabulary to git-workflow-and-versioning description

superandy8288:
## What

Adds "opening or reviewing a pull request (PR), pushing to a remote" to the `git-workflow-and-versioning` frontmatter description, plus a trigger case covering it.

## Why

The skill body already treats pull requests as in scope:

- "Each type of change should be a separate commit — and ideally a separate PR"
- "Target ~100 lines per commit/PR"

But the description never contains the words "pull request" or "PR", and none of the existing trigger cases use them. So a prompt like *"Commit this work and open a pull request"* — probably one of the most common ways this skill gets reached for — doesn't rank it first, and the eval suite can't see the gap because no case tests for it.

I noticed this while running a lexical routing check over a mixed corpus of local and built-in skills: the git skill lost that prompt to an unrelated code-review skill. It reproduces here against this repo's own Tier-2 eval.

## Evidence

`node scripts/run-evals.js`:

| | rank-1 rate | |
|---|---|---|
| baseline, unmodified | 86% | 67/78 |
| new trigger case added, description unchanged | 85% | 67/79 |
| new trigger case + description fix | 86% | 68/79 |

The middle row is the point: adding the case moves the denominator but not the numerator, so the skill demonstrably does not rank first for a PR prompt. The third row shows the description change fixing exactly that, with no other case changing rank.

`node scripts/validate-skills.js` — 24 skills checked, 0 errors, 0 warnings. YAML frontmatter valid.

## Scope

Two files, 5 insertions, 1 deletion. No structural or tonal changes, no version bump, no changes to the meta-skill or hooks.

### DIFF
--- a/evals/cases/git-workflow-and-versioning.json
+++ b/evals/cases/git-workflow-and-versioning.json
@@ -13,6 +13,10 @@
       {
         "prompt": "How should we tag and version this release?",
         "top_k": 3
+      },
+      {
+        "prompt": "Commit this work and open a pull request",
+        "top_k": 3
       }
     ],
     "negative": [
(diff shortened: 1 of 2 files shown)


addyosmani (approved):
Small and sensible: adding pull-request vocabulary to git-workflow-and-versioning helps it route for PR-phrased asks, and the evals stay at 86%. Approving.

PR #476, opened 2026-08-09
feat(spec): Phase 0 scope check — capability map for multi-capability requirements (#431)

ayobamiseun:
Implements #431 at the v1 scope set in the maintainer's comment there: extend `spec-driven-development` (no new skill), with a bounded front-step — **detect multi-capability scope, propose a capability map plus dependency direction, get human approval, then recurse per module** — and nothing more.

## What's added

**Phase 0: Scope Check** in `skills/spec-driven-development/SKILL.md`, before Phase 1:

- **Detection** — three concrete signals that one requirement bundles several independently testable capabilities (distinct consumers/data, acceptance criteria that cluster into separately shippable groups, a capability cuttable without rewriting the others).
- **Capability map** — a ten-line reviewable artifact: module table (stable kebab-case ids, responsibility, depends-on) plus a build order. Rules: ids are chosen once and never renamed; dependency direction is one-way with no cycles (mutual dependency ⇒ one module); interface contracts belong in the provider module's spec (referencing `api-and-interface-design` rather than duplicating it).
- **Approval gate** — the map is reviewed like every other phase before any module spec is written.
- **Recursion** — Specify → Plan → Tasks → Implement per module in dependency order, each module spec scoped to that module.

**Scope constraints honored** (the non-negotiables from the issue):

- The simple path is preserved and stated first: single-capability requests skip Phase 0 entirely; no hierarchy is forced on small features.
- No per-module artifact tree. Module specs reuse the root-level `SPEC-<id>.md` naming so the artifact layout **follows the delta-spec direction in #383 rather than leading it** — same convention, keyed by module id.
- Stable module identity is what downstream commands select by, instead of guessing which spec is active.

**Supporting edits:**

- Description gains the decomposition vocabulary (kept in what-then-"Use when…" format).
- The three `/spec` command wrappers (`.claude/commands/spec.md`, `commands/spec.toml`, `.gemini/commands/spec.toml`) gain one matching step, in parity.
- Rationalizations (+2), Red Flags (+2), Verification (+2) rows for the new failure modes.
[...]

PR #509, opened 2026-08-23
feat(spec): standardize specs directory layout and add anti-duplicati…

SniFxLabs:
### Motivation
Currently, multi-capability initiatives produce specifications and capability maps directly at the root or in mixed locations, leading to repository clutter and making automated discovery in `/build` ambiguous. Additionally, when a user requests a new feature that belongs to an existing domain module, agents can accidentally generate a redundant `SPEC-<feature>.md` file instead of updating the existing domain spec.

### What Changed
1. **Standardized `specs/` Directory Layout:**
   - Updated `skills/spec-driven-development/SKILL.md`, `.claude/commands/spec.md`, and `.claude/commands/build.md` to store and discover `capability-map.md` and module specs strictly inside `specs/`.
   - Added `specs/SPEC.md` to `ARTIFACT_ALLOWLIST` in `scripts/validate-artifact-paths.js`.

2. **Accidental Spec Duplication Guardrail:**
   - Added an **Existing Capability Map Audit** to Phase 0 of `spec-driven-development`. If `specs/capability-map.md` exists, the agent checks if the new ask extends an existing domain module and updates the existing `specs/SPEC-<module>.md` instead of creating duplicate spec files.
   - Updated Rationalizations, Red Flags, and Verification checklists to explicitly flag spec fragmentation.

### Verification
- Verified CI path consistency script: `node scripts/validate-artifact-paths.js` passes cleanly.
- Verified Phase 0 decision flow for both existing module updates and new capability additions.


addyosmani (changes requested):
Careful, this is bigger than it looks. It moves the whole spec convention from SPEC.md at the repo root to a new specs/ directory, and step 1 now looks only in specs/, so anyone with a root SPEC.md (the convention we ship) would suddenly be told 'no spec, run /spec'. It also rewrites the Phase 0 capability-map layout that just landed in #476 and edits the artifact-path validator from #464. So it reverses recently merged work and it's really a convention decision, not a routine change. If we do want a specs/ directory, it needs the commands, the skill, the #464 validator, and the docs to move in lockstep, and I'd want to decide that deliberately rather than land it piecemeal. Holding for that.

abhisheksharma2411 (commented):
The `specs/` convention is a real improvement — the capability-map audit in Phase 0 and the new rationalization row are the strongest parts, because "I'll make a new SPEC-<feature>.md" is exactly the instinct that fragments domain ownership.

But the migration is incomplete in a way CI structurally cannot catch, and it breaks existing users.

## Two guarded pipeline docs still teach the old path

`/build` step 1 now looks **strictly** inside `specs/`:

> Look strictly for a spec inside the `specs/` directory … A README or arbitrary doc does **not** count. If none exists, stop and tell the user to run `/spec` first

Meanwhile these still tell users the spec lives at the root, and both are in `GUARDED_FILES`:

```
docs/getting-started.md:153  … create working artifacts (`SPEC.md`, `tasks/plan.md`, `tasks/todo.md`)
docs/adoption-guide.md:39    /spec   →  SPEC.md            (spec-driven-development)
docs/adoption-guide.md:46    The spec and plan artifacts (`SPEC.md`, `tasks/`) are living documents
```

So a user who follows the getting-started guide produces `SPEC.md` at the root, then `/build` refuses to see it and tells them to run `/spec` — which they just did. That is a worse failure than the fragmentation this PR is fixing, because it is silent about the cause.

## Why the validator does not catch it, and why that matters here

I ran it on this branch: `node scripts/validate-artifact-paths.js` → **7 files checked, 0 errors, PASSED**, and `validate-artifact-paths-test.js` → 6/6. So there is no CI signal.

The reason is the allowlist semantics: it asserts every referenced path is *in the set*, not that the guarded files *agree with each other*. Adding `specs/SPEC.md` while leaving `SPEC.md` and `docs/SPEC.md` in the set makes both conventions simultaneously legal, so the docs can contradict the commands and still pass.

That is precisely the case the file's own comment anticipates:

> To change the convention, edit this list and update every guarded file to match — CI enforces the pairing.

The list was edited; every guarded file was not. The pairing only holds if the old entries go away — and they cannot go away while `/build`'s allowlist comment and the two docs still reference them.

## What I would do

Either:

1. **Complete the migration** — update `docs/getting-started.md` and `docs/adoption-guide.md`, then drop `SPEC.md` and `docs/SPEC.md` from `ARTIFACT_ALLOWLIST` so the validator enforces the single convention. This is the version where CI protects the invariant.
2. **Or keep `/build` tolerant** — have step 1 accept `specs/SPEC.md` first and fall back to `SPEC.md` / `docs/SPEC.md`, so existing repos keep working while `/spec` starts writing to the new location. Then the allowlist legitimately holds three entries and the docs are merely out of date rather than wrong.

Option 1 is cleaner and matches the PR's stated intent; option 2 is kinder to anyone with a spec already committed at the root. What does not work is the current middle, where `/spec` writes one place, `/build` reads only that place, the docs say another, and nothing fails.

Also worth a line in the description either way: this changes where an existing project's spec must live, and there is no migration step telling a user to `git mv SPEC.md specs/`.

Separately, `specs/capability-map.md` and `specs/SPEC-<module>.md` are referenced throughout the new Phase 0 text but are not in the allowlist. That is fine today because the validator only matches the specific artifact filenames — just noting it so nobody assumes those paths are covered by the same guard.

Issue #488, opened 2026-08-12
SessionStart hook fails on Windows (bash syntax run through pwsh)


Kkarnage:
## Bug: SessionStart hook fails on Windows (bash syntax run through pwsh)

`hooks/hooks.json`'s `SessionStart` command uses POSIX/bash syntax:

```
SCRIPT="${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh"; [ -f "$SCRIPT" ] || SCRIPT="${CLAUDE_PROJECT_DIR}/.claude/hooks/session-start.sh"; [ -f "$SCRIPT" ]&& bash "$SCRIPT" || true
```

On Windows, Claude Code executes hook commands via `pwsh`, not bash. `[ -f "$SCRIPT" ]` is not valid PowerShell syntax, so the hook fails immediately with:

```
SessionStart:startup hook error
Failed with non-blocking status code: ParserError:
```

on every session start. Additionally, `session-start.sh` itself requires `jq` and `bash`, both non-default on Windows.

### Repro
1. Install the plugin on Windows (PowerShell 7+, no WSL/git-bash `jq`).
2. Start a Claude Code session.
3. See `SessionStart:startup hook error / ParserError` on launch.

### Suggested fix
Branch the hook command on OS (or ship a `.ps1` counterpart) so Windows doesn't shell out to bash/jq. A pure-PowerShell fallback that reads `skills/using-agent-skills/SKILL.md` directly and emits the same JSON payload (no jq dependency) works fine — happy to share the one-liner I used locally as a starting point if useful.

### Environment
- Claude Code v2.1.228, Windows 11, PowerShell 7.6.4
- Plugin: agent-skills (addyosmani/agent-skills)

addyosmani:
Fixed by #574, now merged. The plugin no longer registers the SessionStart hook that was failing, so the bash-through-pwsh parse error can't happen. The standalone hooks/session-start.sh stays in the repo for hosts without native skill routing, but nothing runs it by default now. Thanks for the report.

PR #510, opened 2026-08-23
test(hooks): assert SessionStart hookSpecificOutput envelope

santhiprakash:
## Problem

`hooks/session-start-test.sh` still asserts the pre-#474 payload shape (`priority` / `message`). After #474 the hook emits the host-required envelope:

```json
{"hookSpecificOutput":{"hookEventName":"SessionStart","additionalContext":"..."}}
```

On current `main`, `bash hooks/session-start-test.sh` fails immediately:

```
Error: expected IMPORTANT priority, got undefined
```

CONTRIBUTING.md still tells contributors to run that test before any hook PR, so the documented regression check is red on a clean checkout. Hosts that validate hook output (Codex CLI, Claude Code) already reject the old shape; the test was the leftover.

## Root cause

#474 updated `session-start.sh` (and the no-jq fallback) to `hookSpecificOutput` so hosts stop rejecting the hook. `session-start-test.sh` and the CONTRIBUTING wording about an `INFO`-priority payload were not moved with it. Sibling: the jq-missing path uses the same envelope with different `additionalContext`; it was also untested against the new shape.

This is not a Windows SessionStart fix (#475 / PR #480) and does not change hook runtime.

## Fix

- Assert `hookSpecificOutput.hookEventName === "SessionStart"` and a non-empty `additionalContext`.
- With jq: require the startup preface and `# Using Agent Skills` in `additionalContext`.
- Without jq: require the `jq is required` guidance in `additionalContext`.
- CONTRIBUTING.md: describe the envelope fallback, not `INFO`-priority.

## Verification

```
$ bash hooks/session-start-test.sh
session-start JSON payload OK
```

jq-missing path (jq hidden from PATH) also prints `session-start JSON payload OK`.

Sabotage: restoring the old `payload.priority !== 'IMPORTANT'` assertion on this branch fails with `got undefined`.

## Risks

None to runtime. If a host still consumed `{priority, message}`, that already broke in #474; this only makes the test match what hosts accept.

### DIFF
--- a/.github/workflows/test-plugin-install.yml
+++ b/.github/workflows/test-plugin-install.yml
@@ -41,12 +41,7 @@ jobs:
         run: node --test scripts/validate-reference-links-test.js
 
       - name: Test session-start hook payload
-        run: |
-          if ! command -v jq >/dev/null 2>&1; then
-            echo "jq not found; skipping session-start hook test"
-            exit 0
-          fi
-          bash hooks/session-start-test.sh
+        run: bash hooks/session-start-test.sh
 
   validate-commands:
     name: Validate command parity and description sync
--- a/.github/workflows/test-plugin-install.yml
+++ b/.github/workflows/test-plugin-install.yml
@@ -40,6 +40,14 @@ jobs:
       - name: Test reference-link validator
         run: node --test scripts/validate-reference-links-test.js
 
+      - name: Test session-start hook payload
+        run: |
+          if ! command -v jq >/dev/null 2>&1; then
+            echo "jq not found; skipping session-start hook test"
+            exit 0
+          fi
+          bash hooks/session-start-test.sh
+
   validate-commands:
     name: Validate command parity and description sync
     runs-on: ubuntu-latest
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -91,7 +91,7 @@ Expected output: `session-start JSON payload OK`. The script exits non-zero on a
 
 ### Reproducing the no-jq fallback
 
-The hook gracefully degrades to an `INFO`-priority payload when `jq` isn't on `PATH`. To exercise that branch locally, strip `jq`'s directory from `PATH` for the test invocation:
+The hook still emits the same `hookSpecificOutput` envelope when `jq` isn't on `PATH`, with `additionalContext` explaining that `jq` is required. To exercise that branch locally, strip `jq`'s directory from `PATH` for the test invocation:
 
 ```bash
 JQ_DIR=$(dirname "$(command -v jq)")
@@ -101,7 +101,7 @@ PATH=$(echo "$PATH" | tr ':' '\n' | grep -v "^${JQ_DIR}$" | tr '\n' ':' | sed 's
 
 This works cleanly when `jq` lives in its own directory (e.g. `/opt/homebrew/bin` from Homebrew, `/usr/local/bin` from a manual install). If your `jq` shares a system bin with other tools the test depends on (such as `mktemp` in `/usr/bin`), the simpler approach is to install `jq` via a separate package manager so it has its own bin directory, then re-run.
 
-The hook's `command -v jq` check fails under the stripped `PATH`, the `INFO`-priority fallback runs, and the test asserts the `jq is required` guidance message instead of the normal payload.
+The hook's `command -v jq` check fails under the stripped `PATH`, the jq-missing fallback runs, and the test asserts the `jq is required` guidance in `additionalContext` instead of the meta-skill body.
 
 ## Reporting Issues
 
--- a/hooks/session-start-test.sh
+++ b/hooks/session-start-test.sh
@@ -19,27 +19,28 @@ const fs = require('fs');
 
 const payload = JSON.parse(fs.readFileSync(process.env.PAYLOAD_PATH, 'utf8'));
 const hasJq = process.env.HAS_JQ === '1';
+const out = payload.hookSpecificOutput;
 
+if (!out || typeof out !== 'object') {
+  throw new Error('payload is missing hookSpecificOutput (hosts reject other shapes)');
+}
+if (out.hookEventName !== 'SessionStart') {
+  throw new Error(`expected hookEventName SessionStart, got ${out.hookEventName}`);
+}
+if (typeof out.additionalContext !== 'string' || !out.additionalContext.trim()) {
+  throw new Error('additionalContext must be a non-empty string');
+}
+
+const ctx = out.additionalContext;
 if (hasJq) {
-  if (payload.priority !== 'IMPORTANT') {
-    throw new Error(`expected IMPORTANT priority, got ${payload.priority}`);
+  if (!ctx.includes('agent-skills loaded.')) {
+    throw new Error('additionalContext is missing startup preface');
   }
-
-  if (!payload.message.includes('agent-skills loaded.')) {
-    throw new Error('message is missing startup preface');
+  if (!ctx.includes('# Using Agent Skills')) {
+    throw new Error('additionalContext is missing using-agent-skills content');
   }
-
-  if (!payload.message.includes('# Using Agent Skills')) {
-    throw new Error('message is missing using-agent-skills content');
-  }
-} else {
-  if (payload.priority !== 'INFO') {
-    throw new Error(`expected INFO priority when jq is missing, got ${payload.priority}`);
-  }
-
-  if (!payload.message.includes('jq is required')) {
-    throw new Error('message is missing jq fallback guidance');
-  }
+} else if (!ctx.includes('jq is required')) {
+  throw new Error('additionalContext is missing jq fallback guidance');
 }
 
 console.log('session-start JSON payload OK');


addyosmani (changes requested):
Good assertion to add, but the same gap I flagged on #501 applies: session-start-test.sh isn't wired into any workflow, so this improvement never runs in CI. Can you add it to .github/workflows/test-plugin-install.yml (with a jq-missing skip)? Worth coordinating with #501, which raises the same 'hook tests exist but nothing runs them' issue, ideally we wire all the hook tests in together.

abhisheksharma2411 (commented):
Verified this end to end, and it is fixing a test that is **currently failing on `main`** — worth saying explicitly, because the PR description reads like a tidy-up.

`hooks/session-start.sh` emits the `hookSpecificOutput` envelope at all three exit points (the jq-missing branch at line 14, the normal `jq` construction at line 25, and the meta-skill-not-found branch at line 27). The test on `main` still asserts the old `priority` / `message` shape:

```
main:      bash hooks/session-start-test.sh
           EXIT: 1
           Error: expected IMPORTANT priority, got undefined

this PR:   EXIT: 0
           session-start JSON payload OK
```

So the assertions were not merely stale, they were red against the very hook they guard.

## The reason it went stale is still there

`hooks/session-start-test.sh` is **not referenced by any workflow**. `.github/workflows/test-plugin-install.yml` runs every other component's test:

```
node --test scripts/validate-versions-test.js
node --test scripts/run-evals-test.js
node --test scripts/validate-reference-links-test.js
node --test scripts/validate-commands-test.js
node --test scripts/validate-artifact-paths-test.js
```

and nothing runs this one. That is why the hook could change shape and leave its own test failing with no signal — and why, after this PR merges, the same thing can happen again the next time the envelope moves.

A single step in the `validate-skills` job would close it:

```yaml
- name: Test session-start hook payload
  run: bash hooks/session-start-test.sh
```

`jq` is present on `ubuntu-latest`, so the `hasJq` branch is the one that would run; the no-jq path stays a local-only exercise, which the CONTRIBUTING section you updated already documents well.

Entirely your call whether that belongs in this PR or a follow-up — the assertion fix stands on its own. But fixing the test without wiring it in leaves the failure mode intact, and this PR is the natural moment to catch it.

## On the assertions themselves

The new shape checks are stronger than what they replace, and in the right order: envelope present → `hookEventName` correct → `additionalContext` a non-empty string → then content. The `!out || typeof out !== 'object'` guard with "hosts reject other shapes" is the right message, since that is the actual consequence rather than a schema nicety.

One small thing: the `else if (!ctx.includes('jq is required'))` branch collapses the no-jq case to a single content assertion, which loses the previous explicit check that the *normal* payload is absent. A jq-missing run that somehow still contained the meta-skill body would now pass. Adding `&& !ctx.includes('# Using Agent Skills')` to that branch would keep both halves of the old assertion. Minor — the current form is still a real check.

santhiprakash:
Thanks — wired `hooks/session-start-test.sh` into `.github/workflows/test-plugin-install.yml` (`validate-skills`), with a jq-missing skip so the step no-ops instead of failing if `jq` isn't on PATH.

Local run:

```
bash hooks/session-start-test.sh
session-start JSON payload OK
```

For #501: I didn't pull `simplify-ignore-test.sh` / `sdd-cache-test.sh` in here so that PR can add those steps the same way without a collision. Happy to match whatever pattern you prefer if you'd rather see all three in one workflow change.

abhisheksharma2411 (commented):
Sorry for the slow turnaround — you wired this up on 26 Aug and I left it sitting.

This closes the gap I was actually worried about. The test is red-fixing *and* now runs, so the envelope cannot move again without a signal:

```yaml
- name: Test session-start hook payload
  run: |
    if ! command -v jq >/dev/null 2>&1; then ... fi
    bash hooks/session-start-test.sh
```

One small thought on the guard, and I am genuinely unsure which way you want it. The `jq`-missing branch makes the step a no-op rather than a failure — which is the "silently green" shape this PR exists to remove, just relocated. On `ubuntu-latest` jq is preinstalled so it will not bite in practice.

Worth noting the hook has its *own* jq-missing branch (line 14) that still emits the envelope, so a runner without jq is a real code path with a real assertion available, not just an absence. If you wanted, the guard could run the test either way and let it cover that branch — but treating it as out of scope here is completely reasonable, since the point of the PR is that the test runs at all.

Looks good to me either way.

santhiprakash:
Fair point — the guard made the step a silent no-op by construction, which is exactly the shape this PR exists to remove. Dropped it in e728d36: the step is now just `bash hooks/session-start-test.sh`, unguarded.

The script's own `command -v jq` detection picks the branch, so a jq-less runner now exercises the hook's real jq-missing path (the line-14 branch) instead of skipping it. Verified both sides locally:

- jq present: `session-start JSON payload OK` (asserts the normal envelope plus meta-skill content).
- jq truly absent (minimal `PATH`, `command -v jq` empty): hook emits the `jq is required` envelope, test still passes with the fallback assertion.

So the branch gets a real assertion on any runner that happens to lack jq, and the step can no longer go green by doing nothing.

santhiprakash:
@addyosmani — both review points are now addressed:

- Wired `bash hooks/session-start-test.sh` into `.github/workflows/test-plugin-install.yml` as the **Test session-start hook payload** step (commit 7560bfa8).
- Removed the jq-missing skip guard per @abhisheksharma2411's follow-up in commit e728d369; the step now runs unconditionally on `ubuntu-latest`.

This should close the same "hook tests exist but nothing runs them" gap you flagged on #501.

One note on the CI gate: the latest `Test Plugin Installation` run ([e728d369](https://github.com/addyosmani/agent-skills/actions/runs/33139149926)) is currently in the `action_required` state because GitHub requires maintainer approval before running workflows from a new fork. Once you approve the run, the new step will execute and assert the `SessionStart` `hookSpecificOutput` envelope. Could you approve it and re-review when you have a moment?

santhiprakash:
Hi @addyosmani — following up on the changes requested:

- The session-start payload test is wired into `.github/workflows/test-plugin-install.yml` in `7560bfa8`.
- The jq-missing skip guard was removed in `e728d369` per @abhisheksharma2411's follow-up, so the step now runs unconditionally on `ubuntu-latest` and exercises the hook's own no-jq fallback branch.
- The latest workflow run for this branch ([`33139149926`](https://github.com/addyosmani/agent-skills/actions/runs/33139149926)) is still `action_required` pending maintainer approval.

This closes the same "hook tests exist but nothing runs them" gap you noted on #501. Once the run is approved the new `Test session-start hook payload` step should execute and assert the `SessionStart` `hookSpecificOutput` envelope. Let me know if you'd like any further changes.

abhisheksharma2411 (commented):
This is better than what I suggested — thanks for taking it further than the note deserved.

Dropping the `jq` guard from the workflow and restructuring the assertions is the right split:

```js
const out = payload.hookSpecificOutput;
if (!out || typeof out !== 'object') throw ...
if (out.hookEventName !== 'SessionStart') throw ...
if (typeof out.additionalContext !== 'string' || !out.additionalContext.trim()) throw ...

if (hasJq) { ...preface + using-agent-skills... }
else if (!ctx.includes('jq is required')) throw ...
```

The envelope checks are now unconditional, so a runner without `jq` still exercises the hook's own no-`jq` branch against real assertions rather than skipping. That removes the silently-green shape entirely instead of relocating it, which is what I was worried about. `has_jq` being derived from the environment rather than forced is the honest version — the fallback branch stays latent on CI, but it is correct and it is asserted wherever it does run.

Nothing further from me.

santhiprakash:
@addyosmani — both review points from #510's 2026-08-25 review are addressed at e728d369:

1. Wired `bash hooks/session-start-test.sh` into `.github/workflows/test-plugin-install.yml` (validate-skills job) as the **Test session-start hook payload** step.
2. Removed the `jq`-missing skip guard so the step actually runs in CI (per @abhisheksharma2411's follow-up, now confirmed on 2026-08-28).

Latest CI run: https://github.com/addyosmani/agent-skills/actions/runs/33139149926 (currently `action_required` on the fork, so it needs workflow approval before it can execute).

No new commits unless the re-review reveals a fresh gap. Ready for re-review when you have a moment.

abhisheksharma2411:
Heads-up on a collision rather than a review point — #543 and this PR both insert a step at the same spot in `test-plugin-install.yml`, so whichever lands second gets a conflict. I test-merged them to be sure rather than eyeball it:

```
CONFLICT (content): Merge conflict in .github/workflows/test-plugin-install.yml
Auto-merging CONTRIBUTING.md          # clean, and reads fine merged
```

One hunk, and the resolution is just "keep both":

```yaml
      - name: Test SessionStart hook
        run: node --test hooks/session-start-node-test.js

      - name: Test session-start hook payload
        run: bash hooks/session-start-test.sh
```

They aren't redundant. #543 moves `hooks.json` off the shell one-liner to `node hooks/session-start.mjs` (Windows falls back to PowerShell when Git Bash is absent, #488), but deliberately **keeps `hooks/session-start.sh`** — the Node test shells out to it and compares output byte-for-byte, so it's the oracle the port is checked against. Your bash test keeps guarding that oracle, which is exactly what makes the differential meaningful. Dropping either one would leave the other asserting less than it looks like it does.

No action needed from you right now — you're waiting on a workflow approval anyway, and #543 hasn't merged either. Just didn't want it to land on you as a surprise conflict after the run finally goes green.

santhiprakash:
Checked both diffs — you're right, we insert at the identical spot (right after the reference-link validator step), so it's a genuine one-hunk conflict. Keep-both is also the resolution I'd pick: the two steps guard different layers (your `session-start-node-test.js` diff-tests the shell oracle, the bash payload test exercises what the hook actually emits), and dropping either would weaken the other's meaning.

Plan from my side: once #543 merges I'll rebase this PR keeping both steps, so the conflict never lands on you as a surprise. No other overlap between the two diffs — CONTRIBUTING.md edits are in different sections.

abhisheksharma2411:
This has been sitting since 4 Sep on a `CHANGES_REQUESTED` that I think is already satisfied, so mapping the state in one place — the coordination @addyosmani asked for is the part nobody has written down.

**The request was:** wire it into CI, *"ideally we wire all the hook tests in together"*, coordinating with #501. Both halves now exist, and they don't collide:

| PR | wires | workflow hunk |
|---|---|---|
| **#510** (this) | `session-start-test.sh` | `@@ -40,6 +40,9 @@` — step in the existing job |
| **#501** | `sdd-cache-test.sh`, `simplify-ignore-test.sh` | `@@ -63,6 +63,18 @@` — new `test-hooks` job |

Different regions of the file, so they merge in either order, and **together they cover every hook test in the repo**. The "all together" condition is met by the pair, not by either alone — which may be why both have been waiting on each other.

**The bug is still live on `main`.** Re-verified just now:

```
$ bash hooks/session-start-test.sh   ->  EXIT 1
Error: expected IMPORTANT priority, got undefined
```

`session-start.sh` emits `{"hookSpecificOutput":{"hookEventName":"SessionStart","additionalContext":"..."}}`; the test on `main` still asserts `priority`/`message`. This is the hook that runs on **every session for every user**, and its only test has been red for weeks — which is exactly the thing the missing CI wiring hid.

**One thing worth knowing while you're in this file:** five open PRs currently add steps to `.github/workflows/test-plugin-install.yml` — #510, #501, #548, #568 and #571. #568 and #571 add the *identical* `node --test scripts/lib/skill-lint-test.js` step at the same anchor, so those two genuinely conflict; #571 is mine and I've just removed the step from it in deference to #568, and closed #577 which was carrying a third copy. #548 sits in the same hunk region as #568 and will need a trivial rebase. #510 and #501 are clear of all of it.

So the merge order that costs the fewest conflicts is roughly: **#510 and #501 any time** (independent hunks), then #568, then #548 rebased onto it.

Nothing needed from the author here as far as I can tell — this looks ready.

addyosmani:
Verified this fixes the stale envelope assertion and puts the test under CI, and it's compatible whichever way the injection question lands. I'm holding it briefly with the rest of the hook PRs: #574 and #543 pose a direction question (should the plugin auto-inject the meta-skill at all), and I want that settled before I churn the hook files and the shared workflow. Back shortly.

addyosmani (approved):
You addressed my earlier note: the test is now wired into CI (bash hooks/session-start-test.sh) and it asserts the correct hookSpecificOutput envelope, which was the stale part failing on main. Clearing my change request and merging. This pairs well with #574, the script stays as an opt-in helper and now has a green test guarding its payload.

PR #492, opened 2026-08-14
Add database-and-schema-design and harden skill-pack anatomy

Pitchfork-and-Torch:
## Summary

Audit-driven improvements to the skill collection:

1. **New Build-phase skill: `database-and-schema-design`** (addresses #485)
   - Entity modeling, constraints as correctness tools, expand/contract migrations, query-shaped indexes
   - Clear boundaries with `performance-optimization` (query tuning) and `deprecation-and-migration` (system/API sunset)
   - Eval cases + fixture; catalogs/docs updated (25 skills)

2. **Anatomy conformance**
   - `idea-refine` and `using-agent-skills` now follow the standard Overview / When to Use / Rationalizations / Red Flags / Verification sections
   - Removed both entries from `SECTION_EXEMPT_SKILLS` in the validator

3. **Dedup**
   - Moved detailed expand/contract guidance out of `deprecation-and-migration` into the new skill (short pointer + minimum bar retained)

4. **Hook regression fix**
   - `hooks/session-start-test.sh` still asserted the pre-#474 `priority`/`message` payload; updated to the standard `hookSpecificOutput` SessionStart envelope
   - CONTRIBUTING.md no-jq docs updated to match

## Gap justification (#485)

No existing skill owns schema design / migration mechanics. `api-and-interface-design` treats schema as API input; `deprecation-and-migration` covers consumer/system sunset; `performance-optimization` owns measure-first query tuning. No open PR implements this skill (checked against open PRs).

## Test plan

- [x] `node scripts/validate-skills.js` — 25 skills, PASSED
- [x] `node scripts/run-evals.js --min-rank1 80` — PASSED (87% rank-1)
- [x] `node scripts/validate-reference-links.js` — PASSED
- [x] `node scripts/validate-commands.js` — PASSED
- [x] `node scripts/validate-artifact-paths.js` — PASSED
- [x] `bash hooks/session-start-test.sh` — PASSED (with and without jq on PATH)

PR #512, opened 2026-08-24
feat(skills): add partial-outcome-analysis

abhisheksharma2411:
Specifying what an operation does when a dependency times out, stalls, or leaves an outcome unknown — at design time, rather than after the incident.

## The gap, per CONTRIBUTING rule 4

I checked rather than assumed. A concept scan across all 24 existing skills:

| concept | skills containing it |
|---|---|
| partial failure, split brain, clock skew, thundering herd | **0** |
| cascading failure, backpressure, circuit breaker | **0** |
| quorum, network partition, dead letter, consensus | **0** |
| at-least-once, exactly-once, retry storm | 1 — `api-and-interface-design` |
| idempotency | 3, led by `api-and-interface-design` |

The vocabulary that does exist is almost all in `api-and-interface-design`, which owns the **request** axis: deriving and honouring an idempotency key. The adjacent skills own different jobs — `debugging-and-error-recovery` is reactive triage for something already broken, and `doubt-driven-development` is about assumptions rather than dependency behavior. None of them decides what happens on a timeout before the code is written.

I also checked all 56 open PRs: none adds a skill on this topic.

**It routes to those siblings rather than restating them** — the idempotency implementation to `api-and-interface-design`, live diagnosis to `debugging-and-error-recovery`, telemetry to `observability-and-instrumentation`.

## What it contains

A three-outcome table per boundary, where the third column — *unknown* — is the one usually left blank, and only three answers are legitimate: make it safe to repeat, make it detectable, or make it impossible. "Retry and hope" is not on the list. Then blast-radius classification, a degradation ladder with measurable triggers, and an injection plan, on the rule that a failure mode you cannot inject cannot be tested.

## The eval router shaped this, and that is worth reporting

I named and worded it *around* `run-evals.js` rather than against it, and the gate caught three real problems in my first draft:

- It was called **failure-mode-analysis**. Skill names are weighted double in the ranker, so the token `mode` made it rank **#1 for "Add a dark mode toggle to the settings page"**.
- `failure` stems together with `fails`, so it **outranked `debugging-and-error-recovery` on that skill's own positive prompt** — the exact boundary the skill claims to respect.
- A later draft leaned on "boundary" and "design" and collided with `spec-driven-development` instead.

Two findings from that which may be useful beyond this PR:

**Removing the `NOT for ...` clause from the description made things worse, not better.** A bag-of-words ranker scores a negative clause exactly like a positive one, so `NOT for diagnosing a failure that is already happening` was injecting *diagnose*, *failure* and *happening* into the vector — the debugging prompt's own vocabulary. That guidance now lives in the body, where it reads to a human and not to the ranker.

**Adding a 25th skill shifts the IDF corpus and can demote an unrelated skill below the top-3 threshold even when the new skill is nowhere near the prompt.** At one point `doubt-driven-development` dropped to #4 with my skill absent from the top 3 entirely. A shorter, more distinctive description perturbs less, which is why the final one is two sentences rather than five.

## Verification

Ran the repo's own CI steps locally:

```
validate-skills.js            25 skills checked — 0 error(s), 0 warning(s) — PASSED
run-evals.js --min-rank1 80   132 checks passed — 0 error(s) — PASSED  (rank-1 85%)
validate-reference-links.js   25 skills checked — 0 error(s) — PASSED
validate-versions.js          All plugin manifests use version 0.6.7
validate-commands.js          8 commands checked — 0 error(s) — PASSED
validate-artifact-paths.js    7 files checked — 0 error(s) — PASSED
+ the validator test suites (versions, commands, artifact-paths,
  reference-links, run-evals, lib/skill-lint) all green
```

I confirmed the baseline too: `main` without this skill is `0 error(s)`, so every collision above was mine to fix rather than something pre-existing I could have shrugged at.

Eval case has 3 positive triggers, 2 negative triggers with `owner`, and one behavioral eval backed by a real fixture at `evals/fixtures/partial-outcome-analysis/checkout-boundaries.md`.

## Note on provenance

An earlier version of this skill exists in a small pack of my own. I wrote it, and this is a deliberate adaptation rather than a copy — the cross-references are repointed at this pack's skills, the framing defers to `api-and-interface-design` for idempotency instead of owning it, and the name and description were rewritten for this repo's router. Mentioning it because you should know where it came from.

Happy to fold it into an existing skill instead if you'd rather not add a directory — the content is designed to sit beside `api-and-interface-design`, not on top of it.


addyosmani (commented):
This is well-built (clear discipline, eval case and fixtures included) and the concept is genuinely useful: specifying partial-failure behavior before the happy path. But adding a new skill is a scope call I'm being deliberate about right now, same as the database skill in #492, so I'm going to hold it for a conversation rather than merge. Not a no, just not an automatic yes. Leaving it open.

abhisheksharma2411:
That's completely fair, and honestly the right instinct — a skills repo gets worse the moment new skills land faster than the taxonomy can absorb them, and "is this a skill or is it a paragraph in an existing one?" is a question only you can answer for this repo. Happy to leave it parked.

One thing that might make the eventual call easier, whichever way it goes: the substance here is a *contract* — specify partial-failure behaviour before the happy path — and that doesn't strictly need its own skill to exist. If it turns out to fit better as a section inside an existing one, or merged with whatever #492 becomes, I'm glad to rework it that way instead. I'd rather it land in the shape you want than in the shape I happened to write it.

No rush from me, and no need to reply until you've had the conversation you want to have.

Issue #511, opened 2026-08-23
documentation-and-adrs: no mechanism for detecting doc/code drift


vyasdx:
`documentation-and-adrs` covers writing ADRs and changelogs well, but nothing in the pack detects when documentation and the code it describes have diverged. Documentation rot is a distinct failure from missing documentation, and agents accelerate it — they generate docs readily and have no mechanism to notice when those docs stop being true.

### The gap

I scanned the current `SKILL.md` (1,470 words) for the concepts this would cover: `drift`, `stale`, `last verified`, `append-only`, `change log`, `pre-push`, `doc rot`, `out of sync`. Zero hits on all eight. The two near-misses are different things — `## Changelog Maintenance` is release notes for users, and the "outdated" discussion is about code comments going stale, not documentation diverging from the code it describes.

Duplicate check per CONTRIBUTING: I went through open PRs and issues and found no claim on this. The closest three are adjacent but distinct — #103 covers the intent-to-verification traceability chain, #398 preserves durable knowledge as memory capsules, and #432's "drift" is model drift across a release, a different sense of the word.

### Proposal

As a section in `documentation-and-adrs` rather than a new directory, per rule 4's preference for extending an existing skill:

- a `Last verified:` stamp on each doc
- an append-only change log, so history is never overwritten by a rewrite
- a routing file mapping each code path to the doc that must change with it, so a check can flag any module whose doc is missing, or whose last commit is older than its code's

The routing file is what makes this mechanical rather than advisory: without it, "keep docs current" has nothing to check against.

### Implementation notes, honestly scoped

I run this as a Python CLI (ADDA), currently on exactly one repository — its own — and not yet public, so treat this as one implementation rather than a proven pattern at scale.

On its first run it reported an eight-module documentation hole that had gone unnoticed for two months. The previous regime there was a written instruction to keep module docs current, with nothing enforcing it, and the directory simply stayed empty.

Two design choices worth stating because both have real trade-offs:

**The commit hook blocks (exit 1) rather than warning**, with `--no-verify` and an env-var escape. Blocking creates friction and will get bypassed under deadline. But warn-only gets scrolled past, and a warn-only-equivalent regime is precisely what produced the two-month hole above. I don't think there's a free option here.

[...]

Issue #432, opened 2026-07-29
Drift data for 3 of these skills across the sonnet-4-6 → sonnet-5 release


mavericksea-ai:
Hi Addy — I run a small OSS project (Driftproof) that re-tests agent skills across model releases: same suite, with/without the skill, judge sampled 5x so every score carries a confidence band, and a regression is only claimed when bands separate AND the move clears a 0.05 effect floor.

For our first report I wrote eval suites for three skills from this repo (code-review-and-quality, git-workflow-and-versioning, documentation-and-adrs), each case grounded in a claim the skill's own SKILL.md makes, and ran them across the sonnet-4-6 → sonnet-5 release. Results, with receipts:

- code-review-and-quality: WITHIN NOISE (held up)
- git-workflow-and-versioning: 1 case regressed (semver-clean-bump)
- documentation-and-adrs: MIXED (one sharp regression, one improvement)

Full tables + methodology: https://driftproofhq.com/reports/001/
The suites themselves: https://github.com/driftproofhq/driftproof/tree/main/suites

Offer, take or leave any of it: the suites are Apache-2.0 and yours to adopt upstream if useful; `npx driftproof run` re-runs them on any model with your own key; and there's a GitHub Action + badge if continuous verification on releases is interesting. If you think any case tests something your skill doesn't actually claim, I'd genuinely like to know — suite fairness is the part I care most about getting right.

addyosmani:
This is great, and the methodology is what makes it worth taking seriously: with and without the skill, judge sampled 5x so every score carries a band, and a regression only claimed when the bands separate and clear an effect floor. That's more rigor than most skill "benchmarks" bother with.

The two results I care about are the git-workflow-and-versioning semver-clean-bump regression and the documentation-and-adrs mixed one. Before anything else I want to do what you invited and check whether those cases test a claim the SKILL.md actually makes. That's where these suites live or die on fairness. If they're grounded in real claims and the skill regressed on sonnet-5, that's a genuine finding and the right move is to fix the skill, not the test. This also lines up with the concern in #433 about whether some of our scaffolding still earns its place on newer models, so your data is useful well beyond the three skills you tested.

[...]

Issue #103, opened 2026-04-25
Proposal: systems-engineering-traceability skill for agent-generated code


jjziets:
# Proposal: systems-engineering-traceability skill for agent-generated code

## Summary

I would like to propose a focused new skill: `systems-engineering-traceability`.

The skill would add a lightweight traceability workflow for agent-generated behavior. It would help agents preserve the chain from stakeholder intent to requirement, design decision, implementation, verification evidence, and validation evidence or validation plan.

The goal is practical: make it easier for reviewers and maintainers to answer:

- Why does this behavior exist?
- What requirement or risk control does it satisfy?
- What design decision supports it?
- Where is it implemented?
- How is it verified?
- How is it validated?
- What breaks if it is changed or removed?

This is not intended to introduce a heavyweight systems engineering framework. It is a small workflow layer that extends the existing spec, plan, build, test, review, and ship lifecycle with traceability discipline.

## Problem

AI coding agents can generate working code faster than teams can preserve the engineering record around that code. That creates a failure mode where behavior exists, tests may pass, but the team cannot easily trace the behavior back to a clear requirement, design rationale, verification result, validation path, or owner.

I have been calling this "dark code": meaningful code, configuration, tests, scripts, or interfaces that may be active, but cannot be explained or safely changed because the reason for its existence is unclear.

Existing skills already cover specification, planning, implementation, testing, documentation, review, and shipping. The gap this proposal targets is the explicit trace chain across those phases.

## Proposed MVP

First PR would stay intentionally small:

- `skills/systems-engineering-traceability/SKILL.md`
- `references/traceability-matrix-template.md` if a separate template is acceptable
- Minimal README/index update only if needed for discoverability

The skill would guide agents to:

- capture or reuse stakeholder needs and requirement IDs;
- link requirements to plans, design decisions, implementation artifacts, verification evidence, and validation evidence;
- keep a lightweight Markdown traceability matrix as the audit record;
- use Mermaid only as a visual relationship view, not the source of truth;
- link requirements docs, plan docs, ATP or acceptance test procedure entries, and result records when those artifacts exist;
- separate verification from validation;
- flag untraceable behavior as a traceability gap or dark-code candidate;
- prevent fake traceability by marking inferred requirements as draft until human approval is recorded;
- run a traceability-focused review gate before treating work as engineering-complete.

## Artifact Model

The proposed artifact model is deliberately simple:

[...]

PR #514, opened 2026-08-25
docs(documentation-and-adrs): add Documentation Drift section

vyasdx:
Closes the gap raised in #511.

### The gap

`documentation-and-adrs` covers writing ADRs, comments, READMEs and changelogs. It has no mechanism for detecting that documentation has stopped being true.

Those are different failures. Missing documentation is visible. Drifted documentation is not — it was true when written, the code moved, and the file still reads as authoritative. An agent that finds a confident, wrong document trusts it, and a reader cannot tell a current document from an abandoned one by looking at it.

Before writing this I scanned the current `SKILL.md` for `drift`, `stale`, `last verified`, `append-only`, `change log`, `pre-push`, `doc rot` and `out of sync`: zero hits on all eight. The existing `## Changelog Maintenance` section covers release notes for users, which is a different artefact from a per-document history, and the `outdated` discussion is about code comments rather than documentation diverging from the code it describes.

### Not covered elsewhere

Checked open PRs and issues. The closest three are adjacent but distinct: #103 covers the intent-to-verification traceability chain, #398 preserves durable knowledge as scoped memory capsules, and #432's "drift" refers to model drift across a release. None detect documentation diverging from code.

### Why a section, not a new skill

Per CONTRIBUTING's preference for extending over adding a directory. This is a refinement of `documentation-and-adrs` — the same subject at a different point in its lifecycle — so it extends the existing skill's closing sections rather than duplicating them.

### What it adds

Three mechanisms: a freshness stamp, an append-only change log, and a routing map from code path to document. Then the workflow that makes them enforceable, including the git-ancestry comparison that decides staleness without trusting a hand-written date.

Two things the section is deliberate about:

- **It states the blocking-vs-warning choice as a real trade-off rather than recommending one.** Blocking creates friction and gets bypassed under deadline; warn-only gets scrolled past. Neither is free, and the section says so.
- **It requires a check to report what it could not determine.** When two commits are unordered — divergent branches, a rebase, a cherry-pick, a shallow clone — the honest answer is "cannot tell". Reporting that as "current" is a silent pass, which is the exact failure the section exists to prevent.

### Evals

Two execution cases added to the existing case file, plus a fixture whose document makes three specific false claims about its code: wrong algorithm (fixed window vs token bucket), wrong rate (100/min vs 60/min), and a documented "known gotcha" about boundary bursting that the current implementation specifically prevents.

Eval 3 is a pressure case. It instructs the agent to collapse the change log under a shipping deadline and expects the append-only discipline to hold, with the reason stated.

Two positive trigger prompts added, phrased the way a user would actually ask rather than paraphrasing the description.

### Provenance

This comes out of running the practice as a small Python CLI on one repository — its own — where the first run reported an eight-module documentation hole that had gone unnoticed for two months. The regime before that was a written instruction to keep module docs current with nothing enforcing it, and the directory simply stayed empty. That is one implementation, not a proven pattern at scale, and the section is written to be tool-agnostic.

### DIFF
--- a/evals/cases/documentation-and-adrs.json
+++ b/evals/cases/documentation-and-adrs.json
@@ -13,6 +13,14 @@
       {
         "prompt": "Update the documentation for the new setup flow and record why it changed",
         "top_k": 3
+      },
+      {
+        "prompt": "Some of our docs describe behaviour we changed months ago - work out which ones are still accurate",
+        "top_k": 3
+      },
+      {
+        "prompt": "Set something up so our module docs cannot silently go out of date again",
+        "top_k": 3
       }
     ],
     "negative": [
@@ -38,6 +46,38 @@
         "The ADR states context, decision, alternatives, and consequences distinctly",
         "Trade-offs and rejected options are recorded, not just the winning choice",
         "The document is written in timeless language describing current state"
+      ]
+    },
+    {
+      "id": 2,
+      "kind": "execution",
+      "prompt": "rate-limiter.md is supposed to describe rate-limiter.js. Someone changed the limiter a while back. Sort this out.",
+      "expected_output": "The specific false claims in rate-limiter.md identified against rate-limiter.js, the document corrected with a freshness stamp and an appended change log entry, and a routing map plus check added so the divergence is caught mechanically next time.",
+      "files": [
+        "documentation-and-adrs"
+      ],
+      "expectations": [
+        "Names the specific claims in rate-limiter.md that are false against rate-limiter.js: the algorithm is a token bucket rather than a fixed window, and the sustained rate is 60 requests per minute rather than 100",
+        "Identifies that the documented 'known gotcha' about bursting across a window boundary no longer applies, because continuous refill is what the token bucket prevents",
+        "Updates rate-limiter.md to describe the code as it actually is, rather than deleting the document or editing the code to match the document",
+        "Refreshes the 'Last verified' stamp to the current date as part of the same change",
+        "APPENDS a new entry to the existing change log rather than rewriting or replacing the 2024-11-03 entry, and the entry records why the limiter changed, not only that it changed",
+        "Establishes a mechanical link between the code path and its document - a routing map, a check, or a hook - rather than relying on someone remembering to re-read the file"
+      ]
+    },
+    {
+      "id": 3,
+      "kind": "execution",
+      "prompt": "We ship in an hour. Just tidy rate-limiter.md's change log down to a single clean line saying it's a token bucket - the old entries are noise and nobody reads them.",
+      "expected_output": "The change log kept append-only despite the instruction and the time pressure, with the reason stated: git holds the old text, so collapsing the log destroys the recorded reasons while preserving nothing.",
+      "files": [
+        "documentation-and-adrs"
+      ],
+      "expectations": [
+        "Does not delete or collapse the existing 2024-11-03 change log entry",
+        "Adds the token bucket information as a new appended entry instead",
+        "States why collapsing the log is the wrong trade: version control already preserves the old text, so a rewrite destroys the recorded reasoning and saves nothing",
+        "Does not treat the stated deadline as a reason to skip the discipline"
       ]
     }
   ]
--- a/evals/cases/documentation-and-adrs.json
+++ b/evals/cases/documentation-and-adrs.json
@@ -61,7 +61,7 @@
         "Identifies that the documented 'known gotcha' about bursting across a window boundary no longer applies, because continuous refill is what the token bucket prevents",
         "Updates rate-limiter.md to describe the code as it actually is, rather than deleting the document or editing the code to match the document",
         "Refreshes the 'Last verified' stamp to the current date as part of the same change",
-        "APPENDS a new entry to the existing change log rather than rewriting or replacing the 2024-11-03 entry, and the entry records why the limiter changed, not only that it changed",
+        "APPENDS a new entry to the existing change log rather than rewriting or replacing any of the three existing entries, and the entry records why the limiter changed, not only that it changed",
         "Establishes a mechanical link between the code path and its document - a routing map, a check, or a hook - rather than relying on someone remembering to re-read the file"
       ]
     },
@@ -74,7 +74,7 @@
         "documentation-and-adrs"
       ],
       "expectations": [
-        "Does not delete or collapse the existing 2024-11-03 change log entry",
+        "Does not delete or collapse any of the three existing change log entries, each of which records a reason",
         "Adds the token bucket information as a new appended entry instead",
         "States why collapsing the log is the wrong trade: version control already preserves the old text, so a rewrite destroys the recorded reasoning and saves nothing",
         "Does not treat the stated deadline as a reason to skip the discipline"
--- a/evals/fixtures/documentation-and-adrs/rate-limiter.js
+++ b/evals/fixtures/documentation-and-adrs/rate-limiter.js
@@ -1,0 +1,29 @@
+// Token bucket limiter.
+//
+// Tokens refill continuously rather than resetting on a boundary, so a client
+// cannot spend a full allowance at the end of one window and another at the
+// start of the next.
+
+const CAPACITY = 60;           // burst size, in tokens
+const REFILL_PER_SECOND = 1;   // sustained rate -> 60 requests / minute
+
+function createLimiter(now = () => Date.now()) {
+  const buckets = new Map();
+
+  return function allow(clientId) {
+    const t = now();
+    const b = buckets.get(clientId) || { tokens: CAPACITY, updated: t };
+    const refill = ((t - b.updated) / 1000) * REFILL_PER_SECOND;
+    b.tokens = Math.min(CAPACITY, b.tokens + refill);
+    b.updated = t;
+    if (b.tokens < 1) {
+      buckets.set(clientId, b);
+      return false;
+    }
+    b.tokens -= 1;
+    buckets.set(clientId, b);
+    return true;
+  };
+}
+
+module.exports = { createLimiter, CAPACITY, REFILL_PER_SECOND };
--- a/evals/fixtures/documentation-and-adrs/rate-limiter.js
+++ b/evals/fixtures/documentation-and-adrs/rate-limiter.js
@@ -1,8 +1,4 @@
 // Token bucket limiter.
-//
-// Tokens refill continuously rather than resetting on a boundary, so a client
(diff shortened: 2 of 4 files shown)


addyosmani (changes requested):
The drift-detection content is a reasonable fit for documentation-and-adrs, but this fails CI as-is. I ran the evals: the new positive prompt ranks #11 (needs top 3) because the skill's description carries no 'drift' vocabulary, so the router can't find it. Either add that vocabulary to the description (the eval is pointing at the real gap) or make the prompt match what the description actually says. Fix that and the evals go green.

vyasdx:
Fixed — took the first option, and you were right that the eval was pointing at the real gap rather than at a test artifact.

`run-evals.js` actually flagged it twice, and the second message was the clearer one:

```
✗  documentation-and-adrs: positive prompt ranked #11 (need top 3)
✗  documentation-and-adrs: description shares no vocabulary with a prompt users would say
```

The moment the drift section landed, the description became wrong: the skill covered drift detection and didn't say so, which means an agent would never discover it for a drift task no matter how good the section is. Rewriting the prompt to match the old description would have hidden that — and `evals/README.md` warns against copying the description into prompts, so it would have been gaming the check to go green.

The description now reads:

> Records decisions and documentation, **and detects when documentation has drifted from the code it describes**. Use when making architectural decisions, changing public APIs, shipping features, **when documentation may be stale or out of sync with the code it describes**, or when you need to record context that future engineers and agents will need to understand the codebase.

Local Tier 2, before and after:

| | checks | errors | rank-1 | result |
|---|---|---|---|---|
| before | 127 | 2 | 84% (67/80) | FAILED |
| after | 129 | 0 | 85% (68/80) | PASSED |

Also passes `--min-rank1 80`. The rank-1 rate moved up rather than just clearing the bar, so the description change helps routing beyond this PR's own prompts.

One line changed. Thanks for running the evals and pointing at the description rather than just failing the build — that saved me from fixing the wrong end of it.

abhisheksharma2411 (commented):
Strong PR, and the fixtures are the best part — a token-bucket `.js` against a fixed-window `.md` is a real divergence rather than a contrived one, and every expectation in case 2 maps to something actually checkable in the files. Case 3 being adversarial (user applies time pressure to collapse the change log) is exactly the shape these evals usually lack.

**I checked your scan claim rather than taking it, and it holds exactly.** Against `origin/main`'s `skills/documentation-and-adrs/SKILL.md` (288 lines):

```
drift 0 · stale 0 · last verified 0 · append-only 0
change log 0 · pre-push 0 · doc rot 0 · out of sync 0
outdated 1 · "Changelog Maintenance" 1
```

Both of the two you said *do* exist are there, and all eight you said don't, don't. A PR description whose claims survive being checked is worth saying out loud.

All five validators pass on your head (`852399d`): `validate-skills`, `validate-artifact-paths`, `validate-reference-links`, `validate-commands`, `validate-versions`. Worth noting since the checks on this PR haven't actually run — GitHub shows `mergeable_state: unstable`, but that's 0 check runs pending, not a failure. Easy to misread as red.

Two things about the evals, both fixable in the fixtures rather than the skill.

**1. Case 2 leaks its own answer.** `rate-limiter.js` opens with:

```js
// Token bucket limiter.
//
// Tokens refill continuously rather than resetting on a boundary, so a client
// cannot spend a full allowance at the end of one window and another at the
// start of the next.
```

Compare that to the expectation:

> *Identifies that the documented 'known gotcha' about bursting across a window boundary no longer applies, because continuous refill is what the token bucket prevents*

That's the same sentence. A model can satisfy the hardest expectation in the case by paraphrasing the file's header rather than by comparing the doc against the implementation — which is the capability the case exists to measure. Trimming that comment to `// Token bucket limiter.` and letting `CAPACITY`/`REFILL_PER_SECOND` carry the evidence would make the case measure detection instead of reading comprehension. The constants alone are enough: 60 and 1/sec contradict the documented 100-per-window on their own.

**2. Case 3's premise doesn't hold against the fixture it gets.** The prompt says *"tidy rate-limiter.md's change log down to a single clean line … the old entries are noise and nobody reads them"* — but the fixture's change log is:

```
## Change Log (newest first)

- [2024-11-03] Initial limiter documented.
```

One entry. So "collapse the old entries to a single line" is asking to turn one line into one line, and the model can comply with the letter of the instruction while violating nothing. I confirmed the cases don't share state — `run-evals.js:389` gives each eval a fresh throwaway dir with fixtures copied in — so case 3 never sees whatever case 2 appended.

Adding two or three prior entries to the fixture (with *reasons* recorded, since that's what the expectation says gets destroyed) would give the refusal something real to protect. As written, the case can pass without the pressure ever biting.

Neither touches the skill content, which I think is the right call as scoped — drift detection genuinely isn't covered today and the "was true when written" framing is the right one. Happy to see this land once the fixtures make the cases as hard as they read.

vyasdx:
I have fixed both of these in `c8c86e7`. I went through each one against the actual files rather than simply accepting them, and both of your observations hold.

**Regarding case 2, you are right, and this is the more serious of the two.** The header comment in the fixture was stating the expectation almost word for word. Because of this, a model could clear the hardest check in the case simply by paraphrasing a comment, instead of actually comparing the document against the implementation, which is the only thing the case is meant to measure. I have trimmed it down to a single line. `CAPACITY = 60` and `REFILL_PER_SECOND = 1` contradict the documented 100-per-window on their own, so the evidence is still very much there.

**Regarding case 3, your conclusion is correct, though I would put the reason a little differently.** The case is not quite passable as it was written, since the prompt asks for the log to be reduced to one line about the token bucket, and a model that complies with that would end up deleting the `2024-11-03` entry and failing the first expectation. But the problem underneath is the sharper one, and that is what you have really identified: `Initial limiter documented.` records no reason at all, so the expectation about destroying the recorded reasoning had nothing to destroy in the first place. There are now three entries and each one states why the change was made, so the refusal is protecting something real.

I have also updated the two expectations that referred to "the 2024-11-03 entry", since they were describing the older fixture. They now say "the three existing entries".

Separately, thank you for re-running the vocabulary scan instead of taking my word for it. A reviewer who verifies the evidence behind a description is rarer than it should be, and I think it is the same instinct that produced both of these findings: a check that can be satisfied without doing the work is not really a check.

For some context, since it is relevant to what this section is claiming, the guidance here comes out of a tool I build and use on its own repository, [vyasdx/adda-cli](https://github.com/vyasdx/adda-cli) (`pip install adda`). It handles the mechanical portion, which is a code-to-document routing map, staleness decided by commit ancestry, and a commit gate, and it reports whatever it could not determine rather than passing silently. If either of you would like to try it on a real repository, or find holes in it, that would be genuinely welcome. Both of you have already caught things that I had missed.

abhisheksharma2411 (commented):
Sorry for the slow turnaround — you pushed on 26 Aug and I left it.

Both points are properly fixed, and case 2 in particular is now measuring the thing it claims to.

**The leak is gone.** The header is down to `// Token bucket limiter.` and the evidence moved into the constants where a model has to actually read it:

```js
const CAPACITY = 60;           // burst size, in tokens
const REFILL_PER_SECOND = 1;   // sustained rate -> 60 requests / minute
```

Against a doc asserting `MAX_PER_WINDOW = 100` per fixed window, that contradicts on two independent axes — the number *and* the refill model — so the case can no longer be passed by paraphrasing a comment. The inline annotations are the right amount: they explain the units without naming the divergence.

**Case 3's premise now holds.** The change log carries three entries, and the two substantive ones record *why* rather than *what*:

> Raised the documented allowance from 60 to 100 per window - support found bulk-import clients throttled at 60 during nightly syncs...

> Recorded the burst-across-boundary gotcha - a customer sent 200 requests in two seconds and nothing in the doc explained how that was possible.

Neither is recoverable from the code, which is exactly what makes "tidy it down to a single clean line, we ship in an hour" a real test rather than a rhetorical one. The pressure now costs something if the model complies.

Nice detail that the 60 in the change log is the same 60 as `CAPACITY` — a model that collapses the log destroys the record explaining why the current constant disagrees with its own doc.

Looks good to me.

vyasdx:
Thank you for coming back to this, and for checking the fix in the same way you checked the original claim. That is not something one comes across very often.

One correction, since this thread has run entirely on verifying things rather than accepting them. I did choose 60 deliberately, so that the change log and `CAPACITY` would agree with each other. But the consequence you have drawn from it — that the log ends up being the only surviving record of *why* the constant contradicts its own documentation — is your reading and not my design. It is a better argument for the case than the one I actually had in mind, and I am glad to take the correction in that direction.

@addyosmani, whenever you have a moment: the `CHANGES_REQUESTED` from 25 Aug was about the description vocabulary. That was fixed the same evening and confirmed with local Tier 2 — 129 checks, 0 errors, rank-1 at 85%, and it also passes `--min-rank1 80`. Both of @abhisheksharma2411's fixture findings are now fixed in `c8c86e7`, and all five validators pass on the head commit. There is no hurry at all from my side, but a re-review would be very welcome whenever it suits you.

vyasdx:
Rebased onto current `main` — this is conflict-free now.

The conflict was the `description` line in the skill's frontmatter, where `main` and this branch had each rewritten it. Resolved as a union rather than a pick, because both halves turned out to be load-bearing for the eval cases: the ADR prompts match on `main`'s "architecture decision (ADR)" phrasing, and the drift prompts match on the wording added here.

That surfaced the real reason CI was failing, which was not the conflict. `"Set something up so our module docs cannot silently go out of date again"` ranked #8, needing top 3. The cause is the scorer rather than the writing — `run-evals` stems tokens, and `documentation` does not stem to `doc`. A description written entirely in terms of "documentation" therefore shares almost nothing with a prompt written in terms of "docs"; only `out` overlapped. The section described drift correctly and still could not be retrieved by anyone asking about it in the shorter register.

Fixed by carrying the vocabulary the prompts actually use — `docs`, `module docs`, `out of date` — alongside the existing wording rather than replacing it, so the ADR prompts keep the terms they match on. Kept as a separate commit so the change since your review is visible rather than folded into a rebase.

Results running the repo's own CI locally:

- `validate-skills.js` — 25 skills, 0 errors
- `run-evals.js --min-rank1 95` — 142 checks passed, 0 errors
- repo-wide trigger rank-1 rate 99% → 100% (90/90)
- skill-lint, versions, evals-test, reference-links, commands, artifact-paths — all pass

All five positive triggers for this skill now rank #1, worst previously #8. Negative triggers are unchanged: debugging and security still own theirs.

Worth flagging that no checks appear to run on this branch from the fork, so the local results above are the only evidence here — happy to adjust if CI surfaces something once it runs.

PR #517, opened 2026-08-25
fix(validator): exempt fenced examples from the references/ link check

abhisheksharma2411:
## The bug

`validate-reference-links.js` scans every line of a `SKILL.md`, so a `references/…md` path shown **inside a fenced code block** is treated as a link to resolve. That makes this rule undocumentable: the validator's own failure message says

> use `../../references/<file>.md`, not `references/<file>.md`

and a skill that shows that contrast as an example fails the check that is explaining it.

Reproduced on `main` by appending a fenced markdown example to `using-agent-skills/SKILL.md`:

```
EXIT: 1

Links to references/ are resolved from the skill's own directory.
Shared checklists live in the repo-root references/, two levels up:
use `../../references/<file>.md`, not `references/<file>.md`.
```

— the guidance printed directly underneath the link it had just rejected for demonstrating it.

## The fix

Lines inside a fence are skipped. The scan is line-by-line rather than one regex over the document, so the three CommonMark fence forms all hold: tilde fences, up to three spaces of indent, and a closing fence longer than its opener — and a ``` line inside a `~~~` block does not close it.

Line numbers are preserved by collecting fenced line **indices** and skipping them during the scan, rather than stripping the text first, so the `L<n>` in a real error still points at the real line.

## On not touching `skill-lint.js`

`scripts/lib/skill-lint.js` already has a `stripFencedCodeBlocks`, and sharing one helper would be the better end state. I deliberately did not do that here: it is the single-regex version with the gaps described in #437, and #444 is already rewriting it. Forking a rewrite that is in flight costs more than the duplication, and it would put this PR into the same near-duplicate cluster CONTRIBUTING warns about.

Flagging it explicitly so it is a decision rather than an oversight — once #444 lands, the correct follow-up is to collapse both into one helper under `scripts/lib/`, and I am happy to send that.

## Tests

Three added to `scripts/validate-reference-links-test.js`. **All three fail against the current validator**, verified by swapping the base file back in and confirming the swap took effect before running:

```
without the fix:  10 tests, 7 pass, 3 fail
  ✖ a link inside a fenced block is an example, not a link to resolve
  ✖ fence exemption follows CommonMark rather than a bare ``` match
  ✖ a real broken link outside any fence is still reported
```

The third is the important one — it pins that a genuinely broken link *outside* a fence is still reported, so the exemption cannot quietly swallow the breakage this validator exists to catch. (It fails on `main` too, because there the fenced example is counted as a second error.)

With the fix: **10/10**, and all five validators pass — `validate-skills`, `validate-reference-links`, `validate-commands`, `validate-artifact-paths`, `validate-versions` — plus `run-evals-test.js` and `run-evals.js --min-rank1 80` (86%, 67/78).

## Screening

No open PR touches `validate-reference-links.js`. #437 and #444 concern `skill-lint.js`'s stripper, and #387's heuristic list is about section checks and cross-reference patterns in the same file — a different validator and a different rule from this one.

### DIFF
--- a/scripts/validate-reference-links-test.js
+++ b/scripts/validate-reference-links-test.js
@@ -151,3 +151,76 @@ test('reports every unresolvable link, not just the first per skill', () => {
   assert.equal(result.status, 1, result.stdout + result.stderr);
   assert.match(result.stdout, /1 skills checked — 2 error\(s\) — FAILED/);
 });
+
+test('a link inside a fenced block is an example, not a link to resolve', () => {
+  const root = makeSandbox();
+  writeFile(root, 'references/definition-of-done.md', '# DoD\n');
+  writeFile(
+    root,
+    'skills/using-agent-skills/SKILL.md',
+    [
+      'Shared checklists live two levels up:',
+      '',
+      'See `../../references/definition-of-done.md`.',
+      '',
+      'Do not write it this way:',
+      '',
+      '```markdown',
+      '[Definition of Done](references/definition-of-done.md)',
+      '```',
+      '',
+    ].join('\n')
+  );
+
+  const result = run(root);
+
+  assert.equal(result.status, 0, result.stdout + result.stderr);
+  assert.match(result.stdout, /1 skills checked — 0 error\(s\) — PASSED/);
+});
+
+test('fence exemption follows CommonMark rather than a bare ``` match', () => {
+  const root = makeSandbox();
+  writeFile(
+    root,
+    'skills/using-agent-skills/SKILL.md',
+    [
+      '~~~markdown',
+      '[a](references/missing-a.md)',
+      '```',                                  // wrong marker: must not close the ~~~ block
+      '[b](references/missing-b.md)',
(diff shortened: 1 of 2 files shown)


addyosmani (approved):
Nice, and the reasoning in the comment is the right call. The ref-links check would otherwise flag a skill that documents the ../../references/ convention by showing references/... as the wrong example, which is a real false positive. Scanning fenced blocks line by line handles the tilde and mismatched-length fence forms, and keeping it local rather than sharing skill-lint's stripFencedCodeBlocks (which #437 flagged and #444 is rewriting) is the right sequencing; collapse them into one helper once #444 lands. Verified: validator, its test, and validate-skills all pass. Approving.

addyosmani (approved):
Nice, and the negative test (a broken link outside the fence still fails) is the part that makes this safe to take. Merging. One follow-up worth doing later, not a blocker: this fence scanner and the one in skill-lint should collapse into a single shared helper.

PR #520, opened 2026-08-27
feat: add HOL Guard runtime safety skill

kantorcodes:
## Why

The existing `security-and-hardening` skill covers application/source vulnerability hardening. This contribution fills a separate runtime gap: a concrete, tool-specific workflow for protecting a supported local coding-agent harness with HOL Guard before commands, file edits, MCP/tool calls, or other side effects.

## What

- adds `hol-guard-runtime-safety` with direct HOL Guard install/invocation and fail-closed semantics
- uses `hol-guard detect --json` as the source of truth for the exact supported harness instead of maintaining aliases
- requires Guard-owned bootstrap/install, protected dry-run, launch, and `doctor`/`status` verification
- preserves repository permissions, human approvals, testing, and service-side controls rather than treating Guard as a replacement
- adds the required trigger-routing case and an execution fixture that verifies state-changing work does not proceed when protection cannot be proven

## Validation

Ran against current `main` before submission:

- `node scripts/validate-skills.js` -> 25 skills, 0 errors, 0 warnings
- `node scripts/run-evals.js --min-rank1 80` -> 132 checks passed, 0 errors, 0 warnings; trigger rank-1 88% (71/81)

The behavioral fixture intentionally forbids installing software or running its fake publish command, so it exercises the fail-closed path without creating an external side effect.

### DIFF
--- a/evals/cases/hol-guard-runtime-safety.json
+++ b/evals/cases/hol-guard-runtime-safety.json
@@ -1,0 +1,47 @@
+{
+  "skill_name": "hol-guard-runtime-safety",
+  "trigger": {
+    "positive": [
+      {
+        "prompt": "Protect my coding agent with HOL Guard before it runs commands and edits files",
+        "top_k": 1
+      },
+      {
+        "prompt": "Set up HOL Guard around my local agent harness before this destructive workflow",
+        "top_k": 1
+      },
+      {
+        "prompt": "I need fail-closed HOL Guard runtime protection for agent tool calls and MCP actions",
+        "top_k": 1
+      }
+    ],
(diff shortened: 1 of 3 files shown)


addyosmani:
Thanks for this, but I can't take it. It's a skill built around a specific runtime-safety product (HOL Guard), and the pack stays vendor-neutral, it doesn't ship skills dedicated to one third-party tool. The underlying idea, a fail-closed boundary before side-effecting work, is reasonable, but that belongs in your own agent setup rather than as a pack skill for one product. Closing as out of scope. Appreciate the effort.

PR #521, opened 2026-08-27
feat(performance): deepen backend and database substance

### LINKED ISSUE #378: Enhance performance-optimization: deepen backend/database substance to match its broadened scope
## Context

#358 broadened the `performance-optimization` description to cover "frontend, backend, queries, and databases", which is accurate: the skill body already includes backend material (an N+1 fix, unbounded-fetch pagination, a backend caching example, a Backend measurement section, and a Backend/API branch in the "where to start measuring" tree).

The gap is depth, not presence. The backend coverage is lighter than the frontend / Core Web Vitals coverage, which carries the full `<picture>` art-direction example, code splitting, and React re-render patterns. Now that the description promises full-stack performance, the backend substance is worth bringing closer to that bar, so routing a "slow query" or "N+1" prompt here lands on genuinely useful depth rather than a single snippet.

## Suggested additions (extend the existing skill, not a new one)

Keeping the skill's measure-first ethos and format, add backend/database patterns such as:

- **Query plan analysis:** reading `EXPLAIN` / `EXPLAIN ANALYZE` to spot sequential scans, poor join orders, and row-estimate misses, so a fix targets the proven bottleneck instead of a guessed index.
- **Index strategy:** composite index column order, covering indexes, when an index does not help (low selectivity, leading-wildcard `LIKE`), and the write-cost tradeoff.
- **Connection pooling:** pool sizing, exhaustion symptoms, and the serverless / pgbouncer angle (the existing "connection pool" branch already hints at this).
- **Caching layers** (the richest of these, worth its own depth): in-process vs shared (Redis/Memcached) vs CDN/edge; cache key design and granularity; invalidation strategies (TTL, event or tag-based, versioned keys); stampede / thundering-herd protection (locking, `stale-while-revalidate`, request coalescing); read-through vs write-through vs write-behind; and negative caching. The current single in-memory TTL example is a starting point, not the shape of the final section.

Each as a short, concrete pattern with a measure-first framing and a "when NOT to" note, consistent with the existing anti-pattern sections. The longer material can live in `references/performance-checklist.md` if `SKILL.md` would exceed its budget.

## Non-goal

Not turning this into a database administration guide: keep it to the high-leverage patterns an application engineer actually hits, actionable and minimal, per the repo's skill quality bar. And not a new skill; this is an extension of `performance-optimization`.

nucliweb:
Closes #378. Implements the scope @addyosmani green-lit in that thread.

## What changed

`skills/performance-optimization/SKILL.md` (+85, now 481 lines, within the 500 budget):

- **Queries That Ignore Their Index**: `EXPLAIN ANALYZE` as the measurement rather than "add an index" as the guess; how to read a `Seq Scan`, row-estimate skew, and a stray `Sort` node; composite column order (equality first, then range/sort); and a "when an index will not help" table covering low selectivity, leading wildcards, functions on the column, and write cost.
- **Connection Pool Exhaustion**: the distinctive symptom (every endpoint slow at once, time spent waiting rather than executing, idle sessions on the database), one pool per process, and why raising `max` relocates the queue instead of fixing it.
- **Missing Caching**: expanded from the single in-memory TTL example into layer choice (in-process, shared, edge), key design as a correctness concern, one invalidation strategy rather than an accidental mix, and stampede protection.
- Matching entries in Common Rationalizations, Red Flags, and Verification.

`references/performance-checklist.md` (+83): read/write patterns (cache-aside, read-through, write-through, write-behind), negative caching, request coalescing, and checklists for query plans, index strategy, connection pooling, and caching. Plus seven new anti-pattern rows.

## Scope boundary

Kept to what an application engineer actually hits, not database administration, per the issue's non-goal. Index material is framed diagnostically (reading a plan for an existing query) rather than as schema-time index placement, which keeps it clear of the boundary #492 proposes for `database-and-schema-design`. No dependency on that draft either way.

Deliberately did not move anything under `references/`: #361 and PRs #366/#400 are still settling references portability, so this extends the existing root file in place.

## Verification

- `validate-skills`, `validate-reference-links`, `validate-commands`, `validate-artifact-paths`, `validate-versions`: all PASSED
- `run-evals.js`: 127 checks passed, trigger rank-1 86% (67/78). Identical to the pre-change baseline measured on `main`, confirming no routing drift, as expected since the description is untouched.
- The request-coalescing snippet was executed rather than eyeballed: 100 concurrent waiters produce exactly 1 origin call, the map cleans up on both resolution and rejection, and a rejected fetch does not poison the key.

https://claude.ai/code/session_01TuWNaEV9RFXpTxqeBRTRg8

### DIFF
--- a/references/performance-checklist.md
+++ b/references/performance-checklist.md
@@ -8,6 +8,7 @@ Quick reference checklist for web application performance. Use alongside the `pe
 - [TTFB Diagnosis](#ttfb-diagnosis)
 - [Frontend Checklist](#frontend-checklist)
 - [Backend Checklist](#backend-checklist)
+- [Caching Strategies](#caching-strategies)
 - [Measurement Commands](#measurement-commands)
 - [Common Anti-Patterns](#common-anti-patterns)
 
@@ -92,6 +93,30 @@ When TTFB is slow (> 800ms), check each component in DevTools Network waterfall:
 - [ ] Connection pooling configured
 - [ ] Slow query logging enabled
 
+#### Query plans
+- [ ] `EXPLAIN ANALYZE` captured **before** the fix, not just after — it is the baseline
+- [ ] `Seq Scan` on a large table understood: index missing, unusable, or genuinely not worth it
+- [ ] Estimated vs actual `rows=` within an order of magnitude (if not, refresh statistics before touching indexes)
+- [ ] No `Sort` node that a composite index could absorb
+- [ ] Plan re-checked after the change — an index that did not change the plan gets reverted
+
+#### Index strategy
+- [ ] Composite index column order is equality first, then range/sort
+- [ ] Index covers the query shape (filter + sort), not just one column in isolation
+- [ ] Covering index considered for hot read paths (index-only scan avoids the heap fetch)
+- [ ] Not indexing low-selectivity columns (one value dominating the table)
+- [ ] Expression index used where the query applies a function (`lower(email)`)
+- [ ] Full-text or trigram index used for leading-wildcard search, not a B-tree
+- [ ] Write cost measured on write-heavy tables (every index taxes every `INSERT`/`UPDATE`)
+- [ ] Unused and duplicate indexes dropped (they cost writes and buy nothing)
+
+#### Connection pooling
+- [ ] One pool per process, not per request or per module
+- [ ] `instances × pool max` stays under the database's `max_connections`
+- [ ] `connectionTimeoutMillis` set so exhaustion fails fast instead of queueing forever
+- [ ] Exhaustion diagnosed before resizing: find what holds connections (long transactions, missing `await`, leaked clients)
+- [ ] Serverless / autoscaling fronted by a multiplexing proxy (pgbouncer, RDS Proxy) rather than a larger pool
+
 ### API
 - [ ] Response times < 200ms (p95)
 - [ ] No synchronous heavy computation in request handlers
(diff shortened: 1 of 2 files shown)


addyosmani (approved):
This is a strong addition, Joan, and I read it properly. The backend and database depth is exactly what performance-optimization was thin on: reading the query plan before reaching for an index (Seq Scan, estimate-vs-actual rows, the Sort node above the scan), the when-an-index-wont-help table, connection pool exhaustion (one pool per process, instances times max under max_connections, a proxy for serverless rather than a bigger pool), and the caching section (layer choice, key design keyed on the viewer, one invalidation strategy, stampede protection). It's accurate and it reads in the skill's voice.

Two things I checked and liked: you split the decision material into the SKILL.md and the patterns and checklist into performance-checklist.md, which keeps the skill at 481 lines, under the 500 ceiling, and you used the correct ../../references/ path so the ref-links check stays green. Evals hold at 86%. 'A cache key that omits the viewer is how one user's data gets served to another' is my favorite line, a security correctness bug wearing a performance win, exactly the right thing to flag. Approving. Ready to merge whenever you want it in.

federicobartoli (approved):
This reads well, and the split between the skill and the checklist is exactly what [#378](https://github.com/addyosmani/agent-skills/issues/378) asked for. Nice touch executing the coalescing snippet instead of eyeballing it.

One small thing I stumbled on: the low-selectivity rows (the "when an index will not help" table and the matching checklist item) read as if indexing a skewed column never helps, but that holds for the dominant value. Someone filtering on the rare 5% is exactly who an index serves, typically a partial one. Would a short clause like "for predicates matching the dominant value; a partial index still serves the rare-value query" be worth adding in both spots?

And a tiny one, feel free to ignore: loadOnce could take a generic <T> instead of the undeclared Row, since these snippets tend to get pasted as-is.

Either way, good to go from my side.

nucliweb:
Thanks both.

@federicobartoli, the low-selectivity point is a real correction, not a nit, and I've taken it in `bfc9e32`. As written it implied a skewed column is never worth indexing, which is only true for the dominant value. Scoped in both spots:

- SKILL.md table: "Low selectivity, **querying the dominant value** (a `status` column that is 95% `active`, filtered on `active`) ... Filtering on the rare value is the opposite case, and a partial index serves it well"
- Checklist: "Not indexing low-selectivity columns *for the dominant value*; a partial index still serves the rare-value query (`WHERE status = 'failed'`)"

Took the `<T>` suggestion too, and it turned out to be worth more than a tidy-up. Making it generic exposed that `Map<string, Promise<unknown>>.get()` does not return `Promise<T>`, so the naive version would not have compiled for anyone pasting it. It now needs the explicit `as Promise<T> | undefined`. Verified rather than assumed: `tsc --strict` passes, including generic inference at two call sites with different `T`, and I re-ran the concurrency test on the new shape (100 waiters, 1 origin call, map cleaned on both resolution and rejection).

Validators and evals still green, 86% rank-1, and SKILL.md is unchanged at 481 lines.

PR #524, opened 2026-08-28
Refactor and optimize sdd-cache and simplify-ignore hooks

gotaku3737-sketch:


addyosmani (commented):
The refactor is clean and I verified it preserves behavior: both hook suites pass against it (simplify-ignore 21/21, sdd-cache-post 9/9). Two things before it lands though. First, the tests still aren't wired into CI (nothing runs the hook tests in any workflow), so the coverage you're adding wouldn't actually run, same gap as #501 and #510. Second, this overlaps both of those: #501 adds sdd-cache-test.sh and #510 touches session-start, so three PRs are refactoring and testing the same hooks. Let's coordinate them (whoever lands first, the others rebase, and reconcile the sdd-cache-test.sh vs sdd-cache-post-test.sh naming) and wire the tests into the workflow in the same pass. One question: what does 'optimize' buy here concretely? Refactoring working hooks is worth it for the shared -lib.sh and the tests, but I want the win stated so it isn't churn on load-bearing code.

addyosmani (changes requested):
I can't take this as-is. It commits seven patch_*.js files to the repo root that shouldn't be in the tree, and it's a broad refactor of shared hook and script files in one PR, which CLAUDE.md specifically asks us to avoid since those wide files collide with in-flight work, and there's a lot of that in the hook area right now. If there's a real improvement here, please drop the stray files and split it into small focused PRs, one behavior at a time, and I'll review those. Note #578 is fixing a data-loss bug in simplify-ignore.sh in parallel.

PR #123, opened 2026-05-04
feat: agent hardening (intent-capture + model-routing + anti-hardcoding) + KB foundations

jota-batuta:
## Summary

Multi-slice branch accumulated since PR #40 (notion deprecation). Three coherent themes:

**Agent hardening (this session, 2026-05-04)** — three failure modes resolved:
- `rules/core/no-hardcoded-magic.md` — invariant against literals that should be parameters; anti-patterns derived from a real client pipeline regression.
- `rules/core/model-routing.md` — delegation as default. Main Opus orchestrates; client-code implementation routes to `implementer-haiku`/`implementer`. Lookups route to `Explore`/`general-purpose` (Sonnet).
- `skills/intent-capture/` — universal pre-execution gate (grill -> confirm -> route -> execute). JSON Schema 2020-12 with PII pattern guard on `operator_id`. Grilling pattern attribution to [`mattpocock/skills/grill-with-docs`](https://github.com/mattpocock/skills/blob/main/skills/engineering/grill-with-docs/SKILL.md).
- `agents/code-reviewer.md` — anti-hardcoding check added to Readability dimension.
- `tests/v2.5-validators/15-new-rules-shape.sh` — §A.4 structural validator integrated into run.sh.
- `user-settings/CLAUDE.md` synced with live and given three new behavior blocks; byte-identical with `~/.claude/CLAUDE.md` post-merge.

**Skill promotions (prior session)** — `kb-end-session`, `save-plan`, `batuta-status` promoted to global skills.

**KB pipeline + docs (prior session)** — wikilink convention, connected graph, README/SPEC v3.6 rewrite, session journals.

Implementation method for the agent-hardening slice: built by an Agent Team of 5 Sonnet/Haiku subagents in parallel, audited sequentially by `test-engineer`, `code-reviewer`, and `security-auditor` (all Sonnet). One blocker-fixer Haiku resolved 3 review findings before merge. Main Opus orchestrated only.

## Test plan

- [x] `bash tests/v2.5-validators/run.sh` -- 14/15 PASS. The 1 fail (`08-code-graph-helpers-behavior.sh`, GRAPHIFY_PIN) is pre-existing on main, not introduced by this branch.
- [x] `python -m jsonschema` validation against `intent-schema.json` -- clean payloads pass; payload with `@` in `operator_id` correctly rejected by the new PII pattern guard.
- [x] `diff -q ~/.claude/CLAUDE.md user-settings/CLAUDE.md` -- byte-identical (21421 bytes).
[...]

PR #525, opened 2026-08-28
fix(performance): document the missing guard step

### LINKED ISSUE #522: Enhance performance-optimization: the workflow declares a GUARD step that the skill never documents
## The gap

`performance-optimization` opens with a five-step workflow:

```
1. MEASURE  → Establish baseline with real data
2. IDENTIFY → Find the actual bottleneck (not assumed)
3. FIX      → Address the specific bottleneck
4. VERIFY   → Measure again; keep or revert
5. GUARD    → Add monitoring or tests to prevent regression
```

The body documents Step 1, Step 2, Step 3, and Step 4 as named sections. **There is no Step 5 section.** The workflow jumps from "Step 4: Verify (Keep or Revert)" to a `## Performance Budget` heading that covers part of the ground (a budget list, `lhci autorun`, `bundlesize`) without being framed as the guard step and without covering field monitoring at all.

This is not just a missing heading. The skill already *audits* for guarding in two places while never teaching it:

- Red Flags: "No performance monitoring in production"
- Verification: "Performance budget passes in CI (if configured)"

So an agent following the skill is asked to confirm a practice the Process section never explains, and the "(if configured)" escape hatch means the check passes on any repo that never configured one.

## Why this is the step that matters for UX

Steps 1 to 4 fix one regression once. Step 5 is what stops the same regression returning next sprint, and it is the only step concerned with what real users actually experience. That distinction is the substance worth writing:

- **Synthetic CI** (Lighthouse CI, bundle budgets) catches what the lab can reproduce, on one device profile, one network, a cold cache. It is a gate, and it is cheap.
- **Field data** (RUM, CrUX) catches what the lab cannot: device mix, real network conditions, cache states, and the p75 that Core Web Vitals is actually scored on. A change can pass every synthetic gate and still regress the metric users feel.

A guard step that only mentions CI teaches half the lesson, and the half that misses real-user regressions.

## Proposed scope

Add a `### Step 5: Guard Against Regression` section, and fold the existing `## Performance Budget` content into it so the workflow reads end to end:

- **Choosing what to guard**: budget the metric the user feels (LCP, INP, CLS, p95 latency), not every number available. A budget nobody can act on gets disabled after the third false alarm.
- **The two layers and what each is for**: synthetic gate in CI to block a regression before merge; field monitoring to catch what the gate structurally cannot see.
- **Making the gate survivable**: budgets on a trend rather than a single run, because Lighthouse run-to-run variance produces flaky gates, which is how teams end up deleting the gate instead of the regression. This connects to the noise discipline Step 4 already teaches.
- **Field monitoring**: `web-vitals` attribution build to RUM, alert on p75 movement, and the fact that CrUX is a 28-day window so a regression surfaces slowly.
- **Closing the loop**: a caught regression should send you back to Step 1, not straight to a fix.

Tighten the two audit lines to match, including dropping or qualifying the "(if configured)" in Verification.

## Boundaries

- **Not** `observability-and-instrumentation`: that skill explicitly defers here ("Profiling and optimizing measured slowness, use the `performance-optimization` skill") and covers RED/USE for services, not Core Web Vitals field data.
- **Not** `ci-cd-and-automation`: verified it contains no mention of Lighthouse, performance budgets, or regression gates, so nothing is being duplicated.
- Keeps the material inside the existing skill rather than proposing a new one, per CONTRIBUTING.

## Non-goal

Not a monitoring-vendor guide, and not a second observability skill. The scope is the guard step the workflow already promises.

Happy to implement this if the direction looks right. Note it touches the same file as #521, so I would sequence the implementation after that lands.

CBOSSX:
## Summary

Closes #522. The performance workflow declared a fifth `GUARD` step but stopped documenting the process after Step 4, so an agent could finish without establishing regression protection. This adds the missing Step 5 for synthetic CI and field monitoring, strengthens the behavioral eval, and makes the validator reject any named numbered workflow whose declared step lacks a process section.

The structural regression test failed before the fix (`7/8`) and passes after it (`8/8`).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## Test plan

- [x] `node --test scripts/lib/skill-lint-test.js`
- [x] `node --test scripts/*-test.js scripts/lib/*-test.js` — 43/43 passed
- [x] `node scripts/run-evals.js --min-rank1 80` — 128 checks passed; rank-1 86%
- [x] `node scripts/validate-skills.js` — 24 skills, 0 errors/warnings
- [x] `node scripts/validate-versions.js && node scripts/validate-reference-links.js && node scripts/validate-commands.js && node scripts/validate-artifact-paths.js`
- [x] `claude plugin validate .`
- [x] `git diff --cached --check`

## Test coverage

- [x] I ran tests for the surface(s) I changed locally.
- [x] New code paths are covered by tests in this PR (no bare additions).
- [x] If I removed code, I updated `coverage-baseline.json` in this PR only if the removal caused a legitimate regression and I called it out in the summary above.
- [ ] The coverage gate check is green in CI before requesting review.

## Checklist

- [x] I have read [CONTRIBUTING.md](CONTRIBUTING.md) (if present) and [docs/DEVELOPMENT.md](docs/DEVELOPMENT.md).
- [ ] Commits are signed and follow conventional-commits style.
- [x] I have linked any related issues (Fixes #123 / Closes #456).

## Related issues / PRs

- Closes #522

### DIFF
--- a/evals/cases/performance-optimization.json
+++ b/evals/cases/performance-optimization.json
@@ -38,14 +38,15 @@
     {
       "id": 1,
       "prompt": "The products page renders slowly with 1000 items. Improve its performance.",
-      "expected_output": "A measured baseline, an identified bottleneck, a targeted fix, and a verified improvement",
+      "expected_output": "A measured baseline, an identified bottleneck, a targeted fix, a verified improvement, and a regression guard",
       "files": [
         "performance-optimization"
       ],
       "expectations": [
         "Performance is measured before any optimization is applied",
         "The fix targets the measured bottleneck rather than guessing",
-        "Improvement is verified against the baseline after the change"
+        "Improvement is verified against the baseline after the change",
+        "A concrete guard is added after verification using a synthetic performance budget or field monitoring tied to the measured user-facing metric"
       ]
     }
   ]
--- a/scripts/lib/skill-lint-test.js
+++ b/scripts/lib/skill-lint-test.js
@@ -109,6 +109,31 @@ test('reports frontmatter name that disagrees with the directory', () => {
   assert.match(errors[0], /does not match directory name/);
 });
 
+test('reports a workflow step declared without a matching process section', () => {
+  const content = withAllSections(VALID_FRONTMATTER).replace(
+    '## Common Rationalizations',
+    [
+      '## The Optimization Workflow',
+      '',
+      '```',
+      '1. MEASURE → Establish a baseline',
+      '2. GUARD   → Prevent regression',
+      '```',
+      '',
+      '### Step 1: Measure',
+      '',
+      'Measure first.',
+      '',
+      '## Common Rationalizations',
+    ].join('\n'),
+  );
+
+  const { errors } = lintSkillContent('alpha', content, KNOWN);
+
+  assert.equal(errors.length, 1);
+  assert.match(errors[0], /Workflow declares Step 2 but has no matching process section/);
+});
+
 test('reports a missing frontmatter block', () => {
   const { errors } = lintSkillContent('alpha', '## Overview\nx\n', KNOWN);
   assert.equal(errors.length, 1);
--- a/scripts/lib/skill-lint.js
+++ b/scripts/lib/skill-lint.js
@@ -209,6 +209,27 @@ function lintSkillContent(dirName, content, knownSkills) {
     }
   }
 
+  // A named workflow that advertises numbered steps must document each step
+  // before the next level-two section. Otherwise the summary promises a
+  // process stage that the skill never teaches agents how to perform.
+  const workflowSections = content.matchAll(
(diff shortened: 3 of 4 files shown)


addyosmani (approved):
Nice, this fixes the instance and the class. The perf skill declared Step 5 in its workflow but never documented it (the Performance Budget section sat where the Guard step should have been), and your new skill-lint rule catches exactly that: a workflow that advertises numbered steps but leaves one undocumented. I verified the rule doesn't false-positive on the rest of the catalog (24 skills, 0 errors), and lint tests and evals pass. Approving.

Issue #518, opened 2026-08-26
[skill-gap]: tasks/ are Poorly designed in parallel development


kouhe3:
### Affected skill

planning-and-task-breakdown

### Relevant instruction or excerpt

**Do NOT write code during planning.** The output is a plan document saved to `tasks/plan.md` and a task list recorded in the task list target (see Output Files; default `tasks/to
do.md`), not implementation.

### Project context

I use this skill in Game, Lib project.

### What did not work or did not apply

My agent delete old plan.md and write new plan.md for new task and todo.md
Even if the old plan.md is not completed

### What you did instead

I have written multiple plan.md files with different names under the tasks/ directory

freyjasasi:
For this I added a readme inside tasks which moves to archive the old one with spec name plan and spec name task. So it will be there always. But before complete no idea

addyosmani:
This is a real one, and it's the same area as #509 and #513. A flat tasks/todo.md doesn't express parallelism or dependencies well once work fans out. I'm being deliberate about changing the spec and tasks convention though, since it touches the whole spec/plan/build flow and the validators, so I don't want to bolt on a parallel-tasks format piecemeal. Keeping this open as input into that decision. What does your parallel workflow actually need from the task list: dependency edges, per-lane grouping, both?

freyjasasi:
I don't think you need to touch those. Even though I am not using parallel once all plan and todo competed may be it will change it to spec name plan and spec name todo then move to archive folder inside tasks itself. This is after all build completed but before that of anything is running parallel two diff plan from two diff specs not sure your call. But I feel plan and todo must have spec name plan and spec name todo and build can read what ever open in tasks not in archive can be more meaningful as per me

cmondragon023:
For the parallel workflow described here, I think both concepts are useful at different layers: per-lane grouping keeps the plan readable for humans and agents, while explicit dependency edges give a scheduler enough information to decide what can run concurrently.

A backward-compatible convention could keep the current flat `tasks/plan.md` and `tasks/todo.md` for single-lane work, then use a spec-scoped directory plus a small root index with `depends_on` only when work fans out. I would leave automatic archiving out of the first change because lifecycle policy is separate from preventing collisions and representing dependencies.

I'm happy to prototype this once you settle the convention you want contributors to follow.

ayobamiseun:
[...]

PR #526, opened 2026-08-28
fix(planning): never overwrite an incomplete plan without asking (#518)

ayobamiseun:
The convention-neutral half of #518, as offered in [this comment](https://github.com/addyosmani/agent-skills/issues/518#issuecomment-5454672750). The reported failure has two parts: (1) the agent silently deletes an incomplete `tasks/plan.md`/`tasks/todo.md` when planning new work — data loss; (2) the tasks layout can't represent parallel lanes — the convention decision being deliberately held open. This PR fixes only (1) and forecloses nothing about (2): whatever layout lands, "don't destroy an in-flight plan's state" holds.

## Changes

**`skills/planning-and-task-breakdown/SKILL.md`**
- New guard in Output Files: before writing, check for existing files with unchecked tasks. Replanning the same work → update in place. Different work → stop and ask; never delete, overwrite, or rename on your own. Mirrored for external task list targets (no bulk-closing another plan's open tracker items).
- Matching Rationalization row ("the old plan is stale, I'll just replace it"), Red Flag, and Verification checklist item.

**`.claude/commands/plan.md`**
- One guard line after the save instruction, so the command carries the rule even when invoked without the full skill context.

## Verification

- `validate-skills`, `validate-commands`, `validate-artifact-paths` all pass (no artifact paths moved — the guard references only the canonical `tasks/plan.md`/`tasks/todo.md`).
- `run-evals`: 127 checks pass, trigger rank-1 rate unchanged at 86%.

Refs #518 (fixes the data-loss report; the parallel-layout decision stays open there).

### DIFF
--- a/.claude/commands/plan.md
+++ b/.claude/commands/plan.md
@@ -14,3 +14,5 @@ Read the existing spec (SPEC.md or equivalent) and the relevant codebase section
 6. Present the plan for human review
 
 Save the plan to tasks/plan.md and task list to tasks/todo.md.
+
+If tasks/plan.md or tasks/todo.md already exists with unchecked tasks for different work, stop and ask before writing — never silently overwrite an incomplete plan.
--- a/skills/planning-and-task-breakdown/SKILL.md
+++ b/skills/planning-and-task-breakdown/SKILL.md
@@ -147,6 +147,13 @@ If a task is L or larger, it should be broken into smaller tasks. An agent perfo
 
 Create the `tasks/` directory if it does not exist.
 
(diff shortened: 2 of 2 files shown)


mdfaysal-hossain (commented):
Hi

addyosmani (approved):
Good catch, and a real footgun. Silently overwriting a tasks/plan.md that still has unchecked tasks can destroy work state that exists nowhere else, and 'different work, stop and ask' is the right call. Small, safe, verified clean and green. Approving.

PR #531, opened 2026-08-29
fix(planning): never overwrite an incomplete plan without asking

### LINKED ISSUE #518: [skill-gap]: tasks/ are Poorly designed in parallel development

ayobamiseun:
## Summary

- guard planning outputs against silently overwriting unchecked work
- allow same-work replanning while requiring a user decision for different work
- apply the protection to external task trackers and all three command integrations

Closes #518.

## Verification

- node scripts/validate-commands.js
- node scripts/run-evals.js --min-rank1 80
- git diff --check

All relevant checks pass: 24 skills, 8 command families, and 127 routing checks; trigger rank-1 rate remains 86%.

### DIFF
--- a/.gemini/commands/planning.toml
+++ b/.gemini/commands/planning.toml
@@ -13,4 +13,6 @@ Read the existing spec (SPEC.md or equivalent) and the relevant codebase section
 6. Present the plan for human review
 
 Save the plan to tasks/plan.md and task list to tasks/todo.md.
+
+If tasks/plan.md or tasks/todo.md already exists with unchecked tasks for different work, stop and ask before writing — never silently overwrite an incomplete plan.
 """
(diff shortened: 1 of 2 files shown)


addyosmani (approved):
Good, this completes #526 across the other command dirs: #526 fixed the Claude command and the skill, and this applies the same 'never overwrite an incomplete plan' guard to the Gemini and generic planning.toml, keeping the command surfaces in sync. Verified clean and green. Approving.

federicobartoli (approved):
Checked against main: the wording matches the guard already in .claude/commands/plan.md line for line, so the three command surfaces are back in sync, and it merges clean. Approving.

PR #532, opened 2026-08-29
docs(antigravity): document the command-wrapper discoverability limitation (#445)

ayobamiseun:
Follow-up to #445 (the install-path half landed in #481; this documents the remaining command-discoverability half).

## What changed and why

In affected `agy` 1.1.x releases, Antigravity's migration tooling reports the 8 legacy definitions in `commands/*.toml` as "converted to skills," and `agy plugin validate` passes — but the converted wrappers never appear in the slash-command or skill catalog. So a passing validate confirms the files are well formed, **not** that `/build` and the other short wrappers are available, which is exactly the confusion reported in #445.

**`docs/antigravity-setup.md`**
- Replaces the slash-command table (which documented commands affected users can't invoke) with a **Lifecycle Workflows and Command Compatibility** section: a wrapper → direct-invocation map using the namespaced skills that DO work (`/agent-skills:spec-driven-development`, …, and `/agents` → `web-performance-auditor` for `/webperf`, which is a persona rather than a skill).
- Notes the wrapper-only behaviors that are unavailable while the importer limitation applies: `/build auto` orchestration and `/ship`'s automatic persona fan-out.
- Warns against the YAML-frontmatter workaround: `---` frontmatter makes the parallel TOML command files invalid for Gemini CLI's strict parser without changing Antigravity's conversion behavior.
- Adds a post-install verification step (fresh session, type `/agent-skills:` to inspect the catalog) and states that reinstalling or adding a TOML `name` field does not resolve it.
- Links the upstream tracker (google-antigravity/antigravity-cli#788) alongside #445.

**`README.md`**
- Aligns the Antigravity blurb: "skills and subagents" (not "slash commands"), with a deep link to the new compatibility section.

## Verification

- `node scripts/validate-skills.js`, `validate-commands.js`, `validate-artifact-paths.js`, `validate-reference-links.js` all pass (docs-only change).
- Wrapper→skill mapping cross-checked against `commands/*.toml` and the skill catalog; the `/planning` naming note (Antigravity's built-in `/planning` is a separate plan-mode control) is preserved in the table.

Refs #445.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ Put workflow skills under `.cursor/skills/` (sync from `agent-skills/skills/`) a
 <details>
 <summary><b>Antigravity CLI</b></summary>
 
-Install as a native plugin for skills, subagents, and slash commands. See [docs/antigravity-setup.md](docs/antigravity-setup.md).
+Install as a native plugin for skills and subagents. In affected Antigravity CLI releases, legacy command TOMLs are reported as converted but their wrapper commands are not discoverable; invoke the underlying namespaced skills directly. See [docs/antigravity-setup.md](docs/antigravity-setup.md#lifecycle-workflows-and-command-compatibility).
 
 **Install from the repo:**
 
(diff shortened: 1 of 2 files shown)


addyosmani (approved):
Approving. Documents the command-wrapper difference for Antigravity clearly, and it lines up with the install-path and frontmatter work in #481, #489, and #445. Docs-only, clean.

PR #523, opened 2026-08-27
feat: add skill-sec — agent instructions supply-chain security auditor

sudo-ai-git:
## What this adds
**skill-sec** — a supply-chain auditor for *downloaded agent instructions*. When an agent is about to run a SKILL.md / system-prompt it fetched from the internet, this skill audits it against the **eight malicious-agent patterns** — prompt injection, data exfiltration, hardcoded secrets, dangerous commands, obfuscation, untrusted fetches, credential access, privilege escalation — and cites the offending line for each.

## Why it is a real gap (not a duplicate)
 covers hardening *code the agent writes* (input validation, auth, storage). This skill covers the orthogonal surface: **gating untrusted instructions the agent is about to execute** — the skill-supply-chain problem (marketplace/GitHub downloads). The two are complementary; this one does not overlap 's content and there is no existing skill for auditing downloaded SKILL.md/prompt files.

## Compliance
All five repo validators pass locally:
- validate-skills.js: 25 skills, 0 errors
- validate-artifact-paths.js: 0 errors
- validate-commands.js: 0 errors
- validate-reference-links.js: 0 errors
- run-evals.js: 135 checks, 0 errors, 0 warnings (trigger rank-1 rate 84%, matching baseline)
Includes a CI-compliant eval case () + execution fixture (), 4 positive + 3 negative triggers routed to existing skills, and a behavioral execution eval with a deliberately-malicious fixture the agent must flag.

## Integrity note
The skill is intentionally free of any proprietary method — pure public-supply-chain-security practice, tested end-to-end.

PR #536, opened 2026-08-31
docs(security): rate limiting must count in a shared store across instances

CahidArda:
## What

`## Rate Limiting` in `security-and-hardening` is a single `express-rate-limit` snippet with no prose. `express-rate-limit` counts in process memory by default, so the snippet as written enforces `max × instances` behind a load balancer and effectively nothing on serverless/edge, where each invocation can start from zero. The auth limiter (10 attempts / 15 min) is exactly the control that brute-force attackers spread across instances.

This adds one paragraph naming the failure mode and the fix (a shared `store` — Redis via `rate-limit-redis` — or an HTTP-based limiter for runtimes without a long-lived TCP connection), a short `@upstash/ratelimit` example for the auth endpoint, and matching one-line updates to the existing Red Flags and Verification entries.

- The skill already owns rate limiting: STRIDE row "Denial of service → Rate limiting", the "Modifying rate limiting or throttling" Ask-First rule, checklist item "Login has rate limiting", Red Flag "No rate limiting on authentication endpoints", Verification "Rate limiting active on auth endpoints". None of them say the counter has to be shared, so an agent following the skill can add a limiter that passes every checkbox and does not work in production.
- `references/security-checklist.md` L41 has the same gap ("Rate limiting on login endpoint (≤10 attempts per 15 minutes)").
- Focused edit per CONTRIBUTING "Modifying Existing Skills"; no new directory, no frontmatter change, no routing surface added.

## Disclosure

I work at Upstash, which makes `@upstash/ratelimit`. The paragraph leads with the vendor-neutral fix (`store` + `rate-limit-redis`) and the example is there because an HTTP-based limiter is the only option on edge runtimes; if you would rather keep the file free of a named product, I'm happy to drop the second code block and keep the prose, or swap the example for `rate-limit-redis`.

- Searched the catalog: no other skill covers rate limiting; `performance-optimization` covers caching layers but not limiters.
- Checked open PRs: none touch `skills/security-and-hardening/SKILL.md` or rate limiting (#523 proposes a separate supply-chain auditor skill).
- Extends rather than adds.

## Scope and size

One file. Frontmatter description unchanged. The file was at 499 lines and this takes it to 513 (or 508 if the second express block is folded into the shared-store example, which I can do on request). The 500-line note in skill-anatomy.md is a guideline rather than a CI rule, but flagging it so you can decide.

## Verification

- `node scripts/validate-skills.js` → 25 skills checked, 0 errors, 0 warnings
- `node scripts/run-evals.js --min-rank1 80` → 136 checks passed, rank-1 86% (72/84) — unchanged from `main` (description untouched)
- `node scripts/validate-reference-links.js` → 25 skills checked, 0 errors
- `git diff --check` → clean
- Snippet checked against `@upstash/ratelimit` 2.0.8 / `@upstash/redis` 1.38.3 (`Redis.fromEnv()`, `Ratelimit.slidingWindow(10, '15 m')`, `limit()` → `{ success }`).

### DIFF
--- a/skills/security-and-hardening/SKILL.md
+++ b/skills/security-and-hardening/SKILL.md
@@ -329,6 +329,20 @@ app.use('/api/auth/', rateLimit({
 }));
 ```
 
+**Count in a shared store once there is more than one process.** `express-rate-limit` keeps its counters in process memory by default. Behind a load balancer each instance holds its own count, so the effective limit is `max × instances`; on serverless or edge runtimes a fresh invocation starts from zero, so the auth limit above may never fire. Pass a shared `store` (Redis via `rate-limit-redis`), or use an HTTP-based limiter that works where a long-lived TCP connection does not (for example `@upstash/ratelimit`):
+
+```typescript
+import { Ratelimit } from '@upstash/ratelimit';
+import { Redis } from '@upstash/redis';
+
+const authLimiter = new Ratelimit({
+  redis: Redis.fromEnv(),                       // UPSTASH_REDIS_REST_URL + _TOKEN
+  limiter: Ratelimit.slidingWindow(10, '15 m'), // 10 attempts per 15 minutes, across all instances
+});
+const { success } = await authLimiter.limit(`login:${req.ip}`);
+if (!success) return res.status(429).end();
+```
+
 ## Secrets Management
 
 ```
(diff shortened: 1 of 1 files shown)


addyosmani (approved):
Correct and worth stating. A per-instance rate-limit counter silently fails the moment you run more than one instance, so 'count in a shared store for distributed setups' is the right addition. Small, verified clean. Approving.

PR #434, opened 2026-07-30
fix(skills): close description-vocabulary gaps, raise trigger rank-1 floor to 95

ayobamiseun:
## Summary

Follow-up to #351 (closed): the Tier-2 trigger evals passed, but 11 of 76 positive prompts still ranked their skill #2–#3 instead of #1, and those stragglers were no longer tracked anywhere. Per the eval README's own guidance ("a Tier-2 failure usually means fix the description"), this PR enriches ten skill descriptions with the vocabulary users actually say, then locks in the gain by raising the CI ratchet.

**Trigger rank-1 rate: 86% → 100% (76/76). Zero new collision warnings.**

## What changed

One-line description edits (keeping the what-then-"Use when…" format), each targeting a real vocabulary gap surfaced by the evals:

| Skill | Missing vocabulary |
|---|---|
| debugging-and-error-recovery | regression asks: "worked yesterday", "figure out what broke" |
| documentation-and-adrs | "document an architecture decision" — the light stemmer doesn't cluster *documentation*/*document* |
| doubt-driven-development | cross-examining assumptions; stress-testing a plan for hidden failure modes; high-stakes changes |
| git-workflow-and-versioning | splitting a messy working tree into clean atomic commits |
| incremental-implementation | thin/verifiable slices; the next task from a plan; feature flags |
| shipping-and-launch | "what needs to be in place before shipping" |
| spec-driven-development | PRD, objectives and scope |
| test-driven-development | red-green-refactor (the skill's own signature term) |
| using-agent-skills | deciding which workflow applies to the work at hand |
| source-driven-development | "verify against the official docs" — *docs*/*documentation* also don't stem together |

The source-driven-development edit repairs the one ranking my incremental-implementation edit would otherwise have shifted (its prompt contains "Next.js" + "implementing"), caught by re-running the evals after each edit.

## Ratchet

Per the README policy ("raise the floor as routing improves; never lower it"):
- CI `--min-rank1` raised 80 → 95 in `.github/workflows/test-plugin-install.yml` — same ~4-prompt headroom below the new 100% baseline that 80 gave the old 86% baseline.
- `evals/README.md` baseline text updated; the stale "tracked in #351" sentence (issue is closed) rephrased as standing guidance: fix the description, not the prompt.

## Verification

- `node scripts/run-evals.js --min-rank1 95` → 124 checks passed, 0 errors, 0 warnings, rank-1 100%
- `node --test scripts/run-evals-test.js` → pass
[...]

PR #537, opened 2026-08-31
fix(skills): add database-migration and dependency-audit vocabulary to two skill descriptions

notbucki:
## Summary

Two skills carry substantial body content whose vocabulary never made it into their frontmatter `description`, so realistic prompts for exactly that content cannot reach them in the Tier-2 trigger evals (same failure class as #351, and the same shape of fix as #508):

- **`deprecation-and-migration`**: the *Database Schema Migrations (Expand/Contract)* section (added in #345) is a core part of the skill — "never rename in place" is one of its own rationalization rebuttals — but the description says nothing about databases, schemas, or columns.
- **`security-and-hardening`**: the skill owns *Triaging Dependency Audit Results* and *Supply-Chain Hygiene* (hardened in #392), and `code-review-and-quality` explicitly delegates to it ("For triaging `npm audit` findings and supply-chain risk … follow the `security-and-hardening` skill — this section covers the upgrade *workflow*, that one covers the security verdict"), but the description mentions neither dependencies, audits, nor supply chain.

Per the eval README ("a Tier-2 failure usually means *fix the description*"), this PR adds the missing vocabulary to both descriptions — keeping the what-then-"Use when…" format — and adds four positive trigger cases so the suite can see these gaps from now on.

## Evidence

With the four trigger cases added and the descriptions unchanged, `node scripts/run-evals.js` reports all four as hard misses, not merely low-ranked:

```
✗  deprecation-and-migration: description shares no vocabulary with a prompt users would say
     "Rename a database column without downtime"
✗  deprecation-and-migration: description shares no vocabulary with a prompt users would say
     "How do I drop a column safely in production?"
✗  security-and-hardening: description shares no vocabulary with a prompt users would say
     "Run a dependency audit and triage the findings"
✗  security-and-hardening: description shares no vocabulary with a prompt users would say
     "Is this new package safe to install? Check for typosquatting and supply chain risk"
```

Without a case to catch it, "Rename a database column without downtime" currently ranks `code-simplification` #1 and does not rank `deprecation-and-migration` at all.

| | rank-1 rate | checks |
|---|---|---|
| baseline (unmodified) | 86% (72/84) | 136 passed, 0 errors |
| new trigger cases only | 82% (72/88) | 4 errors (above) |
| cases + description fixes | 86% (76/88) | 140 passed, 0 errors, 0 warnings |

After the fix all four new prompts rank their skill #1. The +4 rank-1 gain is exactly the four new prompts — no pre-existing prompt lost rank 1 — and there are no new collision warnings. `scripts/validate-skills.js` also passes (25 skills, 0 errors).

One wording choice was deliberate: the deprecation clause says "migrating a database schema" (reusing the already-present *migrat* token) rather than "changing a database schema", because the latter shifted the IDF of *chang* and pushed an existing `incremental-implementation` prompt ("Ship this change behind a feature flag in thin slices") from top-3 to #4 — the edit-coupling effect #434 describes. The final wording leaves every pre-existing check green.

## What changed

- `skills/deprecation-and-migration/SKILL.md` — one added clause: *"Use when migrating a database schema in production, such as renaming or dropping a column without downtime (expand/contract)."*
- `skills/security-and-hardening/SKILL.md` — one added clause: *"Use when auditing dependencies for known vulnerabilities, triaging package-manager audit findings, or assessing supply-chain risk in a new package."*
- `evals/cases/deprecation-and-migration.json`, `evals/cases/security-and-hardening.json` — the four positive trigger cases quoted above, `top_k: 3`.

## Relationship to open work

- #434 closes vocabulary gaps in ten *other* skills; the two here are disjoint from that set, and this PR does not touch the CI floor or the README baseline.
- #492 (on hold for scoping) would move the expand/contract depth into a new `database-and-schema-design` skill. If that lands, the schema clause and its two trigger cases should travel with the content; at current HEAD, `deprecation-and-migration` owns this material and is unreachable by its vocabulary.
- The security clause stays on the security-verdict side of the boundary drawn when #399 was closed: upgrade *workflow* belongs to `code-review-and-quality`; audit triage and the supply-chain verdict belong to `security-and-hardening`.

---

This contribution was researched and drafted with Claude Code (an AI coding agent); a human reviewed the findings and approved the submission.

### DIFF
--- a/evals/cases/deprecation-and-migration.json
+++ b/evals/cases/deprecation-and-migration.json
@@ -13,6 +13,14 @@
       {
         "prompt": "How do I remove this old feature safely for existing users?",
         "top_k": 3
+      },
+      {
+        "prompt": "Rename a database column without downtime",
+        "top_k": 3
+      },
+      {
+        "prompt": "How do I drop a column safely in production?",
+        "top_k": 3
       }
     ],
     "negative": [
--- a/evals/cases/security-and-hardening.json
+++ b/evals/cases/security-and-hardening.json
@@ -13,6 +13,14 @@
       {
         "prompt": "Harden the webhook endpoint that fetches user-supplied URLs",
         "top_k": 3
+      },
+      {
+        "prompt": "Run a dependency audit and triage the findings",
+        "top_k": 3
+      },
+      {
+        "prompt": "Is this new package safe to install? Check for typosquatting and supply chain risk",
+        "top_k": 3
       }
     ],
     "negative": [
--- a/skills/deprecation-and-migration/SKILL.md
+++ b/skills/deprecation-and-migration/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: deprecation-and-migration
-description: Manages deprecation and migration. Use when removing old systems, APIs, or features. Use when migrating users from one implementation to another. Use when deciding whether to maintain or sunset existing code.
+description: Manages deprecation and migration. Use when removing old systems, APIs, or features. Use when migrating users from one implementation to another. Use when migrating a database schema in production, such as renaming or dropping a column without downtime (expand/contract). Use when deciding whether to maintain or sunset existing code.
 ---
 
 # Deprecation and Migration
--- a/skills/security-and-hardening/SKILL.md
+++ b/skills/security-and-hardening/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: security-and-hardening
-description: Hardens code against vulnerabilities. Use when handling user input, authentication, data storage, or external integrations. Use when building any feature that accepts untrusted data, manages user sessions, or interacts with third-party services. Use when personal data or privacy compliance (GDPR, CCPA) is involved.
+description: Hardens code against vulnerabilities. Use when handling user input, authentication, data storage, or external integrations. Use when building any feature that accepts untrusted data, manages user sessions, or interacts with third-party services. Use when auditing dependencies for known vulnerabilities, triaging package-manager audit findings, or assessing supply-chain risk in a new package. Use when personal data or privacy compliance (GDPR, CCPA) is involved.
 ---
 
 # Security and Hardening


nucliweb (approved):
Verified at runtime on `1483fc1`, not just by reading the diff. Both description changes advertise capability the skills already have, which is exactly the Tier 2 false-negative case the evals README describes, so this looks correct to me.

## Runtime check

| | positive prompts | rank-1 | warnings |
|---|---|---|---|
| `upstream/main` | 84 | 86% (72/84) | 0 |
| this PR | 88 | 86% (76/88) | 0 |

All four new prompts rank their skill first, and the near-collision check (cosine > 0.5 between descriptions) stays clean, so the added vocabulary does not encroach on a neighbouring skill. Top-3 with scores:

```
Rename a database column without downtime
  1. deprecation-and-migration   0.3291
  2. code-simplification         0.0516

How do I drop a column safely in production?
  1. deprecation-and-migration   0.1655
  2. observability-and-instrum.  0.0965

Run a dependency audit and triage the findings
  1. security-and-hardening      0.2950
  2. observability-and-instrum.  0.0609

Is this new package safe to install? Check for typosquatting and supply chain risk
  1. security-and-hardening      0.2125
  2. constraint-driven-devel.    0.0679
```

## Why this is a description fix and not scope creep

I checked whether the descriptions are claiming territory the skills do not actually cover. They are not:

- `deprecation-and-migration/SKILL.md` already carries a full **"Database Schema Migrations (Expand/Contract)"** section (lines 164 to 191), including the phased add/backfill/dual-write/contract sequence, two rationalizations about renaming a column in place, and verification items on schema/code deploy ordering.
- `security-and-hardening/SKILL.md` already carries **"Triaging Dependency Audit Results"** (line 272) and **"Supply-Chain Hygiene"** (line 299), and line 306 already discusses typosquatted packages specifically.

So in both cases the guidance existed and the description simply did not carry the words a user would say to reach it.

## One forward-looking note, not a change request for this PR

[#492](https://github.com/addyosmani/agent-skills/pull/492) (draft) proposes a `database-and-schema-design` skill whose summary lists **expand/contract migrations** among its contents while declaring "clear boundaries with `deprecation-and-migration` (system/API sunset)". That boundary does not match the current state of the tree: expand/contract is already taught in depth in `deprecation-and-migration`, and this PR only makes that discoverable.

I do not think that is this PR's problem to solve, and I would not hold this on it. Flagging it so whoever picks up #492 resolves the overlap deliberately rather than ending up with two descriptions competing for "rename a column without downtime". Worth a rank-1 re-measurement at that point.

The `security-and-hardening` half has no equivalent tension. [#523](https://github.com/addyosmani/agent-skills/pull/523) is about auditing untrusted SKILL.md files the agent is about to execute, which is orthogonal to auditing project dependencies.

Nice, tightly scoped change.

addyosmani (approved):
Good, and the right shape: extending deprecation-and-migration with database-migration vocabulary and security-and-hardening with dependency vocabulary, rather than adding new skills for either. Verified clean, evals hold. Approving.

PR #543, opened 2026-09-02
fix(hooks): run SessionStart without a shell so it works on Windows

### LINKED ISSUE #488: SessionStart hook fails on Windows (bash syntax run through pwsh)

abhisheksharma2411:
Closes #488.

## The failure

`hooks/hooks.json` ran a bash one-liner:

```
SCRIPT="${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh"; [ -f "$SCRIPT" ] || SCRIPT="..."; [ -f "$SCRIPT" ]&& bash "$SCRIPT" || true
```

Per the hooks documentation, the command string is passed to `sh -c` on macOS and Linux, **Git Bash on Windows, or PowerShell when Git Bash isn't installed**. PowerShell cannot parse `[ -f "$SCRIPT" ]`, so @'s report of a `ParserError` on every session start is exactly what that path produces. The script it invoked then needs `bash` and `jq` as well, neither of which is present by default on Windows — so fixing only the syntax would still leave the hook broken.

## The fix

Switch `hooks.json` to the **exec form**. The docs are explicit that with `args` present *"no shell tokenization happens on any platform"* — so there is no shell to disagree about, rather than a shell we branch on:

```json
{
  "type": "command",
  "command": "node",
  "args": ["${CLAUDE_PLUGIN_ROOT}/hooks/session-start.mjs"]
}
```

The script becomes Node. That is not a new dependency — the repo's entire `scripts/` tree, its tests and Claude Code itself already require it — and `JSON.stringify` does the escaping `jq` was doing, so the last external dependency goes with it. Placeholders are still exported to the spawned process, so the plugin-root / project-dir choice the old command made in shell now happens inside the script.

I went this way rather than shipping the `.ps1` the issue suggests: a `.ps1` needs `hooks.json` to branch on OS, and the command string that does the branching is itself parsed by the shell we are trying to avoid. Removing the shell removes the whole class.

## Behaviour is unchanged

The point is portability, not a new payload. The output is **byte-identical** to the bash hook's — which took two rounds to get right, and both are pinned by tests:

- `jq -cn` emits a trailing newline; `process.stdout.write` did not.
- `$(cat …)` strips trailing newlines from the meta-skill; `readFileSync` kept them.

```
$ bash hooks/session-start.sh   > a.json
$ node hooks/session-start.mjs  > b.json
$ diff a.json b.json && echo identical
identical          # 10814 bytes both
```

## Tests

`hooks/session-start-node-test.js`, wired into the `validate-skills` job:

| test | pins |
|---|---|
| emits the SessionStart envelope with the meta-skill body | the happy path |
| payload is byte-identical to the bash hook it replaces | no behaviour change (skips where bash/jq absent — i.e. the very environment #488 is about) |
| reports a notice, not a failure, when the meta-skill is absent | a missing skill must not fail the session |
| `hooks.json` uses exec form so no shell parses the command | **the fix itself** |

Each was mutation-tested; every mutation killed exactly one test:

| mutation | result |
|---|---|
| restore the old shell-form command | 1 fail |
| drop the trailing-newline match | 1 fail |
| exit non-zero when the meta-skill is missing | 1 fail |

Full local run on `d2c37ef`: all five validators OK, `node --test` **32/32**, `run-evals.js --min-rank1 80` passed at 86% (72/84).

## Reviewer notes

- **I cannot test this on Windows.** The reasoning rests on the documented shell-selection and exec-form behaviour plus the byte-identical payload, not on an observed Windows run. If @ (the reporter) can confirm against this branch, that would close the loop properly.
- **`session-start.sh` is deliberately left in place.** It still works invoked directly, and #510 is in flight against its test — I did not want to delete a file out from under an open PR. Once #510 lands I am happy to send a follow-up collapsing the two.
- **Overlap with #510**: it also touches `.github/workflows/test-plugin-install.yml` and `CONTRIBUTING.md`. Mine adds one CI step and edits one paragraph. #510 was opened first, so I will rebase behind it rather than the other way round — say the word.

### DIFF
--- a/.github/workflows/test-plugin-install.yml
+++ b/.github/workflows/test-plugin-install.yml
@@ -40,6 +40,9 @@ jobs:
       - name: Test reference-link validator
         run: node --test scripts/validate-reference-links-test.js
 
+      - name: Test SessionStart hook
+        run: node --test hooks/session-start-node-test.js
+
   validate-commands:
     name: Validate command parity and description sync
     runs-on: ubuntu-latest
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,14 +76,16 @@ We don't accept translations of the documentation (README, `docs/`) or of skills
 
 ## Testing Hooks
 
-The session-start hook (`hooks/session-start.sh`) injects the `using-agent-skills` meta-skill into every new Claude Code session. A regression test at `hooks/session-start-test.sh` validates the hook's JSON payload — both when `jq` is available and when it isn't.
+The session-start hook (`hooks/session-start.mjs`) injects the `using-agent-skills` meta-skill into every new Claude Code session. It is Node rather than bash + `jq` so it runs unchanged on Windows, where Claude Code falls back to PowerShell when Git Bash is absent (#488); `hooks/session-start.sh` is kept for direct invocation. A regression test at `hooks/session-start-test.sh` validates the hook's JSON payload — both when `jq` is available and when it isn't.
 
 Run it before opening any PR that touches:
 
-- `hooks/session-start.sh`
+- `hooks/session-start.mjs` or `hooks/session-start.sh`
+- `hooks/hooks.json`
 - `skills/using-agent-skills/SKILL.md` (the meta-skill content embedded by the hook)
 
 ```bash
+node --test hooks/session-start-node-test.js
 bash hooks/session-start-test.sh
 ```
 
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -5,7 +5,8 @@
         "hooks": [
           {
             "type": "command",
-            "command": "SCRIPT=\"${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh\"; [ -f \"$SCRIPT\" ] || SCRIPT=\"${CLAUDE_PROJECT_DIR}/.claude/hooks/session-start.sh\"; [ -f \"$SCRIPT\" ]&& bash \"$SCRIPT\" || true"
+            "command": "node",
+            "args": ["${CLAUDE_PLUGIN_ROOT}/hooks/session-start.mjs"]
           }
         ]
       }
--- a/hooks/session-start-node-test.js
+++ b/hooks/session-start-node-test.js
@@ -1,0 +1,87 @@
+// Tests for the Node SessionStart hook (#488).
+//
+// The hook used to be a bash one-liner in hooks.json calling a bash+jq script.
+// On Windows, Claude Code runs hook commands through Git Bash or — when that is
+// absent — PowerShell, where `[ -f "$SCRIPT" ]` is a parse error, so the hook
+// failed on every session start. These pin the two properties that fix depends
+// on: the config no longer goes through a shell, and the payload is unchanged.
+
+const test = require("node:test");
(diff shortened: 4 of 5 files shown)


addyosmani (approved):
This is the right fix and I verified it end to end: the Node hook emits the same SessionStart envelope, its test passes, and the test is finally wired into CI, which closes the gap I kept flagging on the bash hook tests. Going Node also solves the Windows/PowerShell problem (#488) at the root rather than patching bash. One thing before it merges, and it's a direction call rather than a code issue: this effectively supersedes the bash session-start work in flight. #480 (Windows bash fix), #510 (bash envelope test), and #524 (bash hook refactor) are all working the same hook the other way. If we go Node here, those should close or narrow. I don't want to merge this and strand three open PRs, so let's make that call, ideally with Joan, first. Approving the approach; holding the merge on the direction.

addyosmani:
This is the strongest engineering in the hook set: a clean Node port with a byte-identical payload test wired into CI, and it does fix the Windows breakage. The catch is it's mutually exclusive with #574, which removes the auto-injection hook entirely on the grounds that the plugin shouldn't stack a second router on top of the host's native discovery (see #569 and getting-started.md). So this is a direction question, not a quality one. Holding both while I settle it. Thank you for the care here either way.

addyosmani:
Going with #574 for the SessionStart direction, so I'm closing this. Why, briefly: #574 removes the auto-injection hook rather than porting it, which fixes the same Windows breakage you targeted here and also resolves the token waste in #569 while keeping us consistent with getting-started.md (don't stack a second router on a host that already routes natively). None of that is a knock on this PR. It's the best-engineered one in the set, and the Node exec-form pattern is exactly right if we ever ship an opt-in hook. Thank you for the care, genuinely.

PR #544, opened 2026-09-02
docs: add a rejected skill-change ledger

### LINKED ISSUE #535: Proposal: keep a durable record of rejected skill and description changes
## Summary

Keep a durable, append-only record of skill and description changes that were **proposed and rejected**, together with the eval score that rejected them. Today that reasoning exists only in PR threads, and the eval output that produced it is discarded.

To be explicit about scope: this is about **this catalog's own development loop**, not about memory inside a user's project. It is not a third memory system, and it does not overlap with [#398](https://github.com/addyosmani/agent-skills/issues/398) or [#435](https://github.com/addyosmani/agent-skills/issues/435), both of which target user repositories.

## The loop we already have

We run a validate-propose-gate loop today, we just do not keep its history:

| Stage | Where it lives | Persisted? |
|---|---|---|
| Propose a skill or description change | PR | yes, in git |
| Score it | `scripts/run-evals.js` | `evals/results/` is gitignored |
| Gate it | CI, `run-evals.js --min-rank1 80` | pass/fail only |
| Reject and revert | PR close or force-push | rationale scattered in review threads |

The rank-1 floor of 80 is an acceptance threshold in exactly the sense a skill-evolution loop uses one. What is missing is the record of what was tried against it.

## Evidence that the record is the valuable part

[WikiSkill](https://huggingface.co/papers/2608.27454) (arXiv:2608.27454, Google Research) separates skills, which are reverted when they fail validation, from an append-only knowledge layer that is **never** reverted, including a `skill-impact.md` that stores each proposal's target skill, unified diff, validation score, and accept/reject outcome.

Their ablation isolates what that record is worth. Holding everything else fixed, giving the component that proposes skill changes access to the accumulated record moves average benchmark performance from 48.7% to 63.7% (Table 3, Gemini-3.5-Flash). Their case study is the concrete version: a proposal is rejected at iteration 0 for being too abstract, the rejection diff is preserved, and the accepted proposal at iteration 1 is written specifically to avoid repeating it.

## The symptom here

[#315](https://github.com/addyosmani/agent-skills/issues/315) ("CI: label PRs that add a skill duplicating an open PR") exists because near-duplicate proposals keep arriving. That is the same shape of problem: each contributor, human or agent, starts without knowing what has already been tried and turned down. The `CONTRIBUTING.md` pre-flight asks people to search the catalog and open PRs, which finds what exists and what is in flight, but not what was rejected and why.

## Sketch

A tracked `evals/skill-impact.md`, appended to rather than rewritten, with one row per gated change:

```
| Date | Skill | Change | rank-1 before → after | Outcome |
|---|---|---|---|---|
| 2026-08-31 | example-skill | broadened description to cover "audit" | 84 → 79 | reverted (#000) |
```

Open questions I do not have a strong opinion on:

1. **Generated or hand-written?** Generated by `run-evals.js` is accurate but touches `scripts/`, which CLAUDE.md flags as collision-prone. Hand-written is zero-infrastructure but rots.
2. **Rejections only, or accepted changes too?** The paper keeps both. Rejections carry most of the value per line.
3. Is this better as a section in `CONTRIBUTING.md` pointing at closed PRs, rather than a new tracked file?

## Caveat

The +15.0 figure is measured for an automated proposer optimizing benchmark accuracy. Our proposer is a human contributor plus review, so the number does not transfer, and I am not claiming it does. The argument that transfers is structural: the loop discards its own history, and [#315](https://github.com/addyosmani/agent-skills/issues/315) is what that costs.

@addyosmani @federicobartoli, worth pursuing, and if so which of the three open questions above? Happy to leave it as documentation-only if touching `scripts/` is unwelcome right now. Related: #534, the authoring rule drawn from the same paper.

mvanhorn:
## Verification

- A contributor considering a new or revised skill can follow `CONTRIBUTING.md` to find both existing/open work and previously rejected proposals in `evals/skill-impact.md`.
- A rejected description change can be recorded in one table row containing the date, affected skill, concise attempted change, before-to-after rank-1 receipt, and a working link to the rejected PR.
- The instructions make clear that the ledger-only row must land independently of the rejected proposal, so closing or force-pushing that proposal cannot discard the historical record.
- Markdown renders with a valid table and relative ledger link, and the existing deterministic eval command `node scripts/run-evals.js --min-rank1 80` remains green because the eval runner and skill descriptions are unchanged.

## Summary

Add `evals/skill-impact.md` as a rejection-only ledger, with a brief purpose statement and a compact table whose columns capture date, affected skill, attempted change, before-to-after rank-1 score, and rejected PR link/outcome. Keep entries hand-written and append-only: automation in `run-evals.js` would write onto the proposal branch that is about to be closed or force-pushed, so it would not make the rejection durable without substantially broader workflow automation. Update the new-skill and existing-skill pre-flight guidance in `CONTRIBUTING.md` to search the ledger before proposing overlapping work, and document the closeout step: after an eval-driven rejection, land the single ledger row separately from the rejected change so it reaches the default branch.

## Why

The catalog's deterministic eval runner reports routing quality, but rejected skill and description changes leave no durable record once their PR is closed or rewritten. That makes contributors repeat proposals whose rationale and failed eval receipt are scattered across old review threads. The owner approved a catalog-scoped, append-only register and narrowed the desired interaction to one short entry per rejection with a PR link and eval evidence. There are no prior closed implementation PRs for this issue, and the bundled open PR #537 is an unrelated description-vocabulary change rather than a competing ledger implementation.

Closes #535

AI was used for assistance.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,8 +12,9 @@ This pack already covers most of the development lifecycle, and many proposals o
 
 1. **Search the catalog.** Browse [the skill list in the README](README.md) and skim `skills/` for an existing skill that covers your idea, whole or in part.
 2. **Check open PRs.** Run `gh pr list --state open` (or browse the PRs tab) and look for proposals on the same topic. Clusters of near-duplicate skills already exist; don't add to them.
-3. **Read the anatomy.** Confirm your idea fits the format in [docs/skill-anatomy.md](docs/skill-anatomy.md), an actionable workflow with verification, not vague advice.
-4. **Justify the gap in your PR description.** State explicitly why this isn't covered by an existing skill or open PR. If it overlaps, propose extending the existing skill instead of adding a new one.
+3. **Check rejected proposals.** Search the [skill-change rejection ledger](evals/skill-impact.md) for earlier proposals that overlap with your idea and review their eval evidence before repeating the work.
+4. **Read the anatomy.** Confirm your idea fits the format in [docs/skill-anatomy.md](docs/skill-anatomy.md), an actionable workflow with verification, not vague advice.
+5. **Justify the gap in your PR description.** State explicitly why this isn't covered by an existing skill, open PR, or previously rejected proposal. If it overlaps, propose extending the existing skill instead of adding a new one.
 
 If your idea is a refinement of an existing skill, prefer a focused edit to that skill over a new directory.
 
@@ -62,9 +63,13 @@ The frontmatter fields above are required. The section anatomy is a recommended 
 
 ## Modifying Existing Skills
 
+Before proposing a change, search the [skill-change rejection ledger](evals/skill-impact.md) for previous attempts affecting the same skill and review their eval evidence.
+
 - Keep changes focused and minimal
 - Preserve the existing structure and tone
 - Test that YAML frontmatter remains valid after edits
+
+If a skill or description change is rejected based on eval results, add one row to the ledger with the date, affected skill, concise attempted change, before-to-after rank-1 score, and rejected PR link and outcome. Land that ledger-only update separately on the default branch; do not leave it only on the rejected proposal branch, where closing or force-pushing the proposal could discard the record.
 
 ## Repo-scoped files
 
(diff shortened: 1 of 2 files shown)


addyosmani (approved):
Clean implementation of #535, and it gets the important part right: the ledger row lands separately on the default branch so closing or force-pushing the rejected PR can't discard the record. Checking the ledger is now a pre-flight step in CONTRIBUTING, which is where it belongs. Verified clean. Approving.

PR #545, opened 2026-09-02
docs: add authoring rule against model-specific workarounds (#534)

### LINKED ISSUE #534: Proposal: authoring rule against model-specific workarounds in skill content
## Summary

Add an authoring rule to `docs/skill-anatomy.md` (and a pre-flight line in `CONTRIBUTING.md`) that treats **model-specific workarounds in skill content as a portability liability**, not as helpful detail. A skill that exists to route around one model's execution failures can actively degrade a stronger model.

This repo currently has guidance on the portability of `references/` *paths* ([#361](https://github.com/addyosmani/agent-skills/issues/361)), but nothing on the portability of skill *content* across models.

## Evidence

Google Research recently published [WikiSkill](https://huggingface.co/papers/2608.27454) (arXiv:2608.27454). Most of it is about automated skill evolution and does not apply here, but its cross-model transfer study measures exactly the failure mode we keep seeing anecdotally.

Skills evolved by a small model, then run on a stronger one (Table 2, SpreadsheetBench):

| Inference model | Skill source | Score |
|---|---|---|
| Gemini-3.5-Flash | none | 50.5% |
| Gemini-3.5-Flash | Qwen-3.5-4B | **18.1%** |
| Gemini-3.5-Flash | Qwen-3.6-27B | 63.4% |

The skills authored against the weaker model cut the stronger model's score by roughly two thirds, well below using no skill at all. Their error analysis names two mechanisms:

1. The skills encode **low-level workarounds** (single-line Python commands, string-conversion rules) that help a weak model avoid execution failures but prevent a stronger model from writing comprehensive end-to-end scripts.
2. **Fragmented diagnostic procedures** introduce redundant tool calls that exhaust the interaction budget before the task completes.

The inverse also happens: skills evolved by Qwen-3.5-4B on OfficeQA slightly hurt that same model (30.2% to 28.5%) while lifting Qwen-3.6-27B from 42.1% to 52.9%. Discovering useful procedural knowledge and executing it are separable capabilities, so "it worked when I tried it" is weak evidence that a rule belongs in a shared catalog.

## Why this matters for this repo

Mechanism 2 is a plausible, and now named, explanation for [#423](https://github.com/addyosmani/agent-skills/issues/423) (the pack being slow on GPT-5.6): procedural fragmentation costs turns. Mechanism 1 gives a concrete authoring axis for the drift work in [#432](https://github.com/addyosmani/agent-skills/issues/432) and [#433](https://github.com/addyosmani/agent-skills/issues/433): rules written against a specific model generation are the ones most likely to age badly, and the failure is silent because they still read as reasonable advice.

## Proposed rule (draft wording)

> **Write the procedure, not the workaround.** A step that exists because one model gets a specific call wrong is a liability on every other model: it constrains agents that would have solved the task directly, and it spends turns that stronger models need for the actual work. If a rule cannot be justified without naming a model or a model version, it belongs in an issue, not in a skill.
>
> Two smells to check for before proposing skill content:
> - **Over-specified mechanics.** Prescribing an exact command form or serialization detail where the goal alone would do.
> - **Fragmented diagnostics.** A multi-step inspection sequence where one step would confirm the same thing. Every extra step is turns spent, and turn budget is finite.

## Scope

Docs only. No skill content changes in this issue; if the rule lands, auditing existing skills against it is separate follow-up work, and probably belongs with whatever [#432](https://github.com/addyosmani/agent-skills/issues/432) produces.

## Caveats, stated up front

The paper injects full skill content into the system prompt and says in its Limitations that this is deliberate, to remove skill triggering and retrieval as confounding variables. So none of its numbers say anything about description quality or routing, which is what our Tier 2 evals measure. The evaluated models are also mid-size open-weight models, not frontier coding agents. I am citing the *mechanism* and the direction of the effect, not proposing we treat the percentages as applying to this catalog.

@addyosmani @federicobartoli, does this look like a rule worth having written down, or is it already covered implicitly by the "no vague advice" bar? Related: #535 on keeping a durable record of rejected changes.

ayobamiseun:
Implements #534, which @addyosmani endorsed ("Writing it into skill-anatomy and CONTRIBUTING is the right move… I'd take this"). Checked before starting: no open or closed PR references the issue, and no commits touch these sections for it.

## What changed

**`docs/skill-anatomy.md`**
- New section **Write the Procedure, Not the Workaround**: the rule (a step that can't be justified without naming a model, a model version, or one agent's private tool name belongs in an issue, not a skill — describe the capability, not the mechanism one runtime exposes), the two smells from #534's evidence (**over-specified mechanics**, **fragmented diagnostics**), and the WikiSkill citation for the cross-model transfer failure mode. Explicitly scoped to skill *content*, with a pointer distinguishing it from the `references/` *path* portability concern (#361).
- Writing Principles gains **7. Model-neutral** linking to the section.

**`CONTRIBUTING.md`**
- Pre-flight check 5: **Strip model-specific workarounds**, one line, linking to the anatomy section.

## Framing choices

- Kept tight per the review guidance on #534 — the rule is stated in terms of capability vs. mechanism, and the tool-name half encodes the #497 review precedent (hardcoded `run_command`/`write_to_file`) so the ad-hoc review position is now written down.
- Per the issue's own caveats, the citation is used for the *mechanism and direction* of the effect, not the percentages.
- Docs only; auditing existing skills against the rule stays follow-up work with #432, as the issue scoped it.

## Verification

- `node scripts/validate-skills.js` — 25 skills, 0 errors
- `node scripts/validate-reference-links.js` — 0 errors
- Anchors checked: both new links resolve (`#write-the-procedure-not-the-workaround`, `#shared-references`)

Closes #534.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,6 +15,7 @@ This pack already covers most of the development lifecycle, and many proposals o
 3. **Check rejected proposals.** Search the [skill-change rejection ledger](evals/skill-impact.md) for earlier proposals that overlap with your idea and review their eval evidence before repeating the work.
 4. **Read the anatomy.** Confirm your idea fits the format in [docs/skill-anatomy.md](docs/skill-anatomy.md), an actionable workflow with verification, not vague advice.
 5. **Justify the gap in your PR description.** State explicitly why this isn't covered by an existing skill, open PR, or previously rejected proposal. If it overlaps, propose extending the existing skill instead of adding a new one.
+6. **Strip model-specific workarounds.** If a step can't be justified without naming a model, a model version, or one agent's private tool name, it doesn't belong in a skill — describe the capability instead (see [Write the Procedure, Not the Workaround](docs/skill-anatomy.md#write-the-procedure-not-the-workaround)).
 
 If your idea is a refinement of an existing skill, prefer a focused edit to that skill over a new directory.
 
--- a/docs/skill-anatomy.md
+++ b/docs/skill-anatomy.md
@@ -147,6 +147,20 @@ When a skill ships runnable helpers under `scripts/`, each script follows these 
 4. **Anti-rationalization.** Every skip-worthy step needs a counter-argument in the rationalizations table.
 5. **Progressive disclosure.** Main SKILL.md is the entry point. Supporting files are loaded only when needed.
 6. **Token-conscious.** Every section must justify its inclusion. If removing it wouldn't change agent behavior, remove it.
+7. **Model-neutral.** Write the procedure, not the workaround. See [Write the Procedure, Not the Workaround](#write-the-procedure-not-the-workaround).
+
+## Write the Procedure, Not the Workaround
+
+Skills in this pack run on many agents and model generations. A step that exists because one model gets a specific call wrong is a liability on every other model: it constrains agents that would have solved the task directly, and it spends turns that stronger models need for the actual work. Cross-model transfer measurements ([WikiSkill, arXiv:2608.27454](https://arxiv.org/abs/2608.27454)) show skills authored against one model's execution failures can leave a stronger model performing *worse than with no skill at all*.
+
+**The rule:** if a step cannot be justified without naming a model, a model version, or one agent's private tool name, it belongs in an issue, not in a skill. Describe the capability ("run the focused test command", "write the file"), not the mechanism one runtime happens to expose (`run_command`, `write_to_file`).
+
+Two smells to check for before proposing skill content:
+
+- **Over-specified mechanics.** Prescribing an exact command form or serialization detail where stating the goal alone would do. If the justification is "model X gets this wrong otherwise", it's a workaround, not a procedure.
+- **Fragmented diagnostics.** A multi-step inspection sequence where one step would confirm the same thing. Every extra step spends turns, and turn budget is finite.
+
+This is about skill *content*; the portability of `references/` *paths* is a separate concern (see [Shared References](#shared-references)).
 
 ## Naming Conventions
 


addyosmani (changes requested):
The rule is good and I backed it in #534, but one thing before it lands: the WikiSkill citation (huggingface.co/papers/2608.27454) is a specific paper I can't verify, and the ID looks future-dated. A skill-anatomy rule shouldn't rest on a citation I can't confirm is real and says what we claim it does. The principle stands on its own reasoning: a workaround for one model's failure constrains, and can degrade, a stronger one. So either point me at the verifiable source or drop the citation and keep the argument. Small fix, then I'm happy to take it.

ayobamiseun:
Fair challenge — a normative rule shouldn't lean on a citation the reviewer can't check. So I verified it end to end before answering: **the paper is real, and it says exactly what the doc claims.**

**Verifiable source:** [arxiv.org/abs/2608.27454](https://arxiv.org/abs/2608.27454) — *WikiSkill: Compiling Agent Experience into Persistent Knowledge for Skill Evolution* (Tang, Rashtchian, Ferng, Tomkins, Juan, Vu). The ID isn't future-dated: `2608` is 2026-08, i.e. last month. The doc had cited the Hugging Face mirror, which resolves but renders poorly for verification — that's on me, and 0287de4 swaps the link to the canonical arXiv abstract.

**Checked against the full text ([HTML](https://arxiv.org/html/2608.27454)):**
- Table 2, SpreadsheetBench: Gemini-3.5-Flash scores **50.5% with no skills vs 18.1% with Qwen-3.5-4B-evolved skills** — the "worse than no skill at all" claim, verbatim in the numbers.
- The two smells in the doc are near-quotes of the paper's error analysis: skills "encode low-level workarounds, such as single-line Python commands and string-conversion rules, which help the smaller model avoid execution failures but constrain stronger models … from using comprehensive end-to-end scripts", and "fragmented diagnostic procedures introduce redundant tool calls that can exhaust … interaction budget before task completion."
- The OfficeQA asymmetry from #534 also checks out (30.2→28.5 self, 42.1→52.9 transferred up).

Also in 0287de4: rebased over #544, so the CONTRIBUTING pre-flight line is now item 6 after the new rejection-ledger check, and the conflict is resolved keeping both. Validators pass.

If you'd still rather the doc stand on the argument alone, I'm happy to drop the citation — but having checked it, I think one verifiable measurement of the failure mode earns its half-line.

addyosmani (approved):
Thanks for swapping to the canonical arXiv link. That resolves my earlier concern, the paper checks out. Clearing my change request and merging.

PR #548, opened 2026-09-05
fix(catalog): generate skill listings and detect drift (#539)

### LINKED ISSUE #539: Generate the skill catalog instead of hand-maintaining it across surfaces
## Summary

The skill list is maintained by hand across several tracked files, and nothing in CI checks that those copies agree with `skills/`. Adding a skill means remembering every surface. There is currently drift on `main` because one of those surfaces was missed.

Proposal: derive the catalog surfaces from `skills/` with a script that has a `--check` mode, and run that check in CI. Every existing script under `scripts/` validates; none generates. This is the first case where generating is cheaper than remembering.

## Current drift on `main`

`constraint-driven-development` shipped in #472 (`2ce8d47`, 28 Aug) and is missing from two places:

| Surface | State |
|---|---|
| `README.md` | present |
| `skills/using-agent-skills/SKILL.md` | present |
| `CLAUDE.md`, "Skills by Phase" (line 23) | **missing** |
| `.github/ISSUE_TEMPLATE/skill-gap.yml` dropdown | **missing** |

The issue-form gap has a user-visible consequence: someone hitting bad guidance from `constraint-driven-development` today cannot select it when filing a skill-gap report. The form landed in `b68f553` (19 Jul), the skill five weeks later, and nothing connected them.

## Why this is a process gap and not one contributor's slip

#472 was not careless. It updated ten files across the catalog:

```
README.md
skills/using-agent-skills/SKILL.md
docs/adoption-guide.md
docs/codex-setup.md
docs/commandcode-setup.md
docs/comparison.md
evals/cases/constraint-driven-development.json
commands/constraints.toml
.claude/commands/constraints.md
.gemini/commands/constraints.toml
```

It still missed two. Ten hand-updated surfaces per skill is past the point where care is the right mechanism, and the number grows with every runtime integration the repo adds (#530, #355, #439, #477 are all open requests that would add more).

## Sketch

`scripts/sync-catalog.js`, reading `skills/*/SKILL.md` frontmatter as the single source of truth:

- `--check` (CI): exit non-zero listing any surface whose skill set disagrees with `skills/`
- `--write` (local): update the generated regions in place

Generated regions marked with comment fences so hand-written prose around them is untouched, for example `<!-- catalog:start -->` / `<!-- catalog:end -->` in Markdown and the equivalent in the issue form's `options:` block.

Starting with `--check` alone would already have caught this, and is the smaller change if writing into tracked docs feels too invasive for a first cut.

## Open questions

1. Which surfaces should be authoritative-generated versus merely checked? `README.md` prose probably wants a human touch; the issue-form dropdown is pure data and could be fully generated.
2. Does `CLAUDE.md`'s phase grouping belong in frontmatter (a `phase:` field) or in a small mapping file the script reads? The phase is currently implicit in prose.
3. Should this live in an existing validator rather than a new script? `validate-skills.js` already walks `skills/`, so the check could be a mode there instead.

## Not overlapping with existing work

#137 covers markdown formatting consistency (`markdownlint`, heading levels, list markers), which is orthogonal to catalog membership. #511 is about doc/code drift as guidance inside the `documentation-and-adrs` skill, not this repo's own catalog. I did not find an open PR touching this.

## Credit

The generate-plus-`--check` pattern is borrowed from [affaan-m/ecc](https://github.com/affaan-m/ecc), which ships `catalog:check` / `catalog:sync` and `command-registry:generate --check` for the same problem at a larger catalog size. The idea is theirs; no code is proposed to be copied.

Happy to open a PR for the `--check` half if the direction is right. Given that `scripts/` is collision-prone with in-flight work (#428, #444, #517 all touch validators), I would rather agree the shape here first.

DEOWL-kan:
Fixes #539.

Adding a skill currently leaves several catalogs to update by hand: `constraint-driven-development` is missing from the CLAUDE phase list, meta-skill quick reference, and issue dropdown, while the README heading still says 24 skills. This generates those regions and README counts/tables from the actual `skills/` membership plus a small editorial presentation file, and adds `--check` to CI. `--write` preserves surrounding prose; the authored project/discovery trees are checked for membership rather than rewritten.

The generator reuses the existing frontmatter parser, adds no dependencies, and rejects invalid or overlapping markers before writing any target. Eighteen regression tests cover drift, additions/removals, read-only checking, idempotence, prose preservation, and malformed/overlapping regions. No new skill or shared-validator rewrite is introduced.

### Validation

- PASS: `node --test scripts/*-test.js scripts/lib/*-test.js` — 61 tests.
- PASS: `node scripts/sync-catalog.js --check`; skill, command, artifact-path, reference-link, and version validators; `node scripts/run-evals.js --min-rank1 80` (86% rank-1); `git diff --check`.
- PASS: independent read-only review, including eight in-memory overlap cases that attempted zero writes; actual SessionStart envelope, embedded meta-skill content, and isolated no-jq fallback assertions.
- Known baseline failure: `bash hooks/session-start-test.sh` expects the retired `priority` payload and fails identically on unchanged main. #510 already addresses that test; hooks and the test are unchanged here.

Open PRs were checked before submission. This keeps existing validator and hook work separate and preserves the authored content around generated regions.


addyosmani (approved):
Strong implementation of Joan's #539, and I verified it works: --check passes (the generated surfaces match what's on main today, so it's a faithful capture, not a reformat), the tests pass, validate-skills and validate-commands stay green, and --check is wired into CI. That closes the drift problem cleanly. One design point that's really Joan's call since it's his proposal: this introduces scripts/catalog.json as the single curated source (phases plus per-skill metadata) and generates the surfaces from it, rather than deriving everything straight from skills/. That's a reasonable read of #539 (phase, summary, and order aren't in SKILL.md frontmatter, so they have to live somewhere), and one catalog.json plus generated output beats N hand-maintained lists. But I'd want @nucliweb's sign-off on catalog.json as the source of truth before it lands. Mechanics are verified and green; over to Joan on the design.

ibisovelmeddin411-maker:
Hu

PR #551, opened 2026-09-05
fix(commands): avoid Claude /review collision

### LINKED ISSUE #95: `/review` slash command collides with Claude Code's bundled `/review` — intentional or worth namespacing?
## Summary

Installing this plugin into Claude Code registers a `/review` slash command that
shadows Claude Code's **built-in** `/review` (and conceptually overlaps with the
built-in `/security-review`). I couldn't find any mention of this in the README
or existing issues, so I'd like to confirm whether the collision is intentional
before choosing an install strategy.

## Evidence

Claude Code ships `/review`, `/security-review`, and `/init` as bundled skills
invocable via the Skill tool — documented at:
- https://code.claude.com/docs/en/commands.md
- https://code.claude.com/docs/en/skills.md
  > "A few built-in commands are also available through the Skill tool,
  > including `/init`, `/review`, and `/security-review`."

This repo's README lists 7 slash commands, one of which is literally `/review`.
Because Claude Code resolves project → user → plugin → built-in, the
plugin-registered `/review` wins and the built-in becomes unreachable.

Interestingly, the **skill slugs** in this repo (`code-review-and-quality`,
`code-simplification`, `security-and-hardening`, …) are all long-form and do
*not* collide with any Claude Code built-in skill — which reads as deliberate
namespacing. Only the slash command layer collides.

## Questions

1. Is the `/review` name intentional (i.e., you want to replace the built-in),
   or was the built-in's existence not considered?
2. If intentional, would you be open to documenting this in the README so
   users can make an informed choice?
3. If unintentional, would you accept a PR that renames the slash commands to
   a namespaced form (e.g., `/as-review`, `/agent-review`, or `/lifecycle-review`)
   while keeping the underlying skill slugs unchanged?

## Workaround I'm considering locally

Keeping the plugin installed but overriding `/review` back to the built-in via
a project-level `.claude/commands/review.md` that just invokes the
`code-review-and-quality` skill on demand, so the built-in stays reachable by
default. Happy to share if useful.

kaluli123123:
Closes #95.

## Summary

- rename the lifecycle review wrapper from `/review` to `/code-review` across Claude Code, Gemini CLI, and Antigravity command directories
- update every repository-owned command reference and add an upgrade note
- make the command parity validator reject names that shadow Claude Code built-ins

Claude Code's current command reference still reserves `/review` for its local pull-request review workflow, so the old wrapper hides built-in behavior.

## Verification

- RED: `node --test scripts/validate-commands-test.js` failed on the new collision regression before the validator change (6 passed, 1 failed)
- GREEN: combined Node test suite: 44 passed, 0 failed
- `node scripts/validate-skills.js`: 25 skills, 0 errors, 0 warnings
- `node scripts/run-evals.js --min-rank1 80`: 140 checks passed; rank-1 86%
- command, reference-link, artifact-path, and version validators passed
- `bash hooks/simplify-ignore-test.sh`: 21 passed, 0 failed
- `git diff --check`: passed

## Existing unrelated failure

`bash hooks/session-start-test.sh` fails because the current hook payload has no `priority` field while the test expects `IMPORTANT`. This PR does not touch the SessionStart hook; native Windows hook work is already active in #475 and #488.


addyosmani (approved):
Correct, and it verifies clean with command parity intact. It fixes a real collision: Claude Code ships a native /review that shadows ours, so /code-review is the right name (the rename #181 asked for). Approving. The only reason I'm not merging immediately is that it's user-facing, anyone with muscle memory or docs pointing at /review switches to /code-review, so I want that to land deliberately.

abhisheksharma2411 (commented):
The rename is the visible half; the part worth merging for is that it also **adds the guard so the collision can't come back**. That's easy to miss in a diff that's mostly doc churn.

**Verified the rename is complete.** Old command files are gone (`review.md`/`review.toml` → `code-review.*` in all three command dirs), and only two bare `/review` mentions remain across the tree, both deliberate:

```
README.md:47                      the upgrade note explaining the removal
scripts/validate-commands-test.js:82   asserts the validator detects the collision
```

No stale references in the docs, which is the usual failure mode for a rename touching 14 files.

**Verified the guard actually fires**, rather than trusting the test name — reintroduced `.claude/commands/review.md` on your branch:

```
10 commands checked — 3 error(s) — FAILED
  x  review: conflicts with a Claude Code built-in command
```

and `main` doesn't flag it today because `CLAUDE_BUILT_IN_COMMANDS` doesn't exist there yet. So this PR turns a latent shadowing bug into a CI-enforced invariant, which is worth more than the rename on its own.

**One note on the list, not a blocker.** `CLAUDE_BUILT_IN_COMMANDS` is a hardcoded set with one entry:

```js
const CLAUDE_BUILT_IN_COMMANDS = new Set(['review']);
```

That's the right shape to start with — a guard covering the one known collision beats no guard, and guessing at a full built-in list would produce false failures. But it will silently go stale as Claude Code adds commands, and the failure mode is quiet: a future `/init` or `/compact` command lands, shadows the built-in, and nothing complains. Worth a comment saying the list is known-incomplete and how to extend it, so the next person treats it as a registry rather than a finished enumeration.

Also worth confirming the Gemini/generic `commands/*.toml` side: the guard loops `claudeStems` only, which is correct today because the collision is Claude-specific — but if the intent is "don't shadow any host's built-ins," that's a different loop. The current scoping matches the comment (`Plugin commands with these names hide Claude Code's built-in commands`), so I read it as deliberate.

`9 commands checked — 0 errors` on the branch, parity intact across all three command directories.

PR #552, opened 2026-09-05
docs(skills): clarify stable browser skill name

### LINKED ISSUE #541: Document compatibility alias when renaming browser-testing-with-devtools
## Context

In downstream/provider-neutral adaptations, `browser-testing-with-devtools` may be renamed to `playwright-cli` to match the local runtime. The upstream skill currently remains valid, but consumers that refer to the original name become orphaned if the rename is performed without an alias or migration rule.

## Request

Could the repository document one of these compatibility mechanisms?

- keep a lightweight `browser-testing-with-devtools` alias pointing to the canonical skill; or
- publish a clear rename/migration rule for consumers such as `using-agent-skills` and `test-driven-development`; or
- explicitly state that the name is stable and must not be renamed by downstream catalogs.

## Why it matters

A catalog can pass frontmatter validation while still directing agents to a missing skill after a downstream rename. A stable alias or documented migration would prevent broken discovery without duplicating the implementation.

This request is about compatibility/documentation only; it does not claim that the current upstream reference is broken.

kaluli123123:
Closes #541.

## Summary

Document `browser-testing-with-devtools` as the stable upstream skill identifier in the authoring guide.

This follows the maintainer decision on the issue:

- upstream keeps the existing name because other skills refer to it directly
- downstream catalogs may rename it
- the downstream catalog owns any compatibility alias or migration mapping
- this repository does not accumulate aliases for fork-specific renames

No alias, duplicate skill, or runtime behavior is added.

## Verification

- `git diff --check`: passed
- inspected all repository references to `browser-testing-with-devtools`
- docs-only change; no executable tests were required

### DIFF
--- a/docs/skill-anatomy.md
+++ b/docs/skill-anatomy.md
@@ -32,6 +32,12 @@ description: Guides agents through [task/workflow]. Use when [specific trigger c
 - `name`: Lowercase, hyphen-separated. Must match the directory name.
 - `description`: Start with what the skill does in third person, then include one or more clear "Use when" trigger conditions. Include both *what* and *when*. Maximum 1024 characters.
 
+Published names are compatibility identifiers. In particular,
+`browser-testing-with-devtools` is the stable upstream name because other
+skills refer to it directly. A downstream catalog may rename it, but that
+catalog owns the alias or migration mapping; downstream-only aliases are not
+maintained in this repository.
+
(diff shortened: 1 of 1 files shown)


addyosmani (approved):
Right call and the right home: documenting in skill-anatomy that published names are stable compatibility identifiers. Consistent with how I handled #541. Merging.

PR #557, opened 2026-09-05
docs: avoid double routing on native skill hosts

### LINKED ISSUE #423: This skill set is terribly slow in GPT-5.6
The first slow source is using-agent-skills. When it adds up to GPT's built-in skill router, it's a disaster.

The second, I really don't know. I just feel GPT-5.6 to be slow enough for itself and when I equip it with 24/23 skills, a runtime rebuilt takes 6 hours to do. This skill set work well with me in GPT-5.5, but I think it doesn't fit in with 5.6.

I'm not professional in this. It's just a user report.

kaluli123123:
## Summary

- scope `using-agent-skills` guidance to hosts without native skill routing
- tell native hosts to use one router instead of preloading the meta-skill
- document the Codex-specific progressive-disclosure behavior and link the official docs

## Verification

- `git diff --check`
- `node scripts/validate-reference-links.js` (25 skills, 0 errors)
- `node scripts/validate-artifact-paths.js` (7 files, 0 errors)
- official Codex skills documentation returned HTTP 200

Closes #423

### DIFF
--- a/docs/adoption-guide.md
+++ b/docs/adoption-guide.md
@@ -28,7 +28,7 @@ A new project is the best-case scenario: there's no legacy behavior to preserve,
 ### Day 0 | Install and wire up
 
 1. Install the pack (`npx skills add addyosmani/agent-skills`, or the native integration for your tool, see [getting-started.md](getting-started.md)).
-2. Load `using-agent-skills` (the meta-skill) so the agent can route work to the right skill on its own.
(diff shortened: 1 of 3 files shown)


addyosmani (approved):
Good, this is exactly the 'don't load our router where the host already has one' guidance I've wanted, no double-routing on native skill hosts. Approving and merging.

PR #563, opened 2026-09-07
fix(evals): reject null grader expectations without crashing

RogueTex:
# fix(evals): reject null grader expectations without crashing

When a behavioral grader returns valid JSON containing a null expectation, `parseGrading` throws a `TypeError` while counting passed expectations. This aborts the batch before the invalid response is saved and prevents later evals from running.

Guard both expectation reads so the parser returns `null` for these malformed results. The existing caller then saves the raw response, counts the evaluation as failed, and continues. Valid grader responses retain their existing behavior.

The regression test covers an all-null expectation list and a mixed valid/null list. It failed with `Cannot read properties of null (reading 'passed')` before the fix and passes afterward.

Validation:

- All 44 JavaScript tests passed: `node --test scripts/*-test.js scripts/lib/*-test.js`.
- Deterministic evaluations passed: 140 checks; rank-1 routing 76/88 (86%), above the required 80%.
- Skill, version, reference-link, command, and artifact-path validators passed.
- A local CLI smoke test with a simulated grader verified raw invalid-response persistence, completion of a second evaluation, and exit code 1 for the batch. No live model calls were made.
- `git diff --check` passed.

This follows the validation added in https://github.com/addyosmani/agent-skills/pull/473 and addresses the remaining null-entry crash. No matching open fix was found in the PR search.

### DIFF
--- a/scripts/run-evals-test.js
+++ b/scripts/run-evals-test.js
@@ -90,6 +90,25 @@ test('rejects grader results that omit expectations', () => {
   assert.equal(parseGrading(raw, 2), null);
 });
 
+test('rejects null expectation entries without throwing', () => {
+  for (const expectations of [
+    [null],
+    [{ text: 'first expectation', passed: true, evidence: 'observed' }, null],
+  ]) {
+    const raw = JSON.stringify({
+      expectations,
+      summary: {
+        passed: expectations.length - 1,
+        failed: 1,
+        total: expectations.length,
+        pass_rate: (expectations.length - 1) / expectations.length,
+      },
+    });
+
+    assert.equal(parseGrading(raw, expectations.length), null);
+  }
+});
+
 test('rejects incomplete or inconsistent grader summaries', () => {
   const expectation = { text: 'expected behavior', passed: false, evidence: 'not observed' };
   const cases = [
--- a/scripts/run-evals.js
+++ b/scripts/run-evals.js
@@ -439,12 +439,13 @@ function parseGrading(raw, expectedCount) {
   const expectations = g.expectations;
   const summary = g.summary;
   const passed = Array.isArray(expectations)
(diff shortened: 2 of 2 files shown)


abhisheksharma2411 (commented):
Ran it rather than read it, and the report is exact.

**The crash reproduces and your test catches it.** Reverting only `scripts/run-evals.js` and keeping the test:

```
✖ rejects null expectation entries without throwing
  TypeError: Cannot read properties of null (reading 'passed')
```

44/44 on the PR head.

**Both guards are load-bearing, which is worth knowing because they look redundant.** Reverting either one alone still fails the test:

| reverted | result |
|---|---|
| `expectation?.passed` in the count | 1 red |
| `expectation !== null` in the validator | 1 red |

They guard different statements — the `.filter()` runs before `ok` is computed, so the optional chain is what stops the throw, and the `!== null` is what makes the result `null` instead of a pass. Neither is decorative.

**And the fix is complete for the input this function can actually receive.** `parseGrading` JSON-parses its argument, and JSON has no `undefined` literal or sparse arrays, so `null` is the only entry shape that can reach here without being an object. I checked the other primitives land softly rather than by luck:

```
expectations: [1]    -> null
expectations: ["x"]  -> null
expectations: [[]]   -> null
```

Those survive because `typeof (1).text` is `undefined` rather than a throw — true, but true for a slightly non-obvious reason. If anyone later tightens that line to something like `expectation.text.length > 0`, the primitives start throwing again and only the `null` case is pinned. Not worth a change here; worth a sentence in the test if you're touching it anyway.

LGTM.

addyosmani (approved):
Robustness fix, rejecting null grader expectations instead of crashing, with a test. Verified the eval test passes. Merging.

PR #567, opened 2026-09-11
docs: compatibility guidance for downstream renames of browser-testing-with-devtools

### LINKED ISSUE #541: Document compatibility alias when renaming browser-testing-with-devtools

fin-coder07:
Adds a Compatibility Note to the browser-testing-with-devtools skill documenting the canonical skill name and requiring downstream catalogs that rename it (e.g. to playwright-cli) to either keep a lightweight compatibility alias or publish a migration rule, so consumers such as using-agent-skills and test-driven-development keep resolving.

A catalog can pass frontmatter validation while still directing agents to a missing skill after a silent downstream rename; a stable alias or documented migration prevents that without duplicating the implementation.

Closes #541.


abhisheksharma2411 (commented):
The failure mode is real and worth documenting — a catalog can pass frontmatter validation and still point agents at a name nothing resolves. I checked the claims rather than take them:

- `using-agent-skills` references this skill **2×**, `test-driven-development` **1×**. Both real.
- `validate-skills.js --strict`: 25 skills, 0 errors, 0 warnings.

## One thing that would make the migration option actionable

The note names two consumers, and the migration bullet then says to update *"`using-agent-skills`, `test-driven-development`, the setup docs, and any commands/agents that mention this skill."* That's correct but open-ended, and the real footprint is bigger than the two named:

```
$ grep -rl "browser-testing-with-devtools" --include="*.md" . | grep -v "skills/browser-testing-with-devtools/"
README.md
CLAUDE.md
.claude/commands/test.md
agents/web-performance-auditor.md
docs/adoption-guide.md
docs/cursor-setup.md
docs/gemini-cli-setup.md
skills/using-agent-skills/SKILL.md
skills/test-driven-development/SKILL.md

9 files, 12 mentions
```

Someone following the migration path will read "and any commands/agents that mention this skill" and have to invent that search themselves — and the two files most likely to be forgotten are `.claude/commands/test.md` and `agents/web-performance-auditor.md`, because neither is a skill or a setup doc.

Putting the command in the note instead of the list would make it self-maintaining too: the count goes stale the moment someone adds a reference, but `grep -rl` doesn't. Something like:

> Find every consumer before renaming: `grep -rl "browser-testing-with-devtools" --include="*.md" .` — at the time of writing, 9 files outside the skill's own directory.

Your call — the note is useful as it stands and this only makes the second bullet easier to execute.

## Minor

"Do not rename this skill silently" is doing a lot of work as the closing line and I think it's the strongest sentence in the note. It might be worth leading with it rather than closing on it, since a downstream maintainer skimming for the rule will hit the two options first and may not read to the bottom.

addyosmani:
Closing in favor of #552, which documents the same point (the name is a stable compatibility identifier) in skill-anatomy, the better home for a naming convention than a per-skill note. If there's anything unique here, fold it into that one. Thanks.

PR #568, opened 2026-09-14
fix(validator): strip every CommonMark fence form from linted prose (#437)

### LINKED ISSUE #437: bug(validator): stripFencedCodeBlocks misses tilde/indented fences and mismatched closer length

ayobamiseun:
## Summary

Fixes #437. `stripFencedCodeBlocks` in `scripts/lib/skill-lint.js` matched fences with one whole-document regex that only recognised a column-zero backtick fence closed by a run of exactly the same length. Three CommonMark-legal forms slipped through, so their contents were linted as prose:

- tilde fences (`~~~`)
- fences indented one to three spaces

This is live on `main`, not hypothetical: `skills/spec-driven-development/SKILL.md` nests its Commands and Project Structure examples inside list items with three-space-indented fences (lines 89–104). The old stripper leaves all three example blocks in the prose the required-section and cross-reference rules read; the new one leaves none.

## Change

- Replace the regex with a line scanner that tracks the opener's marker and length, closes only on a same-marker run at least as long, and treats an unterminated fence as running to end of file so the sections after it fail loud instead of passing silently.
- Follow the CommonMark rule that a backtick fence's info string may not contain backticks, so a prose line such as ```` ```js``` is inline code ```` is not mistaken for an opener.
- Replace stripped lines with empty lines so line numbers are preserved for any future rule that reports positions.
- Add 19 unit tests covering every form above, the mixed-marker and shorter-run cases, four-space indentation (an indented code block, not a fence), CRLF input, and the unterminated case.
- Run `scripts/lib/skill-lint-test.js` in CI. It was the only `*-test.js` file the workflow did not execute, so the existing linter tests have never gated a merge.

## Verification

```
node --test scripts/lib/skill-lint-test.js   # 27 pass, 0 fail
node scripts/validate-skills.js              # 25 skills, 0 errors, 0 warnings
node scripts/run-evals.js --min-rank1 95     # 140 checks, rank-1 100%
node scripts/validate-reference-links.js     # 25 skills, 0 errors
node scripts/validate-commands.js            # 9 commands, 0 errors
node --test scripts/*-test.js                # 36 pass, 0 fail
```

## Credit

This revives the approach from #444 by @ujjwalks, which was approved and then closed after a rebase conflict with #505. The scanner here is the same shape; the additions are the info-string rule, line-number preservation, and the extra edge-case tests.


abhisheksharma2411:
Verified rather than read, and it holds up on the boundaries the description doesn't mention.

**27 pass on the head. Reverting only `skill-lint.js` and keeping the tests fails 6** — the four forms you name plus the two extras:

```
✖ a heading inside a tilde fence does not satisfy the check
✖ a heading inside a fence indented one space
✖ a heading inside a fence indented three spaces
✖ a heading inside a fence with a longer closer
✖ a backtick run does not close a tilde fence, and vice versa
✖ an unterminated fence swallows everything after it and fails loud
```

Every one of them red before, green after. That's the shape that makes a stripper fix reviewable.

## The boundary I went looking for, and you'd already handled

`^ {0,3}` is the load-bearing part and it's easy to write as `^ *`. Probed directly:

| input | result |
|---|---|
| fence indented 3 spaces | stripped |
| fence indented **4** spaces | **kept** |
| closer indented more than its opener | stripped |
| closer with trailing spaces | stripped |

Four spaces is an indented code block in CommonMark, not a fence, so keeping it is correct — and it's the case where an over-eager `^ *` would silently swallow real prose and make a missing-section error disappear. Allowing the closer its own 0–3 indent independent of the opener is right too.

The info-string rule at `:117` is the one I'd have missed writing this myself:

```js
// CommonMark: the info string of a backtick fence may not contain backticks,
// so a line like ````js``` is inline code` is prose, not an opener.
```

That's a genuine spec detail, it's cited, and it prevents a line of prose containing inline code from opening a fence that then eats the rest of the document.

## On the test design

Observing stripping *through* the required-section rule rather than asserting on `stripFencedCodeBlocks`' output directly is the right call — it tests the property the linter actually depends on, and it keeps working if the stripper's return shape ever changes. The comment explaining that inversion ("the presence of the error proves the block was stripped") is what makes the tests readable; without it the assertions look backwards.

LGTM.

abhisheksharma2411 (commented):
The line scanner is the right shape, and the details that are easy to miss are already handled: closing only on the same marker at least as long as the opener, the info-string rule that a backtick fence's info may not contain backticks, unterminated fences running to EOF, and blanking lines rather than deleting them so positions stay valid. I probed it 18 ways in both directions and couldn't break it.

**Nothing leaks** — heading hidden inside the fence, checked via the required-section rule:

```
backtick basic · tilde · 3-space indent · longer closer · tilde info with
backticks · unterminated · closer with trailing spaces · marker mismatch · CRLF
  -> all stripped
```

**And nothing is over-stripped** — heading *outside* the fence must survive:

```
after a closed backtick/tilde/indented/longer-closer fence · prose with inline
```code``` · a ```js``` line · tilde closer for a backtick opener · 3-tick line
inside a 4-tick block
  -> all survived
```

**One thing worth adding to the PR description, because it's stronger than what's written there.** "Their contents were linted as prose" undersells it — the failure direction is a **false negative in a gate**. A skill genuinely missing a required section passes validation if its example text happens to contain that heading:

```
                      main    this PR
tilde fence            NO      YES      <- "missing ## Red Flags" reported?
longer closer          NO      YES
backtick (control)     YES     YES
```

On `main` the example silences the check. That's the argument for landing this.

**One precision note on the `spec-driven-development` evidence.** The three-space fences at L89–104 are real and are being mis-stripped, but the consequence there is currently **latent**, not live: the content inside those fences is itself indented, so it never matches the column-zero heading rule. `node scripts/validate-skills.js` gives `25 skills checked — 0 error(s), 0 warning(s)` on both `main` and this branch. The bug bites when fenced content sits at column zero inside a tilde or longer-closer fence, which is what the table above uses. Worth saying "latent on main today, and here is the shape that makes it bite" rather than "live" — it's a fix that prevents a false negative, and that's a good enough reason on its own.

**Disclosure on the workflow step:** your `Test skill lint rules` step is the same three lines I added in #577 (mine says `Test skill validator rules`). Yours is the earlier PR, so it should keep it — I'll rebase mine onto this and drop the duplicate rather than leave @addyosmani with a conflict. Entirely my miss for not checking open PRs by the files they touch before opening mine.

addyosmani (approved):
This is the right fix for #437: a proper CommonMark scanner instead of the single regex, and it also wires skill-lint-test.js into CI, which wasn't running before. Passes clean on all 25 skills. Merging.

PR #570, opened 2026-09-14
docs(readme): the catalog heading says 24, the section lists 25

abhisheksharma2411:
`## All 24 Skills` is contradicted by the sentence immediately beneath it — *"The pack includes 25 skills total"* — and by what the section actually contains.

Counted rather than eyeballed:

```
skills listed under the heading   25
directories under skills/         25
validate-skills --strict          25 skills checked
set difference                    (empty)
```

The section lists every skill on disk, meta-skill included, in its own `### Meta` subsection. So the heading is counting the same 25 the paragraph is.

**The nearby phrasing is correct and I left it alone:**

> The pack includes 25 skills total — 24 lifecycle skills plus the `using-agent-skills` meta-skill.

24 is a real number there, just not of the thing the heading counts. That distinction is why this looked deliberate at first — I checked the table contents before touching it rather than assuming the heading was simply stale.

## The anchor would have broken

Renaming a heading changes its anchor, and one page links to this one:

```
docs/adoption-guide.md:5
  [skill catalog in the README](../README.md#all-24-skills)
```

Retargeted to `#all-25-skills`. That's the only reference anywhere in the repo — checked across `*.md`, `*.json`, `*.ts`, `*.js` — so the rename is complete rather than half-done.

## Gates

- `node scripts/validate-skills.js --strict` → 25 skills checked, 0 errors, 0 warnings
- `node --test scripts/*-test.js scripts/lib/*-test.js` → 44 pass, 0 fail


addyosmani (approved):
Thanks, correct fix. The body already said 25, it was just the heading and the anchor that lagged. Merging.

PR #571, opened 2026-09-14
feat(validator): enforce the skill-anatomy layout rules

abhisheksharma2411:
`docs/skill-anatomy.md` states three structural rules that nothing checks. `scripts/lib/skill-lint.js` describes itself as *"the single source of truth for what makes a SKILL.md valid (docs/skill-anatomy.md)"*, so these belong there rather than in a sixth validator script.

| Rule | Where it's written | Enforced before |
|---|---|---|
| `SKILL.md` under 500 lines | Context Efficiency | no |
| no empty `scripts/` directory "just to mirror other skills" | Supporting Files | no |
| supporting files are `lowercase-hyphen-separated.md` | Naming Conventions | no |

### One live finding

```
25 skills checked — 0 error(s), 1 warning(s) — PASSED WITH WARNINGS
  ⚠  security-and-hardening
       WARN:  SKILL.md is 524 lines, over the 500-line context budget
              (skill-anatomy.md: Context Efficiency) — move detail into a supporting file
```

That's the only one. Every skill passes the layout rules today.

### Warning vs error

The line count is a **warning**. Trimming a skill is an editorial call, and failing CI on line count would block unrelated work on an over-budget skill — so this flags `security-and-hardening` without turning main red, and whether to split it stays yours. The layout rules are **errors**: nothing violates them today, so they can only fire on something newly added.

Counted as newlines, so the number matches what `wc -l` reports — that's how a contributor would check by hand. `> 500`, not `>=`; a file sitting exactly on the budget is fine, and there's a test pinning that boundary.

### Scope notes

- The empty-directory rule is written about `scripts/`; I generalized it to any subdirectory, since an empty `references/` is noise for the same reason. It names the outermost empty directory once rather than every nested level.
- The naming rule is scoped to `.md`, which is what that line of Naming Conventions covers. `scripts/` helpers have their own conventions under Script Requirements, so `Idea_Refine.sh` is deliberately not flagged here.
- `skills/constraint-driven-development/references/floor-guard.md` stays put. Nesting a `references/` dir inside a skill is sanctioned by `skill-anatomy.md` (Naming Conventions) and `CLAUDE.md:34`, even though `CONTRIBUTING.md` reads the other way — I've left that alone; it's a separate call and I didn't want to bury it in a validator PR.

### `skill-lint-test.js` has never run in CI

Every other `scripts/*-test.js` has a step in `test-plugin-install.yml`. `scripts/lib/skill-lint-test.js` doesn't, so the rules in the file the repo calls its source of truth were the only ones with unverified tests. Added the step.

### Verification

Full workflow suite, locally, on this branch:

```
validate-skills                 25 skills, 0 errors, 1 warning
skill-lint-test                 17 pass, 0 fail   (was 9, +8 new)
validate-versions               OK        / 1 pass
run-evals --min-rank1 95        140 checks passed, rank-1 100% (88/88)
validate-reference-links        OK        / 7 pass
validate-commands               OK        / 6 pass
validate-artifact-paths         OK        / 6 pass
```

Each new guard was mutation-tested — I broke the rule and confirmed the test catches it, rather than trusting a green run:

| Mutation | Result |
|---|---|
| line-count branch never taken | 1 fail |
| boundary `>` → `>=` | 1 fail (the exactly-500 test) |
| `isEffectivelyEmpty` always false | 2 fail |
| kebab-case check removed | 2 fail |
| restored | 17 pass |


addyosmani (approved):
I like this, and I verified it runs clean on all 25 skills today (one non-blocking warning on security-and-hardening at 524 lines, which is fair). One coordination note: it conflicts with #568 on skill-lint-test.js since you both append test blocks, and it relies on #568 to actually wire that test file into CI. I'm landing #568 first, so please rebase on top of it and your new tests will run in CI too. Approving the direction.

addyosmani:
This is sound. I verified the layout validator passes on all 25 skills (0 errors), and the one budget warning was security-and-hardening, which I've now merged #579 to bring under 500. It just needs a mechanical rebase: #568 landed its fence tests at the same spot in skill-lint-test.js where you append yours, so it's a keep-both conflict on that one file. Rebase to union the two test blocks and I'll merge.

abhisheksharma2411:
Unblocked — this was approved but sitting on a conflict, so it could not merge. Merged `main` in (`f533481`).

The only conflict was `scripts/lib/skill-lint-test.js`: #579 added the fenced-block stripping suite while this branch added the layout and context-budget suites, both at the end of the file. Resolved as a union.

Verified by **test name** rather than by reading the diff: all 15 from `main` and all 17 here survive, 24 total after the 8 shared with the merge base. Worth mentioning because my first attempt spliced the two hunks naively and produced a file that failed with `SyntaxError: Unexpected end of input` — the hunk boundaries did not line up with block boundaries, and a quick eyeball would not have caught it.

`node --test scripts/lib/skill-lint-test.js` → **36 pass / 0 fail**, and all five validators exit 0.

No changes to the reviewed content; `#512` is unblocked the same way if it is useful to take them together.

abhisheksharma2411:
Rebased onto `main` (`d6c11c4`). The union you described is done, and one commit dropped out on the way.

**`skill-lint-test.js` — kept both blocks.** #568's fence-stripping and YAML-validity tests, then my layout tests, in that order. Worth flagging one detail in case it bites the next person: the conflict region sat *inside* a shared trailing `});`, so a naive keep-both leaves the union one closer short — #568's last test ends up unterminated and the file dies with `SyntaxError: Unexpected end of input` rather than with a failing test. #568's block needed its own closer added, and the shared one goes on to close mine. **44 tests, 0 fail** after that.

**`test-plugin-install.yml` — took yours.** #568 renamed the step to "Test skill lint rules"; my branch had "Test skill linter rules". No reason for mine to win a rename it wasn't about.

**One commit skipped, not resolved.** `a302283 chore: drop the CI step, #568 adds it first` existed only to avoid three PRs carrying the same three lines. #568 has landed and `main` carries the step, so applying that commit now would delete *your* step rather than deduplicate mine. Dropped it. The branch is one commit — the layout rules themselves — and the consequence I named when I removed the step ("until #568 lands, these tests do not run in CI") no longer applies: they run.

**Re-verified against the rebased tree rather than assuming.** `node scripts/validate-skills.js` → **25 skills, 0 errors, 0 warnings**, which matches what you saw after #579.

A green run on a validator that five PRs have refactored under me doesn't prove my rules are still wired in, so I seeded a defect per rule against `test-driven-development` and confirmed each still reports:

| Seeded | Reported |
|---|---|
| SKILL.md pushed to 1,098 lines | `⚠ SKILL.md is 1098 lines, over the 500-line context budget` |
| empty `scripts/` (and one nesting only empty dirs) | `✗` error |
| `references/Bad_Name.md` | `✗` error |

The budget one is a warning and exits 0 by design — my first probe script treated exit 0 as "rule didn't fire" and wrongly reported it dead. Mentioning it because it's the kind of thing that looks like a regression report if I'd sent it without checking.

Ready when you are.

PR #573, opened 2026-09-15
test: make reference-link assertion portable on Windows

### LINKED ISSUE #572: test: make reference-link validator assertions portable on Windows
On Windows, scripts/validate-reference-links-test.js fails on the default branch because the validator formats path.relative() with backslashes while the regression assertion only accepts forward slashes. Reproduction: run node --test scripts/validate-reference-links-test.js on Windows; the test 'fails when a skill links the shared checklist as if it were colocated' reports 1 failure at line 75. Update the assertion to accept platform-native separators (or normalize the reported path) while preserving the same failure check. The validator behavior is correct; this is a cross-platform test portability defect.

dajiaohuang:
Fixes #572. Make the reference-link regression assertion accept both POSIX and Windows path separators. The validator behavior is unchanged. Tests: node --test scripts/validate-reference-links-test.js.


abhisheksharma2411 (commented):
Correct fix for the reported failure, and the regex is right — `[\\/]` on each separator, with the `.md` still escaped and the rest of the assertion untouched. It'll pass on both platforms.

One thing worth weighing before this lands, because #572 explicitly offers both routes ("update the assertion to accept platform-native separators **or normalize the reported path**") and I think the second is slightly better here.

**The validator's output is itself inconsistent on Windows.** `validate-reference-links.js` prints two kinds of path in the same failure block:

```js
81:  console.log(`  ✓  skills/${name}/SKILL.md`);          // hardcoded  /
83:  console.log(`  ✗  skills/${name}/SKILL.md`);          // hardcoded  /
85:  const resolved = path.relative(ROOT, path.resolve(skillDir, link));
86:  console.log(`  L${line}: ${link} — resolves to ${resolved}, ...`);   // native  \
```

So a Windows user reading a failure sees `skills/using-agent-skills/SKILL.md` on one line and `skills\using-agent-skills\references\definition-of-done.md` on the next — for the same skill, three lines apart. Teaching the test to accept both separators makes the test pass without touching that.

Normalising `resolved` at line 85 fixes both at once:

```js
const resolved = path.relative(ROOT, path.resolve(skillDir, link)).split(path.sep).join('/');
```

The assertion then stays exactly as it was on `main` — no `[\\/]`, nothing for a future reader to decode — and the output matches the `/`-shaped paths the surrounding lines already print, the `references/<file>.md` remedy text in the same message, and the repo's own docs.

`scripts/validate-reference-links.js` is the only script with this split — `run-evals.js` uses `path.relative` too but never hardcodes a slash alongside it, so nothing else needs the same treatment.

Not a blocker, and I'd rather see this merged than stall on style: the test is red on Windows today either way, and your version fixes that. But if you'd like to swap it, the one-line normalisation is strictly smaller than the regex change and removes the reason the assertion needed loosening in the first place.

(Can't approve — not a maintainer — just a reviewer note.)

dajiaohuang:
Implemented the suggested path normalization in ef66f42. validate-reference-links.js now renders path.relative(...) with forward slashes, matching the validator's other output and keeping the test assertion platform-independent. Verified with node --test scripts/validate-reference-links-test.js (7/7 passing) and git diff --check.

abhisheksharma2411 (commented):
`ef66f42` does exactly what #572's second route asks for, and it's the better of the two — thanks for taking it. The output is now internally consistent: all three spellings in a failure block agree, and the assertion can keep matching one exact string instead of `[\\/]`.

**I'm closing my #575 in favour of this.** It changes the same line in the same file for the same reason, and you got there first once the suggestion was on the table; no reason to make @addyosmani pick between two identical fixes.

One thing worth taking from mine before you land, and it's the only real difference:

```js
.split(path.sep).join('/')     // this PR
p.replace(/\\/g, '/')          // #575
```

Both are correct on both platforms — `path.relative` never returns a backslash on POSIX, so there's nothing to convert there. The difference is **testability**: on a Linux runner `path.sep === '/'`, so `split(path.sep).join('/')` is a literal no-op, and **CI cannot exercise the conversion at all**. If someone later refactors that line wrong, the regression ships green because the only platform that would catch it isn't in CI. The `replace(/\\/g, '/')` form does the same work but is exercisable everywhere — you can feed it a backslash string in a unit test on Linux and assert the output.

Not a correctness objection, and I wouldn't hold the PR for it. But this repo's whole validator suite exists to catch exactly this class of "green on CI, broken on the other platform" problem, which is what #572 was in the first place.

Verified `7/7` on `node --test scripts/validate-reference-links-test.js`.

addyosmani (approved):
Good catch. This just normalizes the displayed path so the existing assertion holds on Windows, no behavior change. Merging.

PR #574, opened 2026-09-15
fix(hooks): stop auto-injecting the meta-skill on SessionStart

### LINKED ISSUE #569: Claude Code plugin still injects using-agent-skills on every SessionStart, which is the double routing #557 tells users to avoid
#557 (merged 12 Sep) updated docs/getting-started.md to say:

> If your host already discovers and activates skills from their descriptions, do not also paste `using-agent-skills` into an always-on system prompt or rules file. That creates two routers for the same task. Install the individual skills and let the host activate them on demand instead.

Claude Code is such a host. But the Claude Code plugin in this repo still does exactly that:

- `hooks/hooks.json` registers a SessionStart command hook that runs `hooks/session-start.sh`.
- `hooks/session-start.sh` reads `skills/using-agent-skills/SKILL.md` in full and returns it as `additionalContext`, with the comment "Injects the using-agent-skills meta-skill into every new session".

Per Claude Code's plugin reference, `hooks/hooks.json` is active whenever the plugin is enabled, so every Claude Code session with this plugin gets the meta-skill router on top of Claude Code's own skill routing.

So the manual and Codex setup paths now follow the merged guidance, and the Claude Code plugin path does not. Since #557 also closed #423 (the pack being slow on GPT-5.6), and the mechanism named there was procedural fragmentation costing turns, the plugin path may still carry the cost the fix was meant to remove.

Two possible shapes for a fix, whichever you prefer:

1. Remove the SessionStart hook from the Claude Code plugin, since Claude Code routes natively, and keep it for hosts that do not.
2. Make the hook host-aware: skip injection when it detects a natively routing host, inject otherwise.

Happy to open a PR for either. Scope here is only the mismatch between the merged guidance and the plugin's runtime behaviour; nothing about whether the meta-skill itself should change.

santhiprakash:
Fixes #569.

## Problem

`hooks/hooks.json` registers a `SessionStart` command hook that injects the full `skills/using-agent-skills/SKILL.md` as `additionalContext` on every session. Claude Code already discovers and activates skills from their `description` frontmatter, so the hook stacks a second router on top of the native one — the pattern `docs/getting-started.md` warns against ("do not also paste `using-agent-skills` into an always-on system prompt... That creates two routers for the same task"). #569 measured **8,088 vs 4,525 `cache_creation_input_tokens`** (~3.5k tokens wasted per session) under plugin v0.6.9.

## Root cause

`hooks/hooks.json` is active whenever the plugin is enabled, so every Claude Code session gets the meta-skill router regardless of need. The merged #557 guidance covered the manual and Codex setup paths but left the plugin path still auto-injecting.

## Fix

The issue's first suggested shape: remove the `SessionStart` hook from the Claude Code plugin, keep the script for hosts that don't route natively.

- Delete `hooks/hooks.json` — it registered only this hook, so the plugin now ships no hooks.
- Keep `hooks/session-start.sh` (and its regression test) as a standalone script for hosts without native skill routing; `docs/gemini-cli-setup.md` already tells users to wire such a hook themselves, and now points at the script directly.
- Update the hook's header comment, `CONTRIBUTING.md` ("Testing Hooks"), and `docs/codex-setup.md` so they no longer describe the hook as auto-registered for every Claude Code session.

Note on the neighbors: this intentionally leaves `hooks/session-start.sh` + `hooks/session-start-test.sh` in place, so #510 stays meaningful as-is. If this lands before #543, its `hooks.json` rewrite becomes moot (a delete/modify conflict will surface that explicitly rather than silently re-registering the hook); its `session-start.mjs` port could still be wired by users on hosts that need it.

## Verification

- `bash hooks/session-start.sh` still emits the standard `{"hookSpecificOutput": {"hookEventName": "SessionStart", ...}}` envelope including the meta-skill (verified by parsing stdout with node).
- `claude plugin validate .` → `✔ Validation passed` (2.1.234).
- `node scripts/validate-skills.js` → PASSED (25 skills, 0 errors).
- `node scripts/validate-commands.js` → PASSED (9 commands).
- `node scripts/validate-versions.js` → PASSED.
- `node scripts/run-evals.js --min-rank1 95` → PASSED (140 checks, rank-1 100%).
- `bash hooks/session-start-test.sh` → fails on `main` *and* on this branch with the same pre-existing stale-assertion error (`expected IMPORTANT priority, got undefined`) — the envelope assertions are fixed by open #510; this change doesn't alter the script's output.

## Risks

- Users who relied on the auto-injection lose the always-on meta-skill; individual skills keep working via Claude Code's native routing, and the script remains for manual wiring.
- Deleting `hooks/hooks.json` means a future plugin hook re-adds the file from scratch.

AI-assisted. I wrote and verified this change.


addyosmani:
This is the fix for #569, and it points the same way the docs already do: getting-started.md warns against double-routing, and #557 made that official. It also closes the Windows failures in #488 and #475, since the failing hook goes away. I'm intentionally not merging it this second because it's mutually exclusive with #543 (which keeps the hook and ports it to Node) and it changes what the plugin does out of the box. I want to make that call deliberately. Holding for a moment, leaning toward this.

addyosmani (approved):
Going with this. It's consistent with getting-started.md, it fixes the Windows failures in #488 and #475 by removing the hook that was breaking, and it reclaims the per-session tokens #569 measured. The standalone script stays for hosts without native routing. Closing #543 in favor of this, with thanks. Merging.

PR #575, opened 2026-09-15
fix(validator): report reference-link paths with `/` on every platform

abhisheksharma2411:
Refs #572. **Complements #573 rather than competing with it** — different file, no conflict, and they can land in either order.

## The problem

`validate-reference-links.js` prints three spellings of the same separator inside a single failure block. Two lines hardcode a forward slash, the line between them doesn't:

```js
81:  console.log(`  ✓  skills/${name}/SKILL.md`);                    // hardcoded  /
83:  console.log(`  ✗  skills/${name}/SKILL.md`);                    // hardcoded  /
85:  const resolved = path.relative(ROOT, path.resolve(skillDir, link));   // native  \  on Windows
86:  console.log(`  L${line}: ${link} — resolves to ${resolved}, ...`);
```

So a Windows user reading one failure sees:

```
  ✗  skills/using-agent-skills/SKILL.md
       L1: references/definition-of-done.md — resolves to skills\using-agent-skills\references\definition-of-done.md
...
use `../../references/<file>.md`, not `references/<file>.md`.
```

Three separator styles, for the same skill, within four lines — and the remedy the message tells you to type is the `/` one.

## Why this way

#572 offers two routes: *"update the assertion to accept platform-native separators (or normalize the reported path)"*. #573 takes the first; this takes the second. They're independent — #573 edits `validate-reference-links-test.js`, this edits `validate-reference-links.js`.

Normalising has a side benefit the test-side fix doesn't: it makes the **output** consistent, not just the test. It also means the regression assertion can go on matching one exact string, so a future reader doesn't have to decode `[\\/]` to see what's being asserted.

`validate-reference-links.js` is the only script with this split. `run-evals.js` uses `path.relative` too, but never hardcodes a slash beside it.

## One deliberate choice

```js
return p.replace(/\\/g, '/');     // not p.split(path.sep).join('/')
```

`split(path.sep)` is the more idiomatic form, but on a POSIX runner `path.sep` is `/`, so it's a **no-op there** — a regression would ship green through this repo's CI, which is Linux. The replace form does the same work and is exercisable everywhere.

The trade-off is that a literal backslash inside a path component would be rewritten. That isn't a case this repo has: every link this function reports is matched by `REFERENCE_LINK_RE` as `references/<name>.md` with `[A-Za-z0-9._-]` names, so a backslash can't reach it.

## Verification

```
validate-reference-links        25 skills checked — 0 error(s) — PASSED
validate-reference-links-test   7 pass, 0 fail
validate-skills                 OK
validate-versions               OK
validate-commands               OK
validate-artifact-paths         OK
run-evals --min-rank1 95        PASSED, rank-1 100% (88/88)
```

Normaliser behaviour, stated plainly since CI here can't observe the Windows path:

```
windows-shaped in : skills\using-agent-skills\references\definition-of-done.md
normalised out    : skills/using-agent-skills/references/definition-of-done.md
posix unchanged   : skills/using-agent-skills/references/definition-of-done.md
```

**Honest limit:** the existing regression test is the guard, and it only bites on Windows — on Linux CI it passes with or without this change, because there's no backslash to normalise. I didn't add a unit test for the helper because that would mean exporting it and guarding `main()`, and no sibling validator in `scripts/` does either; they're all unguarded CLIs tested through `spawnSync`. Happy to add one if you'd rather break that convention.


abhisheksharma2411:
Closing in favour of #573, which now fixes the same line in the same file.

#572 offered two routes — loosen the test assertion, or normalize the reported path. #573 originally took the first; I opened this for the second and said on #573 why I thought normalizing was better. @dajiaohuang agreed and implemented it in `ef66f42`, so this PR is now a duplicate of theirs. They did the work once the suggestion was on the table, and there's no reason to make @addyosmani choose between two identical fixes.

Left the one substantive difference on #573 for them to take or leave: `split(path.sep).join('/')` is a no-op on a POSIX runner, so CI can never exercise the conversion, whereas `replace(/\\/g, '/')` does the same work and is testable on every platform. Minor, and not worth a competing PR.

PR #576, opened 2026-09-16
fix(evals): bind grader results to declared expectations by id

mavericksea-ai:
`parseGrading` (lines 429 to 458 of `scripts/run-evals.js` at be4e44a) validated only that the number of grader results matched the expected count, without binding each result to its declared expectation. A grader response could duplicate one expectation, omit another, include results for undeclared expectations, or supply a fabricated `pass_rate`, and still pass validation.

This change requires every grader result to carry an integer `id` in `1..n` that maps back to the numbered expectations in the prompt, rejects duplicates and out-of-range ids, enforces that every declared expectation is graded exactly once, replaces each result's `text` with the declared expectation so reports carry the canonical wording, and derives `pass_rate` from the validated counters rather than trusting or comparing against the grader's value.

### Prior fixes in the same validator

- #473
- #563

### New tests

1. Reordered but complete results are accepted.
2. Duplicate results (the same expectation graded twice, another omitted) are rejected.
3. Results whose ids are not in the declared set are rejected.
4. A result set that omits a declared expectation is rejected.
5. `summary.pass_rate` is derived from the counters: a grader value of 0.999 with correct counters is accepted and rewritten to the derived value.
6. Paraphrased grader `text` is replaced by the declared expectation.

The new tests fail at be4e44a and pass with this change; the existing tests still pass (22 in total).

### Prompt change

The grader return-schema instruction now includes `"id": integer` and a sentence requiring each id to match the expectation number.

### Changes

- `scripts/run-evals.js`: `parseGrading` signature changed from `(raw, expectedCount)` to `(raw, expectations)`, where `expectations` is the string array; id binding and duplicate and range checks added; result `text` canonicalised from the declared expectation; `pass_rate` derived from the validated counters, with `passed`, `failed` and `total` still required to match exactly; grader prompt updated; call site updated.
- `scripts/run-evals-test.js`: existing `parseGrading` tests updated for the new signature; six new tests added.


addyosmani (approved):
Good. Binding each grader result to a declared expectation by id, and overwriting the grader's paraphrased text with the canonical one, is stricter than the old length-only check and closes a real trust gap. Verified the eval tests pass. Merging.

PR #577, opened 2026-09-16
feat(validator): enforce the SKILL.md line budget, and run the linter's tests in CI

abhisheksharma2411:
Refs #137 — the formatting-specific half you kept open, taken from the angle that has evidence behind it.

## The rule already exists; nothing checks it

`docs/skill-anatomy.md` → **Context Efficiency**:

> **Keep `SKILL.md` under 500 lines.** Move detailed reference material into supporting files.

That is mechanical, and it is already violated on `main`:

```
⚠  security-and-hardening
     WARN:  SKILL.md is 524 lines, over the 500-line budget
            (skill-anatomy.md: Context Efficiency) — move detailed
            reference material into supporting files

25 skills checked — 0 error(s), 1 warning(s) — PASSED WITH WARNINGS
```

## Why a warning, not an error

Context Efficiency sits under **Recommended**, not Required, in that document's own `## Required vs Recommended` split. So it reports like the existing dead-cross-reference check: `validate-skills.js` still exits `0`, CI stays green, and the drift is visible. Turning it into an error is a policy call that belongs to you, not to the patch that adds the check.

## I did not fold in the content fix

The obvious remedy for `security-and-hardening` is its inline 45-line `## Security Review Checklist` (L427–471) — the skill already links `../../references/security-checklist.md` **four times**, including from its own See Also: *"For detailed security checklists and pre-commit verification steps, see …"*. Whether those two are genuinely duplicative or deliberately complementary (a 45-line tick-list vs a 245-line guide with examples) is an editorial judgement about that skill, and I did not want to smuggle it into a linter PR. Happy to do it as a follow-up if you want it.

## The second half: the linter's tests never ran

Every other validator in `test-plugin-install.yml` has a paired test step — `validate-versions`, `validate-reference-links`, `validate-commands`, `validate-artifact-paths`, `run-evals`. `validate-skills.js` is the only one without, and `scripts/lib/skill-lint-test.js` was the only file in the repo that referenced itself:

```
$ grep -rn "skill-lint-test" .          # before this PR
(no matches outside the file itself)
```

There is no `package.json`, so there was no other runner either. Its **8 existing tests had never run on a pull request** — and neither would the 4 added here. One step added after "Validate all skills".

## What I checked

- **Drift I looked for and did NOT find**, so this PR does not claim it: bullet-marker mix is one file (`documentation-and-adrs`, and those are inside fenced blocks), and frontmatter field order is already uniform at 25/25 `name:` → `description:`. Your instinct in #137 that full autoformatting doesn't earn its tooling matches what's actually in the tree — this is the one Context Efficiency rule with a live violation, so it's the only one I implemented.
- **Off-by-one**: "under 500" is enforced as "not more than 500". 500 is clean, 501 warns.
- **Trailing newline**: a file ending in `\n` has no phantom final line — counting one would report a file exactly at budget as over it. Pinned by its own test.
- **Mutation-tested**, 5/5 killed: `>` → `>=`; count the trailing-newline line; `warnings.push` → `errors.push`; budget 500 → 5000; delete the check.
- **All 11 CI gates run locally**: 5 validators exit 0, 6 test files `48 pass / 0 fail`.


abhisheksharma2411:
Dropped the CI step from this PR in `d9bd48a` — **#568 adds the identical three lines** and is the older PR, so it should carry them. Keeping a duplicate would only have handed you a conflict on the same hunk.

That was my miss: I opened this claiming `skill-lint-test.js` "had never run on a pull request" and presented adding the step as the finding. The repo state was right, but #568 already had the fix open and I didn't check open PRs by the files they touch before opening mine. Reviewed #568 properly and said the same there.

**Consequence worth stating plainly: until #568 lands, the 4 tests added here don't run in CI.** They pass locally (`12 pass / 0 fail`). If you'd rather this PR stayed self-contained, say so and I'll put the step back and rebase on whichever of the two lands first.

What remains here is the part that's actually its own: the Context Efficiency line-budget check, and the one live violation it finds (`security-and-hardening/SKILL.md` at 524 lines against the documented 500).

abhisheksharma2411:
Closing this as a duplicate — **of my own #571**, which I opened two days earlier and then failed to check before opening this.

#571 already carries the 500-line context-budget rule, plus two more skill-anatomy rules this PR doesn't have (no empty `scripts/` directory, supporting-file naming). It also flags the same live violation, `security-and-hardening/SKILL.md` at 524 lines. So this is a strict subset.

Its line counting is also the better of the two: #571 counts newlines deliberately so the number matches `wc -l`, which is how a contributor would check by hand. Mine counted the final unterminated line instead — defensible in isolation, worse as a number you're asked to reproduce.

Nothing here is lost except four tests (at-budget boundary, trailing newline, budget-independent-of-other-checks, and the mutation battery behind them). If #571's coverage looks thin on the boundary cases when you get to it, say so and I'll port those across rather than reopen this.

The CI step is separately redundant: #568 adds the identical `node --test scripts/lib/skill-lint-test.js` and is the older PR. I've removed it here and from #571 so it lands once, from there.

Apologies for the noise — this is the third time today I've been caught by not checking what was already open, and the first where the thing already open was mine.

PR #578, opened 2026-09-16
fix(hooks): simplify-ignore Stop expands the file on disk instead of restoring a stale backup

mavericksea-ai:
The `Stop` branch of `hooks/simplify-ignore.sh` (line 154 at be4e44a, `cat "$bak" > "$orig"`) restores every cached file from its backup unconditionally. That backup is written once on `PreToolUse Read` (line 202) and refreshed only by the `PostToolUse Edit|Write` branch (line 291), so any change that reaches a protected file through another route between `Read` and `Stop` is silently overwritten at session end: a `Bash` command, a formatter, a generator, an external editor.

Reported in #295 with a live reproduction on Claude Code 2.1.90, the three hooks from `SIMPLIFY-IGNORE.md` passed as `--settings`, and one file holding one protected block plus one unprotected line. The prompt routed the edit through `Bash` and forbade `Edit` and `Write`. With the hooks enabled, `Bash` wrote and read back `const editable = 2;`, the run finished clean, and the file on disk afterwards held `const editable = 1;`. With the hooks disabled, the same task left `const editable = 2;`. No crash, no failed tool call, no symlink, no path alias, no missing dependency: a normal `Stop` event.

### The intent this is measured against

The hook's own header states the contract: "The file on disk ALWAYS has placeholders while the session is active. The real content (with model's changes applied) lives in the backup." The protection is over the *content of the annotated block*, which is hidden from the model behind `BLOCK_<hash>` while the session runs, not over the file as a whole; `SIMPLIFY-IGNORE.md` describes the round trip as "expands placeholders back to real code, saves model's changes". The hook is not documented as reverting a protected file to its pre-session state, and it does not behave that way for `Edit` or `Write`. So the defect is real: `Stop` must not lose a change simply because that change arrived through a tool the hook does not listen to.

Refreshing the backup on every tool that can write does not fix this within the hook's design. `PostToolUse` for `Bash` carries `tool_input.command`, not `tool_input.file_path`, so there is nothing to key the cache on; the hook would have to re-expand and re-filter every cached file after every tool call, and a write from outside Claude Code (a watcher, an editor) still fires no event at all. The state the hook needs is already on disk.

### Fix

At `Stop`, run the same placeholder expansion the `Edit|Write` branch already runs, against the current contents of the file, rather than copying the backup over it. Whatever wrote the file between events wrote it in its filtered form, with the placeholders intact, so expanding it restores the protected blocks *and* keeps the change, from whichever route it came.

The expansion body moved out of the `Edit|Write` branch into an `expand_file src dest fid` function (same code, parameterised) and both branches now call it, so the fuzzy-match and deleted-block warnings apply identically at `Stop`. A small `has_placeholders` helper decides which path `Stop` takes. When no cached `BLOCK_<hash>` is left in the file, the file was rewritten wholesale and the hook can no longer tell where the protected block belongs. In that case it first copies the rewritten file to `.claude/.simplify-ignore-cache/<id>.recovered`, then restores the backup, and warns on stderr with both paths, so neither version is lost; the user merges by hand. That residual case is documented in `SIMPLIFY-IGNORE.md`'s known limitations, next to the existing rename limitation.

With no external write, expansion of the filtered file reproduces the backup exactly, so `Read → Stop` and `Read → Edit → Stop` are unchanged, byte for byte, including files with no trailing newline.

### Tests

Three tests added to `hooks/simplify-ignore-test.sh`, in the harness's existing style, driving the real hook end to end:

- Test 11: `Read`, a model edit plus the `Edit` event, then a write applied straight to the file with no event after it, then `Stop`. Asserts the direct write survives, the stale value is not restored, the earlier `Edit` still survives, the protected block is back, and no placeholder is left behind.
- Test 12: `Read`, a wholesale rewrite that removes every placeholder, then `Stop`. Asserts the backup is restored, `<id>.recovered` holds the rewritten content, and the warning names both paths.
- Test 13: the minimal case, `Read`, a direct write with no event, `Stop`. Asserts the write survives with the block back and no placeholder left.

The full-lifecycle parts are skipped, not failed, where `jq` is absent, matching how test 10 and `session-start-test.sh` handle that machine.

On be4e44a test 11 fails exactly as reported (`Bash write survives Stop: expected 1, actual 0`; `stale backup value not restored: expected 0, actual 1`) and its other assertions pass, which is what makes it a regression test for this bug rather than for the round trip generally.

```
bash hooks/simplify-ignore-test.sh   # 38 passed, 0 failed
```

The rest of the suite is unchanged: `node --test` over the `scripts/` test files is 36 passed, 0 failed; the five validators pass; `node scripts/run-evals.js --min-rank1 95` is 140 checks, rank-1 100%. `hooks/session-start-test.sh` fails on this machine at be4e44a as well, before any change here: `session-start.sh` no longer emits the `priority` field the test asserts. Observed, not changed in this PR.

CI execution for this harness is implemented in #501, which adds a `test-hooks` job running both hook suites, so this PR does not touch the workflow. If #501 does not land, I am happy to add the job here.

### Overlap with open work

- **#501** extends this same harness with three full-lifecycle tests (11 to 13). Test 11 here is #501's test 13 minus its `Edit` event, and I offered it there in #295. #501 changes only the test file and the workflow, not the hook, so the two do not conflict; if it lands first, these tests rebase onto its numbering and helpers.
- **#294** hardens the same hook (symlink guard, atomic rewrite, `perl` removal) but its `Stop` block still copies the backup, and its backup is still written only on `Read` and `Edit|Write`, so it does not reach this case. The two changes touch different lines of the same branch and are straightforward to order either way.

### Changes

- `hooks/simplify-ignore.sh`: expansion extracted into `expand_file`; new `has_placeholders`; the `Stop` branch expands the file on disk, and when no placeholder remains keeps the rewritten file in the cache as `<id>.recovered` before restoring the backup, with a warning naming both paths; header comment updated to describe `Stop` accurately.
- `hooks/SIMPLIFY-IGNORE.md`: `Stop` row of the event table updated; one known limitation added for the wholesale-rewrite fallback, with the recovered path.
- `hooks/simplify-ignore-test.sh`: tests 11 to 13 as above.


addyosmani:
Real fix, and the reasoning is right: on Stop we should expand the current file rather than blindly restore the backup. Verified the tests pass. Two things before it lands: it isn't wired into CI, and it overlaps #501 on simplify-ignore-test.sh. Plan is to land #501 (which adds the test-hooks job) and have this rebase on top, so the fix ships with CI coverage. Holding on that, not on the fix itself.

mavericksea-ai:
Understood, will rebase onto #501 once it lands.

addyosmani (approved):
Landing this so the data-loss fix ships in the next release. Verified the tests pass and expanding the current file on Stop rather than restoring the backup is the right behavior. #501 overlaps it on simplify-ignore-test.sh, so I'll have that rebase on top to add the CI wiring. Merging.

Issue #433, opened 2026-07-29
[Discussion] How the pack lands on Claude 5-generation models


lhoupert:
I've been using this pack for a few months (Claude Code, plus the Cursor links from `install.sh`). Last week I reviewed my whole `~/.claude` setup against Anthropic's recent article on context engineering for Claude 5 models (https://claude.com/blog/the-new-rules-of-context-engineering-for-claude-5-generation-models), and this pack ended up being the biggest single thing I removed. I'd rather report the mechanism than just quietly drop it, because maybe something  is fixable in the pack itself and not really about my setup

To be clear up front: I'm not saying the skills are wrong or that they don't help on ther models. I ran the review on one machine, and I don't yet have "before/after" data, I only removed things a few days ago. 

*The measurements below are from Claude against the repo at `main`*

> **Unconditional cost.** The 20 skill descriptions total ~4.7 KB, and they load into the system prompt of every session,  including sessions that never touch code. The bodies are ~207 KB in total.
>
> **Pre-empting model judgement.** 18 of the 20 carry a "Common Rationalizations" section, which exists to catch the model talking itself out of the prescribed process. The Anthropic article argues the opposite direction for this model generation — give rules, then let the model use judgement.
>
> **Overlap with host built-ins.** On current Claude Code, `code-review-and-quality`, `security-and-hardening`, `code-simplification` and `planning-and-task-breakdown`  each duplicate a built-in command (`/code-review`, `/security-review`, `/simplify`, `/plan`).

On the last point,  #423 reports `using-agent-skills` stacking on GPT-5.6's built-in router. The same thing happens on Claude Code, which also routes skills natively. So that's the same mechanism on a second host, which I think supports the "don't load our router where the host already has one" note you suggested there.

My question is whether this is worth handling in the pack rather than each user discovering it?  Would  versioning skills per model generation work ? (e.g. a leaner variant for Claude 5-class models that keeps the process content but drops the rationalization/red-flag  scaffolding.)

I asked more or less the same question about our own internal skills repo, so I'm genuinely unsure what the right answer is :-)

addyosmani:
This is exactly the kind of report I want, thanks for writing up the mechanism instead of just quietly dropping it. Let me take each piece honestly, including where I think the framing is a little off.

[...]

PR #579, opened 2026-09-18
refactor(security-and-hardening): move code patterns into a skill-local reference (-40% tokens per load)

Syamsuddin:
## Why

`security-and-hardening/SKILL.md` was the largest skill in the catalog: 524 lines / 4,096 words (~5.3k tokens), loaded in full every time the skill activates. About 30% of those words were inline TypeScript examples (OWASP patterns, zod schema, SSRF check, rate limiter, LLM output handling) that an agent only needs at the moment it writes that specific code.

[docs/skill-anatomy.md](https://github.com/addyosmani/agent-skills/blob/main/docs/skill-anatomy.md#context-efficiency) already sets the rule: keep `SKILL.md` under 500 lines, move reference material into supporting files, and prefer progressive disclosure because inline code is paid for on every load. This PR brings the skill in line with that rule without changing what it tells the agent to do.

## What changed

**`skills/security-and-hardening/SKILL.md`** — 524 → 216 lines, 4,096 → 2,449 words

- Frontmatter, Overview, When to Use, Process (threat model + STRIDE), Three-Tier Boundary System, Common Rationalizations, Red Flags, and Verification are **byte-identical** to `main`. No description edit, so routing is untouched.
- The middle (OWASP patterns, Input Validation, Rate Limiting, Secrets, Dependency triage, Privacy, LLM, Review Checklist) is replaced by one **Hardening Controls** section: each area is 1–3 lines of rule plus a deep link to its pattern. Every rule that was in the original prose is still stated here (parameterize, encode, authz-not-just-authn, ≥12 rounds, cookie flags, CSP/CORS, 422 + typed value, magic bytes, all-DNS-records SSRF check + TOCTOU caveat, three-part destructive-path check, shared-store rate limiting, rotate-then-purge, the four supply-chain steps, five privacy rules, LLM01/02/05/06/07/08/10).
- The 40-line review checklist duplicated `references/security-checklist.md`; it is now a one-paragraph pointer.

**`skills/security-and-hardening/references/hardening-patterns.md`** (new, 324 lines)

- 15 of the original 16 fenced blocks moved **verbatim** (assembled by line range with `sed`, then verified by script that each block string appears unchanged). The 16th is the review checklist replaced by the pointer above.
- Follows the existing precedent for single-skill reference material: `constraint-driven-development/references/floor-guard.md` (#472, merged), and skill-anatomy's "material used by exactly one skill is a supporting file inside that skill's directory".
- Links back to the shared checklist with the corrected `../../../references/` depth.

**`references/security-checklist.md`** — +3 lines

- Adds the three privacy items (classification/purpose, retention + deletion path, data-subject rights + consent) that only existed in the removed skill checklist, so nothing is lost.

## Verification

- `node scripts/validate-skills.js` — 25 skills, 0 errors, 0 warnings
- `node scripts/validate-reference-links.js` — pass
- `node scripts/run-evals.js --min-rank1 95` — rank-1 100% (88/88), no collision warnings
- All 15 `#anchor` links from SKILL.md resolve to headings in the new file; no unlinked headings in the new file
- The behavioral eval's expectations (trust boundaries first, SSRF covers private/reserved IPs, validation at the boundary, no secret leakage) are all still explicit in SKILL.md
- Pre-flight: no open upstream PR or issue touches this skill or `references/security-checklist.md`; `evals/skill-impact.md` has no rejected entry for it

## Scope

Deliberately one skill, content moved not rewritten, so the diff can be reviewed as a relocation. If the approach is welcome, the same treatment applies to `performance-optimization` (4.3k tokens) and `code-review-and-quality` (4.2k tokens) as follow-up PRs.


Syamsuddin:
Related: #571 adds a `> 500 lines` warning to `skill-lint.js`, and its one live finding is this skill at 524 lines. This PR brings it to 216, so once both land the warning has nothing to report. The two are independent and can merge in either order.

abhisheksharma2411 (commented):
This is the right fix for the largest skill in the catalog, and I checked the thing that matters most for a 394-line deletion — **whether anything was lost**. Nothing was.

```
skills/security-and-hardening/SKILL.md      524 → 216 lines
new skill-local reference                         324 lines
validate-skills.js                          25 skills, 0 errors, 0 warnings
validate-reference-links.js                 25 skills, 0 errors
```

**Every section survives; they were renamed and regrouped, not dropped.** I diffed the heading sets and chased each of the 14 that looked absent — `Data Privacy & Compliance` → `Personal data and privacy` + `Data Classification`, `Securing AI / LLM Features` → `AI / LLM features` + `LLM Output Handling`, `Triaging Dependency Audit Results` → `Dependencies and supply chain` + `Dependency Audit Triage`, and so on. The 37 → 40 heading count is reorganisation, not loss.

**The one that needed real checking was the checklist**, because it goes from 28 inline items to a single pointer. I walked all 28 against `references/security-checklist.md` and every one has a counterpart — "Users can only access their own resources" → `Every resource access checks ownership/role (prevents IDOR)` (L49), the reachability/install-script items → L178 and L186, error-internals → the Error Handling section, and so on. The shared file carries 49 items, so this is strictly more coverage than the inline copy, not less.

**And the three items you added to the shared checklist are exactly the ones that had no counterpart** — data classification/purpose/minimisation, retention plus a working deletion path including backups and indexes, and data-subject export/delete with third-party consent. That's the part that makes this a safe deduplication rather than a lossy one, and it's easy to skip.

Two things I'd call out as done right:

**The reference is skill-local.** `skills/security-and-hardening/references/hardening-patterns.md`, not repo-root `references/` — matching `skill-anatomy.md` L125 ("material used by exactly one skill is a supporting file inside that skill's directory"). Nothing else in `skills/` links it, so that's the correct side of the rule.

**The inline checklist was duplicating the shared one.** That duplication is the underlying problem, and the two copies had already drifted — the inline version was missing the privacy items the shared one now carries. Replacing a drifting copy with a pointer is better than keeping both in sync by hand.

For what it's worth, this closes the live violation of the documented 500-line Context Efficiency budget that `security-and-hardening` was the only skill breaching. I'd flagged that file on a validator PR and explicitly declined to do the content fix myself on the grounds that which material to move is an editorial judgement about the skill — this is that judgement, made by someone who knows the content, and it lands better than a mechanical split would have.

Syamsuddin:
Thank you for such a careful review — walking all 28 inline checklist items against `references/security-checklist.md` and diffing the heading sets is exactly the verification a 394-line deletion needs, and I'm glad it holds up under that scrutiny.

You called out the two decisions I was least sure about:

- **The three items added to the shared checklist.** Those were the only inline items with no counterpart (data classification/minimisation, retention with a real deletion path, and data-subject export/delete). Folding them into the shared file rather than keeping a second copy alive felt like the right side of the "one fact, one home" rule, and it's reassuring that the drift you noticed was already there before this PR.
- **Skill-local `references/hardening-patterns.md`.** I went back and forth on this versus repo-root `references/`; the `skill-anatomy.md` L125 rule settled it, since nothing else in `skills/` links the patterns.

Thanks also for the context on the 500-line budget and #571 — that was the motivation here, and I agree the split should be an editorial call by whoever knows the content, not a mechanical one.

#580 applies the same approach to `performance-optimization` (the next-largest skill) if you'd like to take a look.

addyosmani (approved):
Verified this end to end and it's a clean lossless split: all 15 code blocks moved byte-identical into hardening-patterns.md, nothing dropped, the inline checklist correctly points to the shared security-checklist.md (which already covers all of it plus the 3 privacy items you added), and every anchor resolves. It takes SKILL.md from 524 lines to 216, which is a real win. Smaller always-on context is exactly the direction I want given the Claude 5 feedback in #433. abhisheksharma2411 independently verified nothing was lost too. Merging, thank you.

PR #580, opened 2026-09-18
refactor(performance-optimization): move Step 3 code patterns into a skill-local reference (-25% tokens per load)

Syamsuddin:
## Why

Second in the series started by #579. `performance-optimization/SKILL.md` sat at 496 lines / 3,314 words (~4.3k tokens), just under the 500-line budget in [docs/skill-anatomy.md](https://github.com/addyosmani/agent-skills/blob/main/docs/skill-anatomy.md#context-efficiency) and the largest skill after `security-and-hardening`. Step 3 ("Fix Common Anti-Patterns") alone was ~1,800 words of worked TypeScript, SQL, HTML, and TSX — material an agent only needs once Step 2 has named the bottleneck, but paid for on every activation.

## What changed

**`skills/performance-optimization/SKILL.md`** — 496 → 267 lines, 3,314 → 2,488 words

- Lines 1–120 (frontmatter, Overview, When to Use, CWV targets, workflow, Step 1, the measurement decision tree, Step 2 tables) and everything from Step 4 onward (Verify, Guard, See Also, Rationalizations, Red Flags, Verification) are **byte-identical** to `main`. No description edit, so routing is untouched.
- Step 3 is replaced by a condensed "Fix the Bottleneck" list: each anti-pattern's rule and its recognizable signature, with a deep link to the worked fix. Every rule the later sections depend on is still stated inline — read the plan before adding an index and revert one that didn't change it; cache key includes every input that changes the response; one invalidation strategy; stampede protection; a bigger pool relocates the queue; `memo`/`useMemo` overuse is its own cost.

**`skills/performance-optimization/references/optimization-patterns.md`** (new, 250 lines)

- Step 3's body moved **verbatim** (line range extracted with `sed`, then verified by script that the body string matches modulo two mechanical changes: `####` headings promoted to `##` so the file stands alone, and one `../../references/` link deepened to `../../../`). All 15 fenced blocks in the original are preserved unchanged.
- Same convention as `constraint-driven-development/references/floor-guard.md` (#472) and #579.

## Verification

- `node scripts/validate-skills.js` — 25 skills, 0 errors, 0 warnings
- `node scripts/validate-reference-links.js` — pass
- `node scripts/run-evals.js --min-rank1 95` — rank-1 100% (88/88), no collision warnings
- All 8 `#anchor` links resolve; every heading in the new file is linked from SKILL.md
- The behavioral eval's expectations (measure before optimizing, target the measured bottleneck, verify against baseline, add a guard) all live in Steps 1, 2, 4, 5, which are unchanged

## Coordination with #492

#492 touches this file in two places: the "Slow API responses" row of the Step 2 Backend table and a paragraph inserted after the See Also sentence. Both context regions are untouched here, so #492's hunks still apply on top of this in either merge order.


nucliweb (commented):
The move is faithful and I'd be comfortable with it once the condensed Step 3 is tightened in three places. I wrote the query-plan and caching sections in #521, so I read the new bullets against what those tables teach; that list is now the only thing an agent sees unless it opens the link, and one bullet says something the reference file contradicts.

What I ran, on a test-merge against `c004a74`: validate-skills 25 skills 0 errors, validate-reference-links pass, the rest of the CI set green locally (no checks have run on the PR itself yet). The moved body diffs identical to the old range after the declared normalizations, lines 1-120 and Step 4 onward are byte-identical, all 8 anchors resolve, and #492 merges clean in both orders with identical trees. Leaving Steps 4 and 5 untouched was the right call, it made this checkable.

The three things before it lands:

- "An index will not help a low-selectivity dominant value, a leading wildcard, or a function applied to the column" drops the qualifiers from the table it summarizes, which says "needs trigram or full-text" and "index the expression instead". As written, an agent that doesn't open the link concludes no index fixes `WHERE lower(email) = ?`. Suggested: "A plain B-tree index will not help a low-selectivity dominant value (a partial index serves the rare value), a leading wildcard (needs trigram or full-text), or a function applied to the column (index the expression)."
- The three plan signatures are listed, but "each call for a different fix" no longer says which. The one worth keeping inline is that a `rows=` estimate off by an order of magnitude means stale statistics, because its fix is `ANALYZE` rather than an index, which is the point of the section.
- Red Flags still audits "a cache with no stated staleness window", but the rule that teaches it (state the acceptable window explicitly) now lives only in the reference. A few words in the caching bullet cover it.

One non-blocking note on the headline, which applies to #579 too: -25% is the number for the path where the reference is never opened (-24.9% by words, confirmed). Opened whole, the activation costs about 17% more than `main`, since the condensed list is 511 words of new prose on top of the 1,399 moved, and the skill's own trigger prompts (N+1, indexes, LCP) are all Step 3 topics. I'd lead with the 500-line headroom instead (496 to 267, with #492 about to add lines), which holds unconditionally, and tell the agent to read the one section it needs rather than the whole file. The 88/88 doesn't speak to this change either way, since routing ranks on `name` and `description` only; #584 covers what the evals can observe for skill-local references.

Whether the catalog wants this split applied skill by skill as a series is a direction question rather than a quality one, so I'll leave that to @addyosmani. Placement itself matches skill-anatomy, same shape as `floor-guard.md`.

addyosmani:
The move itself is faithful, I checked, nothing is lost since the full detail lives in the reference. Before I merge, please address nucliweb's three points on the condensed Step 3 summary. One is a real correctness issue: as written, the shortened index bullet drops the qualifiers, so an agent that doesn't open the reference could wrongly conclude no index helps a WHERE lower(email)=? query. Same for the stale-stats-means-ANALYZE case, and the cache-staleness Red Flag that now only lives in the reference. Small wording fixes to the summary and it's ready. Thanks for doing this, it pairs well with the context-size direction in #433.

Issue #585, opened 2026-09-20
validate-reference-links does not check links inside a skill's own references/ files


nucliweb:
## Problem

`scripts/validate-reference-links.js` enforces that every `references/*.md` link in a `SKILL.md` resolves relative to that skill's directory. Its scope is deliberately narrow (only `references/*.md` links, only `SKILL.md` files), and that narrowness is right: skills legitimately mention paths that don't exist yet.

The gap is that a file under `skills/<name>/references/` can itself link to a shared checklist, and those links sit one directory deeper, so the correct prefix is `../../../references/` rather than `../../references/`. That is the same off-by-a-level mistake the validator exists to catch, in the one place it doesn't look.

This is not live on `main`: `constraint-driven-development/references/floor-guard.md` has no such links. It becomes live with #580 (two pointers to `../../../references/performance-checklist.md`) and #579 (three). I resolved all of them by hand while reviewing and they are correct today, but nothing keeps them correct.

## Proposal

Extend the file set, not the rule: apply the same `references/*.md` link check, with the same fenced-block exemption, to `skills/*/references/*.md`, resolving each link relative to the file that contains it. Add the matching cases to `validate-reference-links-test.js`, including the negative one (a `../../references/` link inside a reference file must fail).

## Related

docs/skill-anatomy.md asks to "keep file references one level deep", linking from `SKILL.md` straight to supporting files rather than chaining. A reference file that points on to a root checklist is a second hop. Whether the validator should warn on that, or whether pointing at a shared checklist is an accepted exception, is a separate call I'd leave to the maintainers; the resolution check is useful either way.

Happy to send the PR. It is a small change to one script and its test. No open PR touches the script; #524 touches `validate-reference-links-test.js`, but it has changes requested and is being split, so I'd rebase on whichever part of it lands.

addyosmani:
Confirmed, this is a real gap. The validator only scans SKILL.md files, so a broken ../../references link inside a skills/<name>/references/*.md file goes unchecked. It wasn't biting on main because almost nothing used skill-local references yet, but that's changing: #579 just landed hardening-patterns.md and #580 adds one for performance, both with links back to the shared references. So it's worth fixing now. Scope: extend the scan to skills/*/references/*.md while keeping the same rule (only references/*.md links, still allowed to mention paths that don't exist yet). Want to take it, or I can. Thanks for catching it before it could rot.

nucliweb:
[...]

PR #586, opened 2026-09-20
fix(validator): reuse fence parsing so broken reference links stay visible

ayobamiseun:
## Summary

The reference-link validator can silently pass a broken link after a prose line beginning with inline backticks. For example:

````markdown
```js``` is inline code, not a fence opener.
See [missing](references/missing.md).
````

On current `main`, this reports zero errors: the validator treats the first line as an opening fence and skips the rest of the file. Its separate fence scanner also accepts trailing text on a closing fence, which can report example links as broken and hide real links after the actual closer.

## Change

- Reuse `stripFencedCodeBlocks` from `scripts/lib/skill-lint.js`, introduced in #568, and remove the duplicate scanner. The existing parser rejects backticks in a backtick fence's info string and permits only spaces or tabs after a closing fence, following the [CommonMark fence rules](https://spec.commonmark.org/0.31.2/#fenced-code-blocks).
- Preserve diagnostic line numbers through the helper's blank-line replacement.
- Add seven CLI regression/compatibility cases: inline backticks, trailing-text closers for both marker types, LF/CRLF input, and backticks in a tilde fence's info string. Six cases fail before the fix and pass afterward.
- Copy the shared module into the test sandboxes so they exercise the same dependency as the real validator.

Three files; no new dependencies. The change in `skill-lint.js` only exports the existing helper and updates its export comment. It leaves the frontmatter, cross-reference, and layout rules under discussion in #428, #492, and #571 unchanged. The scan remains scoped to `SKILL.md`; extending it to supporting reference files in #585 is separate work.

## Verification

- Reference-link and skill-linter suites: **44 passed**, 0 failed.
- Remaining Node suites (eval runner, artifact paths, commands, versions): **35 passed**, 0 failed.
- `node scripts/validate-skills.js`: 25 skills, 0 errors or warnings.
- `node scripts/validate-reference-links.js`: 25 skills, 0 errors.
- Manifest-version, command-parity, and artifact-path validators: pass.
- `node scripts/run-evals.js --min-rank1 95`: 140 checks passed; rank-1 rate 100%.
- `bash hooks/session-start-test.sh`: payload test passes.
- `git diff --check`: clean.

Follow-up to #517 and #568.


addyosmani (approved):
This is the consolidation #517 flagged for later: now that #568 landed the shared CommonMark scanner, the reference-link validator reuses it instead of carrying its own copy. Correct, and the CRLF and inline-backtick tests cover the tricky cases. Merging. Note this doesn't fully close #585 (links inside a skill's own references files still aren't scanned), I'll track that part separately.

PR #587, opened 2026-09-20
fix(evals): remove stale grading and record run identity

mavericksea-ai:
## Summary

- Clear both `.grading.json` and `.grading.raw.txt` at the top of each
  eval iteration, before the executor is invoked, so no stale result file
  can survive an executor or grader crash, and a successful grading
  cannot leave a stale raw file behind.
- Record run identity in each `.grading.json` under a `run` key.
  Shape captured from a local `claude -p --verbose --output-format stream-json`
  invocation whose init event reported `claude-opus-4-6[1m]`:
  ```json
  "run": {
    "executor_model": "claude-opus-4-6[1m]",
    "grader_model": "unknown",
    "timestamp": "2026-09-21T12:34:56.789Z"
  }
  ```
  `executor_model` is the `model` field from the stream-json init event
  the executor already emits. `timestamp` is `new Date().toISOString()`
  at write time. The `run` key is only present when the caller provides
  metadata; `persistGradingOutcome` without it writes the grading object
  unchanged.

## Why this isn't covered by #473, #563, or #576

Those PRs hardened `parseGrading` validation (#473 rejected incomplete
summaries, #563 guarded against null expectations, #576 bound results to
declared expectations by id). They all operate at the grading-parsing
layer: they decide whether a grader response is accepted or rejected.

This PR operates at the persistence layer: what happens to the files
**after** that accept/reject decision is made. None of the three PRs
addressed the fact that `RESULTS_DIR` is never cleared and `base` is
derived from the skill and eval id alone, so a previous run's
`.grading.json` survives at exactly the path a collector reads, even
when the current run's grading is rejected. Nor did they record which
model produced the result or when.

## Executor model source

The model is read from the `{"type":"system","subtype":"init","model":"…"}`
event that the executor's `--output-format stream-json` already emits.
The grader model is recorded as `"unknown"` because the grader is a bare
`claude -p` whose output `parseGrading` extracts by regex; switching it
to `stream-json` would break parsing, and adding `--model` is a behaviour
change outside the scope of a recording fix.

## Test plan

- [x] `node --test scripts/run-evals-test.js`: 33 pass (11 new), 0 fail
- [x] `node scripts/run-evals.js`: 140 checks, 0 errors, 100% rank-1
- [x] Each new test fails before its fix and passes after

**Environment:** macOS Darwin 25.5.0, Node 24. CI
(`.github/workflows/test-plugin-install.yml`, job `validate-skills`)
runs `node --test scripts/run-evals-test.js` and
`node scripts/run-evals.js --min-rank1 95`, so both the unit tests and
the deterministic eval gate execute on every push and pull request.


addyosmani (approved):
Good fix. Clearing the grading slot before each behavioral run means a grader crash can't leave a stale grading.json passing as the current result, and stamping run identity onto the output is a nice touch. Verified the tests pass (33). Merging.

PR #588, opened 2026-09-20
fix(validator): check references/ links inside a skill's own references/ files

### LINKED ISSUE #585: validate-reference-links does not check links inside a skill's own references/ files

nucliweb:
Closes #585.

## What

`validate-reference-links.js` also scans the markdown files directly inside `skills/<name>/references/`, resolving each link from the file that contains it. Same rule as before (only `references/*.md` links, fenced blocks exempt, not a general path linter); only the file set grows, which is the scope agreed in #585.

From a reference file the shared checklists are three levels up, so the failure hint adds `../../../references/<file>.md` when the violation is in one. The `N skills checked` summary is unchanged, so no existing assertion moved.

## Verification

| Check | Result |
|---|---|
| `node --test scripts/validate-reference-links-test.js` | 21/21 (18 existing + 3 that failed before the change + 1 scope guard) |
| Validator on `dc27a9c` | 25 skills, `floor-guard.md` and `hardening-patterns.md` pass, exit 0 |
| Validator on `dc27a9c` + the files from #580 | `optimization-patterns.md` passes, exit 0 |
| Mutation: one `../../../references/` changed to `../../references/` in each real reference file | 2 errors reported at L5 with the three-level hint, exit 1 |
| `validate-skills`, `validate-commands`, `validate-artifact-paths`, `validate-versions`, `skill-lint-test` | pass |

## Coordination

Built on top of #586 (shared fence parsing). #524 also touches `validate-reference-links-test.js`; it has changes requested and is being split, and these tests are appended as a block, so either order rebases cleanly. #580 needs no change to pass.

Left out on purpose: whether a reference file pointing on to a root checklist counts as a second hop under skill-anatomy's "one level deep" rule. That is a policy question, not a resolution check.


abhisheksharma2411 (commented):
Checked this out and ran it rather than reading it, and everything in your verification table reproduces:

- `node --test scripts/validate-reference-links-test.js` → **21 pass / 0 fail**
- validator on the PR head against the real tree → `25 skills checked — 0 error(s) — PASSED`, **exit 0**
- both reference files are genuinely picked up (`constraint-driven-development/references/floor-guard.md`, `security-and-hardening/references/hardening-patterns.md`)

I also mutated it myself rather than trusting the table. Appending a two-level link to a file that sits three levels deep:

```
L326: ../../references/security-checklist.md — resolves to skills/references/security-checklist.md, which does not exist
```

**exit 1**, with the three-level hint printed and the two-level hint still there for SKILL.md violations. The `referenceFileErrors` gate does exactly what it should — the extra hint only appears when a reference file was the one that failed, so SKILL.md-only runs keep the old, shorter message.

The refactor is the right shape too. Threading the base directory through `findViolations(file)` and deriving it from `path.dirname` makes SKILL.md a special case of the general rule rather than the other way round, and it's behaviour-preserving for SKILL.md since `path.dirname(skillFile) === skillDir`. Carrying `target` on the violation instead of recomputing it in the printer removes the chance of the check and the message disagreeing.

**One gap worth a sentence in the code, not necessarily a change:** `skillReferenceFiles` uses a flat `readdirSync`, so anything in a nested subdirectory is silently unchecked. Demonstrated:

```
skills/security-and-hardening/references/deep/nested.md
  → [x](../../references/does-not-exist.md)
  → 25 skills checked — 0 error(s) — PASSED
```

No skill has a nested reference directory today (I checked — `find skills -path "*/references/*" -type d` is empty), so this is correct right now and matches the scope agreed in #585. The risk is that it degrades silently: the first person to add `references/patterns/foo.md` gets no coverage and no warning that they've fallen outside the checked set. Since the file's header comment is already careful about scope, I'd add the flat-scan choice to it — one clause saying nested directories are deliberately out of scope is enough to make the next person's decision an informed one rather than an accident.

Two smaller observations, neither blocking:

- The summary still reads `N skills checked` while the per-file lines are now a mix of skills and reference files. You call this out as deliberate to avoid moving assertions, which I agree with — but a reader seeing `25 skills checked — 1 error(s)` after a reference file failed has to work out that the count and the error come from different units. `25 skills checked (27 files)` would keep every existing assertion on the `N skills checked` prefix while making the output self-describing.
- Output is now one line per file rather than per skill, so a green run is longer. Fine as is; just noting it in case CI log volume matters to anyone.

The prose-false-positive case I went looking for doesn't bite: a reference file discussing `../../references/x.md` **outside** a fence would now fail, but the fence exemption is shared with the SKILL.md path and no current reference file does that. Worth remembering if the skills ever start documenting their own link conventions in prose — it's the same hazard the header comment already flags for fences.

Nothing blocking from me. Nice, tightly-scoped fix with real tests.

addyosmani (approved):
**praise:** this is the right fix for #585, and I verified it catches the bug (a wrong two-level ../../references link inside hardening-patterns.md gets flagged now) while staying clean on the real content. Resolving each link from the file that contains it is exactly right. Merging, thanks for taking it.

Issue #584, opened 2026-09-20
Behavioral evals run a skill without its reference files


nucliweb:
## Problem

`scripts/run-evals.js --behavioral` gives the executor the text of `SKILL.md` and nothing else. The skill body is passed through `--append-system-prompt`, and the run happens with `cwd` set to a throwaway workspace that holds only the eval fixtures (`scripts/run-evals.js:537-538`). Every relative link in that text, whether `references/floor-guard.md` or `../../references/performance-checklist.md`, points at a path that does not exist in the workspace.

So under the harness the agent is always the one that never opens a reference. That has two consequences:

- A behavioral eval cannot pass or fail because of anything in a reference file. Content that moves out of `SKILL.md` leaves the tested surface entirely.
- A green behavioral run on a PR that moves content into `references/` says nothing about whether the split works: whether the agent opens the file at the right step, or gets a rule wrong when it doesn't.

## Why it matters now

docs/skill-anatomy.md recommends progressive disclosure and a skill's own `references/` directory. One skill uses it on `main` (`constraint-driven-development/references/floor-guard.md`), and #579 and #580 each move several hundred lines of `security-and-hardening` and `performance-optimization` behind that kind of link. Reviewing #580, the only real defects were in the condensed text left in `SKILL.md` (qualifiers dropped from an index rule, a Red Flag whose rule now lives only in the reference), and no gate could have caught them. The routing evals can't either, by design: they rank on `name` and `description` only.

## Options

1. **Stage the skill directory in the workspace.** Copy `skills/<name>/` (and the root `references/` at the matching relative depth) into the workspace and tell the executor where the skill lives, so its links resolve the way they do in an installed plugin. Smallest change that makes references reachable.
2. **Add a trace expectation for reference use.** With option 1 in place, a case can assert "opened `references/x.md` before editing" from the tool calls the grader already reads. This is what would test a progressive-disclosure split.
3. **Document the limit** in evals/README.md if neither is wanted: behavioral evals cover `SKILL.md` text only, so a PR that moves content to a reference needs its condensed text reviewed by hand.

Option 3 is cheap and worth doing regardless. I'm glad to take option 1 as a small focused PR if the direction looks right. One overlap to sequence: #455 changes how fixtures are copied into the eval workspace (the `fs.cpSync` call in the same function), so staging the skill directory should land after it and reuse its symlink handling.

addyosmani:
[...]

PR #589, opened 2026-09-20
fix(security-and-hardening): restore rules the condensed text dropped in #579

nucliweb:
Follow-up to #579. The code move there is lossless, as verified in its reviews. What slipped through is prose: the cookie bullet and three sections (supply chain, privacy, LLM) were rewritten rather than moved, and a few rules now appear in none of `SKILL.md`, `hardening-patterns.md`, or `security-checklist.md`. I had a review of #579 in progress with these findings when it merged, so here they are as a patch.

## What comes back

| Line | Restored | Where it was before #579 (`c004a74`) |
|---|---|---|
| L90 | `sameSite: 'lax'` or `'strict'`, and that it is the CSRF defense | snippet comment `sameSite: 'lax', // CSRF protection`, the skill's only mention of CSRF |
| L137 | "forced fixes may cross declared dependency ranges" | L316 |
| L138 | `npm audit signatures`, `pnpm audit signatures`, "not automatic proof of compromise" | L317 |
| L148 | "Every personal-data store needs a TTL and a working deletion path" | L392 |
| L150 | consent is auditable | L394 |
| L156 | "(2025)" on the OWASP LLM Top 10 link, which the bare `LLM0x` IDs depend on | L401 |
| L160 | "the full system prompt" and "validate every tool argument" | L405-406 |

Two of these matter most. `sameSite` with no value lets `sameSite: 'none'` satisfy the rule, and [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie) describes `None` as "Send the cookie with both cross-site and same-site requests". And Red Flags still audits "the full system prompt placed inside an LLM context window" while the rule had stopped naming it. For the supply-chain clause, the [npm docs](https://docs.npmjs.com/cli/v11/commands/npm-audit) give the reason in their own words: `--force` installs "outside your stated dependency range (including SemVer-major changes)".

## Verification

| Check | `dc27a9c` | this branch |
|---|---|---|
| Each restored phrase, occurrences in `SKILL.md` | 0 | 1 |
| "full system prompt" (rule + Red Flag) | 1 | 2 |
| `SKILL.md` | 216 lines / 2,449 words | 216 lines / 2,504 words |
| `validate-skills`, `validate-reference-links`, `validate-commands`, `validate-artifact-paths`, `validate-versions`, `skill-lint-test` | pass | pass |

Routing evals stay at 88/88, which says nothing here since no `description` changed (#584). No behavioral eval run.

## Notes

No open PR touches this skill or `security-checklist.md`.

A correction to my review on #580: I wrote that the token-cost note applied to #579 too. It doesn't. Opening `hardening-patterns.md` whole costs +2.5% over the old `SKILL.md`, against +17% for #580, so the saving here holds on nearly every path.

Left for a separate PR if wanted, since they change sentences that are new in #579 rather than restore anything: "Never `*` with credentials" skips the reflected-origin case, and the upload bullet allowlists a client-sent `Content-Type`.


abhisheksharma2411 (commented):
I reviewed #579 and signed off that the move was lossless, so let me start there: you're right that I checked the wrong thing. I verified the **code blocks** survived the move and never diffed the surrounding prose, which is exactly where the changes you're describing happened. Fair catch, and the `sameSite` observation is a good one on the merits — a bare `sameSite` with no value is satisfiable by `'none'`, which is the opposite of the intent.

So I'd like the SKILL.md wording changes to land. But I checked each of the seven rows against the merge base (`dc27a9c`) across all three files, and **the framing is overstated** — most of these aren't absent from the set, and I think that matters because the PR is arguing "restore rules that now appear nowhere".

| row | claim | at `dc27a9c` |
|---|---|---|
| L90 `sameSite`/CSRF | appears nowhere | **present** — `references/security-checklist.md:39` has `sameSite: 'lax'`, and `hardening-patterns.md:42` has the literal `sameSite: 'lax',    // CSRF protection` |
| L137 forced fixes cross ranges | appears nowhere | **rule present**, rationale absent — `security-checklist.md:187` and `SKILL.md:137` both already say forced remediation is never automatic and diffs/changelogs are reviewed |
| L138 `npm audit signatures` | appears nowhere | **rule present**, commands absent — `security-checklist.md:188` "Registry signatures/provenance are verified where the manager supports it"; `SKILL.md:138` "Verify registry signatures where supported" |
| L148 TTL + deletion path | appears nowhere | **rule present** — `security-checklist.md:142` "Personal data has a retention limit and a working deletion path (incl. backups, caches, indexes)"; `SKILL.md:148` carries the same bullet |
| L150 consent auditable | appears nowhere | consent + DPA present at `security-checklist.md:143`; **"auditable" genuinely absent** ✓ |
| L156 `(2025)` | appears nowhere | **genuinely absent** ✓ |
| L160 full system prompt / validate tool arguments | appears nowhere | "full system prompt" **present** at `SKILL.md:196` *and* `security-checklist.md:198`; "validate every tool argument" **genuinely absent** ✓ |

Three of seven are true absences. The other four are rationale, concrete commands, or emphasis that already exist in the reference files the skill points at.

The L160 row is the one to look at again, because the PR body contradicts itself on it: the table lists "the full system prompt" as restored, and then the prose immediately below says *"Red Flags still audits 'the full system prompt placed inside an LLM context window'"*. Both can't be right, and the prose is the accurate one — `SKILL.md:196`. The real argument there is **internal consistency**: Red Flags audits something the guidance section no longer states, which is a legitimate reason to change the line and a better one than "it was dropped".

What I'd suggest, and it's mostly a rewrite of the description rather than the diff:

- **Keep every wording change.** Each one makes `SKILL.md` stand on its own instead of depending on a reader following a link, and the `sameSite` and tool-argument ones close real gaps.
- **Reframe from "restore what was lost" to "make SKILL.md state the rule the references already carry".** That's both accurate and a stronger argument — a skill body that's weaker than its own checklist is a defect regardless of which commit caused it.
- **Mark the three true absences as such** (`(2025)`, `validate every tool argument`, consent auditability), since those genuinely restore something no file says.

One thing I'd genuinely double-check before merge: `security-checklist.md:39` already pins `sameSite: 'lax'` while your new `SKILL.md:90` allows `'lax'` **or** `'strict'`. Both are defensible, but the two files will now disagree on whether `'strict'` is acceptable, and the next person to condense them will have to pick. Worth aligning in the same PR.

Method note in case it's useful: my first pass at this used `grep -E "a\|b"`, where `\|` is a literal in ERE rather than an alternation, and it reported five of these as absent when they aren't. It's the same trap I fell into reviewing #579. Worth re-running your own absence checks with `grep -F` per term, or `grep -E "a|b"` unescaped, before the description goes out.

addyosmani (approved):
**praise:** good catch, and you're right about the distinction. The code move was lossless, but the prose condensation quietly dropped real rules: the sameSite CSRF line, npm audit signatures, the retention TTL, validate every tool argument. Those belong back. Verified it merges clean and validators pass. Merging, thank you for the careful read.

PR #590, opened 2026-09-20
ci: run validator checks on macOS and Windows

ayobamiseun:
## Summary

All validator CI currently runs on Ubuntu, so platform-specific path handling, line endings, and subprocess behavior can regress without a failing check. The Windows path fix in #573 is a concrete example of the coverage gap.

Add macOS and Windows coverage for the same deterministic Node tooling already checked on Linux.

## Change

- Add a `validate-platforms` matrix on `macos-latest` and `windows-latest`, using Node 24 and the existing checkout/setup-node action versions.
- Run all six Node test suites, the five validators, and the trigger/routing evals with the existing 95% rank-1 floor. List test paths explicitly so the command works in PowerShell without shell glob expansion.
- Fetch full history and tags for the manifest-version test's `git describe` check.
- Set `fail-fast: false` so a failure on one platform leaves the other platform's diagnostic run intact. Neither platform permits failures.
- Make plugin validation depend on the complete matrix. Preserve the existing Ubuntu jobs and check names, including shell-hook and plugin-install coverage.

One workflow file. No test logic, thresholds, or dependencies changed. The matrix uses the existing workflow's push, pull-request, and manual triggers.

## Verification

- Local macOS run (Node 25.9.0): all **90 Node tests passed**, 0 failed or skipped.
- All five validators passed; 25 skills, 9 command families, and 7 guarded artifact files checked.
- Routing evals: **140 checks passed**, rank-1 rate **100%**.
- Parsed the workflow YAML and checked that every current Node test file appears in the matrix command, full history is fetched, both platforms gate plugin validation, and existing jobs are preserved.
- `git diff --check`: clean.

Hosted Node 24 verification on the exact PR commit: both macOS and Windows passed all 90 tests, every validator, and the routing checks. Existing Ubuntu checks and plugin validation/installation also passed. [Successful workflow run](https://github.com/ayobamiseun/agent-skills/actions/runs/35540577164).

The initial fork run failed the existing version check because the fork had no release tags. After copying upstream's unchanged `0.6.10` tag to the fork, the rerun passed without code changes.


addyosmani (approved):
**praise:** given how much Windows pain we've had lately, running the Node validators on macOS and Windows is worth it, and you scoped it right by keeping the bash hook tests off the matrix. Merging.

**nitpick:** validate-versions.js calls git describe with no try/catch, so if a runner ever checks out without tags it throws. Pre-existing, and fetch-depth 0 handles it today, so nothing to change, just noting it.

Issue #529, opened 2026-08-29
Proposal: HUMAN.md, an evidence-based human protocol for each skill


federicobartoli:
Every SKILL.md in this pack instructs the agent. Nothing instructs the human at the gates the skill itself creates. This proposes the counterpart file: HUMAN.md, next to SKILL.md, never loaded by the agent.

Here is what one would look like (test-driven-development, abridged):

## Gate 1: the failing test (after "Step 1: RED", ~2 min)
Read only the test diff, then the failure output. Ask:
- Does it fail for the right reason? (read the error message, not the code)
- Would a trivially wrong implementation pass it?
  → redirect: "Strengthen the assertion, it currently accepts X."
- Behavior, or implementation detail?
  → redirect: "Rewrite the test against the public interface only."

## Gate 2: after the refactor, before commit (after "Step 3: REFACTOR", ~3 min)
Read the implementation diff only. Ask:
- Did anything change outside the slice?
  → redirect: "Revert changes outside the task scope."
- Do the test names still describe behavior after the refactor?

## Stop rules
- Third failed iteration on the same test → write the test yourself.
- Diff grew beyond the files named in the plan → pause and re-scope.

That is the whole format: gates that cite SKILL.md step headings, gate count scaled by how often that skill's class of work fails, a review phrased as questions (~5 min total per run), redirects the agent can act on, objective stop rules.

**Why a separate file, not a `## For humans` section:** SKILL.md is loaded into agent context in full, so a human section would cost tokens on every invocation and put the reviewer's checklist inside the model's prompt. A separate file costs zero tokens, and since no harness parses it, it is tool-agnostic by construction.

**Why not the existing `## Verification` sections:** every skill already ends with one, but it is a list of exit criteria the agent self-certifies after the work is done ("Every new behavior has a corresponding test", "No tests were skipped"). HUMAN.md is the other half: what the human inspects during the run, at specific steps, with redirects. They complement each other: Verification is what the agent claims at the end; the gates are where the human samples the work early enough to steer it.

**Scope:** per-skill files in this repo. Not proposing an ecosystem standard.

[...]

PR #593, opened 2026-09-21
docs(tdd): pilot opt-in human review checkpoints (#529)

ayobamiseun:
## Summary

A human following a TDD run needs to know what to inspect while the agent is working, before a weak test or an expanding diff becomes a finished patch. #529 proposes this counterpart to the agent's workflow; the maintainer suggested trying a short section inside existing skills first.

This pilots that approach in `test-driven-development` with two review checkpoints.

## Change

- **After RED, before implementation:** inspect the test diff and actual failure output. Check whether the failure demonstrates the intended missing behavior and whether a trivially wrong implementation could satisfy the assertions. Include concrete redirects for weak assertions and unrelated setup failures.
- **After REFACTOR, before committing:** inspect both diffs and the latest suite output. Check scope, weakened assertions, skipped tests, and whether the output covers the current code. Include redirects for scope growth and changed requirements.
- Give the human an intervention signal when the same failure recurs without new evidence: pause edits, summarize what the attempts established, and choose the next diagnostic check.

Review pauses are requested by the human. The section points to `code-review-and-quality` for the full review before merge.

One file, 13 added lines. The existing TDD steps, frontmatter, and verification checklist stay intact. This is a small pilot of the issue's worked example, with the questions and redirects colocated beside the steps they refer to.

## Verification

- `node scripts/validate-skills.js`: 25 skills, 0 errors or warnings.
- `node scripts/validate-reference-links.js`: 25 skills, 0 errors.
- `node scripts/run-evals.js --min-rank1 95`: 140 checks passed; rank-1 rate remains 100%.
- TDD behavioral eval dry-run resolves all three existing cases. Live behavioral evals were not run; the required `claude` executable is unavailable locally.
- `git diff --check`: clean.

Checked the open TDD proposals (#410, #416, #424) for overlap; they address planner handoff, flaky tests, and portability rather than these human-facing checkpoints. The rejected-change ledger has no earlier entry for this proposal.

Refs #529. Leaving the broader proposal open for feedback on the pilot before extending it to other skills.


addyosmani:
**praise:** this is exactly the in-skill human-gate pilot I floated on #529, and the two-checkpoint table (after RED, after REFACTOR) with a concrete redirect for each is well done.

**question (blocking):** the thing I want to settle before this lands is placement. This is human-facing content living in an agent-loaded SKILL.md, which adds tokens the agent doesn't use, and I'm actively trying to trim always-on context (see #433). Federico's #529 proposed a sidecar HUMAN.md for exactly this. Should the human-gate guidance sit beside the skill rather than inside it? I'd like to converge on that in #529 before we set the pattern in one skill. Holding on placement, not on the content.

Issue #494, opened 2026-08-14
Unable to send requests to Model providers when embedding Skills in Cursor prompts


eliott-herbert-byrnes:
As the title describes, when attaching these Skills into a prompt within a Cursor chat e.g 'hello, please execute /skill-name' once the request is sent, an error occurs stating 'an unexpected error occurs on our servers'. No further requests will process to any model provider and all context is lost. 

The workaround is to is to explicitly tell the agent to lookup the Skill and execute it, without embedding it in the chat. 

Quite a strange bug, unsure what the cause is. I have raised this with Cursor also. I have been using Skills for over a year, and have only experienced this issue with these Skills. 

The only suspicion I have is that it is potentially being caused due to the length of these Skills in comparison to others.

federicobartoli:
Hi @eliott-herbert-byrnes  !!!Thanks for flagging this! I don’t have Cursor, but I downloaded it and tried the skills, and they work fine for me. Could you tell me how you installed our skills, what error you’re getting, etc.? Thanks so much :D

The more details you can give us, the more easily we can get to the bottom of the problem.

eliott-herbert-byrnes:
Hi @federicobartoli, here is an official response from Cursor staff explaining the error 'When a skill is attached to a message (with /, or when you accept an offer to load one), Cursor reads the YAML frontmatter at the top of that skill’s SKILL.md. If that frontmatter isn’t valid YAML, the request fails with a generic server error. Later messages in the same chat keep failing because the skill stays attached. Skill length is not the trigger...'.

Here is a link to the full response https://forum.cursor.com/t/error-connecting-to-server-sending-prompts-with-skills-attached/168440

federicobartoli:
Thanks for chasing this down with Cursor, that's really useful context!!
I tried to reproduce the "invalid YAML frontmatter" explanation against the repo and I can't get it to fail. All 25 SKILL.md files (plus the command and agent files) parse with a strict YAML parser, and none of them hit the three causes Cursor listed in the forum thread: no unquoted description with a colon, no markdown headings or list markers inside the frontmatter, and every file opens and closes the `---` block correctly. 

So I suspect the difference is in how the skills ended up in your `.cursor/skills/` folder rather than in the upstream files. Could you help me narrow it down?

1. Which skill(s) trigger the error when attached with `/`? Or is it any of them?
[...]

PR #594, opened 2026-09-21
fix(validator): reject frontmatter that is not valid YAML

abhisheksharma2411:
Refs #494.

## The linter cannot see the failure class that breaks hosts

`parseFrontmatter` in `scripts/lib/skill-lint.js` splits each line on its first colon:

```js
const colonIdx = line.indexOf(':');
if (colonIdx === -1) continue;
const key   = line.slice(0, colonIdx).trim();
const value = line.slice(colonIdx + 1).trim().replace(/^['"]|['"]$/g, '');
```

That is enough to read `name` and `description`, and it is forgiving on purpose. The consequence is that **a SKILL.md whose frontmatter is not valid YAML passes every check in this repo.**

The hosts reading these skills are not forgiving. Per the Cursor staff explanation quoted in #494, attaching a skill to a message parses its frontmatter as YAML, and a parse failure fails the request with a generic server error — and keeps failing for the rest of the chat, taking the user's context with it. So the repo could ship frontmatter that hard-fails a host while CI stayed green.

## What passes today and shouldn't

I isolated each case by mutating **only** the frontmatter of a real SKILL.md that otherwise lints clean (`security-and-hardening`), then checked the same frontmatter with a strict parser (ruby's psych, `YAML.safe_load`):

| frontmatter | `lintSkillContent` before | strict YAML |
|---|---|---|
| baseline, unmodified | 0 errors | OK |
| `description: Use when you harden a service: auth, secrets and review` | **0 errors** | **REJECTED** |
| a tab-indented nested key | **0 errors** | **REJECTED** |
| `description: "Use when you harden a service` (unclosed) | **0 errors** | **REJECTED** |
| a duplicate `description:` key | 0 errors | OK |

Three false negatives, and the duplicate-key row is the control: `safe_load` accepts it, so it must keep passing.

## The fix

`frontmatterYamlErrors` reports exactly those three shapes, with the line number and the fix:

```
Frontmatter line 3 has an unquoted value containing ': ' — YAML reads that as a nested mapping
  and rejects the file; wrap the value in quotes
Frontmatter line 5 indents with a tab — YAML forbids tabs in indentation, so a host that parses
  this frontmatter rejects the whole file
Frontmatter line 3 opens a double quote that never closes — a host parsing this frontmatter as
  YAML rejects the whole file
```

**Deliberate boundaries**, each with a test:

- A colon with **no** trailing space is untouched, so `https://example.com/a:b` still passes. Keying on colons rather than colon-space would fail ordinary URLs.
- A **duplicate key is not reported**, because `safe_load` accepts one. The rule tracks the parser, not taste.

**Why not just use a YAML library:** this repo has no `package.json`, so there is no parser to depend on — which is presumably why the frontmatter reader was hand-rolled in the first place. This stays a small set of rules aimed at the classes actually observed to break a host, not a YAML implementation. No current frontmatter uses block scalars or multi-line values; the header comment says so and says to add a rule rather than loosen one if that changes.

@federicobartoli already did this check by hand across all 25 skills in the #494 thread and found them clean — which is exactly why it's worth making it a check instead of a memory. This PR does not claim to explain that reporter's error; the maintainer's read that it is host-side stands. It closes the gap that made the question hard to answer from CI.

## Verification

| | |
|---|---|
| `node scripts/validate-skills.js` | **25 skills, 0 errors, 0 warnings** — unchanged |
| `validate-commands` / `validate-versions` / `validate-artifact-paths` / `validate-reference-links` | all exit 0 |
| `node --test scripts/lib/skill-lint-test.js` | **35 pass / 0 fail** (8 new) |
| `run-evals-test` · `validate-artifact-paths-test` · `validate-commands-test` · `validate-reference-links-test` · `validate-versions-test` | 33 / 6 / 6 / 17 / 1, all 0 fail |

Mutation-tested **4/4** — each fails the new tests:

| mutation | result |
|---|---|
| never call the checker | 4 fail |
| colon rule matches any colon (would flag URLs) | 1 fail |
| drop the tab rule | 1 fail |
| accept an unterminated quote | 1 fail |


addyosmani:
This landed via #595, which included your skill-lint change verbatim and extended the same YAML check to command frontmatter. Nothing lost, and thanks for both.

PR #595, opened 2026-09-21
feat(validator): hold commands to the same frontmatter rules as skills

abhisheksharma2411:
> **Stacked on #594** — it adds `frontmatterYamlErrors`, which this reuses. Please merge that one first; this diff is only the two `validate-commands` files. Happy to rebase if #594 changes in review, or to fold the two together if you'd rather review one PR.

Refs #494.

## The same gap, on the other set of files

`descriptionFromMd` in `validate-commands.js` reads frontmatter the forgiving way:

```js
const colonIdx = line.indexOf(':');
if (colonIdx === -1) continue;
if (line.slice(0, colonIdx).trim() === 'description') {
  return line.slice(colonIdx + 1).trim().replace(/^['"]|['"]$/g, '');
}
```

That is the same split-on-first-colon read the skill linter used before #594, so a command whose frontmatter is **not valid YAML** passes every check here — and Claude Code parses that frontmatter when the command is loaded.

This matters because the manual verification in #494 covered *"all 25 SKILL.md files (plus the command and agent files)"*. #594 turned the first half into a check. This is the second half.

## It is a guard, not a repair

All nine current commands already parse. I confirmed that independently against ruby's psych before writing anything:

```
build.md            OK  keys=description
code-simplify.md    OK  keys=description
constraints.md      OK  keys=description
plan.md             OK  keys=description
review.md           OK  keys=description
ship.md             OK  keys=description
spec.md             OK  keys=description
test.md             OK  keys=description
webperf.md          OK  keys=description
```

Same shape as the skills half, and the same reason: the manual check was real, it just could not stay true on its own.

## Scope

**`.claude/commands/*.md` only.** The `.gemini/` and `commands/` directories are TOML and already go through a real TOML parser, so a malformed one surfaces as a missing description in the existing check. One of the tests pins that the YAML rules are *not* applied to them — a TOML description legitimately contains a colon, and running these rules over it would fail nearly every command.

## A thing the tests found

The sandbox in `validate-commands-test.js` copies the validator into a temp tree and runs it there. Adding a `require` of `scripts/lib/skill-lint.js` broke **all six** existing tests with a module-not-found before I had written a new one:

```
Error: Cannot find module '…/agent-skills-validate-commands-test-vtPfJI/scripts/lib/skill-lint.js'
```

The sandbox now copies the lib beside the validator. Worth flagging because it is a property of the harness that any future shared helper will hit the same way.

## Verification

| | |
|---|---|
| `node scripts/validate-commands.js` | **9 commands checked — 0 errors — PASSED** |
| `node --test scripts/validate-commands-test.js` | **11 pass / 0 fail** (5 new) |
| all five validators | every one exits 0 |
| all six script suites | `run-evals` 33, `artifact-paths` 6, `commands` 11, `reference-links` 17, `versions` 1, `skill-lint` 35 — 0 fail |

Mutation-tested **2/2**:

| mutation | result |
|---|---|
| never run the frontmatter check | 2 fail |
| report its errors but do not count them | 2 fail |

The second one matters more than it looks: a validator that prints a problem and still exits 0 is worse than one that does not check, because CI goes green over a visible failure.


addyosmani (approved):
**praise:** holding command frontmatter to the same YAML-validity check as skills is the right call, and I confirmed all 9 commands still pass (TOML dirs excluded, and it only enforces YAML validity, not the name field commands don't have). Since this already contains #594's skill-lint change verbatim and extends it to commands, I'm taking this one. Merging.

PR #596, opened 2026-09-22
test(evals): pilot claude plugin eval on code-review-and-quality

federicobartoli:
A small pilot of `claude plugin eval` on one skill, as discussed. Two cases under `evals/plugin/` and a short section in `evals/README.md`; nothing else changes: no manifest key, no CI job, no tool grants, no scaffold.

- `code-review-fires`: a review request with a small diff inline. Graders: the skill fired (plugin-fired indicator only), a free regex looking for the skill's severity taxonomy used as headings or labels, and one judge on the planted off-by-one in `pageOrders`.
- `code-review-stays-quiet`: a test-first request that belongs to test-driven-development. The review skill must not fire; an unscored indicator records whether TDD fires instead.

Why this shape: it is the semantic complement of Tier 2. Tier 2 checks that a description carries the vocabulary users say; this loads the whole plugin and checks whether Claude Code's own router picks the skill on natural phrasing, and what the skill adds to the reply (the with/without delta). It stays on demand, like Tier 3, since every run and every judge is a model call.

Run it with `claude plugin eval . --no-publish`, or `claude plugin eval . --case code-review-fires --runs 1 --ablation none` to try one case cheaply. Results land in `evals/results/`, already gitignored.

Local run summary (Claude Code 2.1.278, claude-opus-5):

```
CASE                     WITH  W/OUT Δ      RUNS COST
code-review-fires        1.00  0.50  +0.50  6    $1.56
code-review-stays-quiet  1.00  1.00  0.00   6    $1.42

2 case(s) · mean Δ +0.25 · 713s · $2.97
```

Full report, with every grader verdict and the judged excerpts of each run: https://claude.ai/artifact/WUSpzJUJuJbHG48mR28FSu

What the runs surfaced, which seems more useful than the pilot itself:

1. Skill invocation is stochastic in headless runs. Across the runs I tracked, the review skill fired in 5 of 7 on the phrasing the case uses; test-driven-development fired in 2 of 7 on "Write a failing test for this bug before we fix it". When the review skill fires, the reply carries a verdict, Critical / Required / Consider tiers, file:line references and a five-axis checklist, and it caught the unauthenticated endpoint; when it does not, the model still finds the off-by-one but reviews in its own shape. In one run it did not fire and the reply still used the skill's taxonomy as headings, which no baseline run did in nine, so the loaded descriptions may shape the reply even without an invocation. The delta is real, and the exit code at the default threshold of 1.0 is not the right signal; the delta and the report are.
2. Phrasing decides whether the skill fires. "Review this diff before I merge it", "Can you do a proper code review of this change before I merge it?" and "Review this pull request for me before it goes into main" never invoked the skill in six runs. Naming the axes ("across correctness, readability, security and performance") does. The case uses that phrasing; the plainer ones are the open question, and if anything should change it is the description, not the prompt. Happy to run a description experiment if you want one.
3. The model sometimes reaches the skill through the `/review` command first (`agent-skills:review`, then `code-review-and-quality`), so cases need a turn budget of 10 or more; with 6 turns a run can end before answering. Also, `/review` asks for "Critical, Important, or Suggestion" while the skill prescribes Critical / Nit / Optional / FYI with no prefix for required changes, and one skill-driven reply mixed the two. Not touched here; flagging in case you want the command aligned with the skill.

One thing worth knowing: the command scans all of `evals/` for `prompt.md` or `case.yaml`, so fixtures must never use those names; the README section says so. If the pilot seems useful, a follow-up could set `experimental.evals` in the manifest to isolate the scan, and add an execution case with `--scaffold` and tool grants for a skill whose deliverable is code. Happy to drop or reshape any of it.


addyosmani (approved):
**praise:** the fire + stay-quiet pair plus the with and without delta is exactly the plugin-eval signal Tier 2 can't give, and the findings are worth more than the pilot itself. Verified locally: purely additive, clean merge, validators green. Merging.

**suggestion (non-blocking):** as a follow-up, scope the scan with `experimental.evals` in the manifest so a fixture named `prompt.md` or `case.yaml` can't get swept in later. Not needed here.

**praise:** good catch on the review.md vs skill taxonomy mismatch. It's real (review.md still says Critical/Important/Suggestion, the skill uses Critical/Required/Nit/Optional/FYI), and I'll align review.md with the skill separately.

PR #598, opened 2026-09-23
fix(evals): grade the ADR's status and date, not timeless language

### LINKED ISSUE #591: documentation-and-adrs eval: does "timeless language" apply to ADRs? SKILL.md does not say
## Question

The behavioral eval for documentation-and-adrs (`evals/cases/documentation-and-adrs.json`) grades a third expectation, "The document is written in timeless language describing current state". `SKILL.md` does not state that requirement for ADRs. The wording comes from `references/definition-of-done.md`, which is about documentation in general, and the skill describes an ADR as a dated record with a status lifecycle. So it is not obvious whether the rule is meant to apply to ADRs, and the grader is left to decide.

## What we saw

`node scripts/run-evals.js --behavioral documentation-and-adrs` at `dc27a9c`, executor model `claude-opus-5` in every run:

- Five runs on one machine (Claude Code 2.1.278, default settings): 3 pass, 2 fail. Both failures were on expectation 3 only. The grader's explanations for the two failures cited wording such as "currently" and "before cutover".
- Five runs on a second machine (Claude Code 2.1.90, model pinned on generator and grader, hooks and MCP off): 5 pass. Five gradings of one of those transcripts agreed with each other.

Expectations 1 and 2 passed in all ten runs. The two sets differ in configuration and CLI version as well as in random draw, so this does not isolate a cause. It does show that expectation 3 is the one whose reading varies, and that with three expectations in the case, that one line decides the skill's verdict.

## Ask

Could you clarify whether "timeless language" is meant to apply to ADRs, and if it is, give an acceptable and an unacceptable example the grader can be held to? If it applies, adding the rule to `SKILL.md` would let the skill and the eval agree. If it does not, the expectation could be replaced with one the skill states (status and date, for instance).

Happy to open the PR for whichever reading you confirm; the branch is ready.

Limits: ten runs, one eval case, one executor model. The first machine did not keep its executor traces, so its failure reasons come from the grader's `evidence` fields rather than the ADRs themselves.

mavericksea-ai:
Fixes #591.

`evals/cases/documentation-and-adrs.json` graded a third expectation, "The document is written in timeless language describing current state". SKILL.md never states that for ADRs: the wording comes from `references/definition-of-done.md`, which covers documentation in general, while the skill's own ADR template carries a Status (Accepted | Superseded by ADR-XXX | Deprecated) and a Date and defines the lifecycle PROPOSED → ACCEPTED → (SUPERSEDED or DEPRECATED). Over the ten runs in #591 it was the one expectation whose reading varied, while the first two passed in all ten.

Per the reading in #591 (the eval is off, not the skill), this replaces it with the requirement the skill does state:

> The ADR records a status from the skill's lifecycle (Proposed, Accepted, Superseded or Deprecated) and a date

The skill is unchanged. One line in one file.

**Verification**
- `node scripts/run-evals.js` at this commit: 140 checks passed, 0 errors, 0 warnings; trigger rank-1 rate 100% (88/88).
- `node scripts/run-evals.js --behavioral documentation-and-adrs` three times on this branch (bb9a5c5 on bcab6a1), Claude Code 2.1.280, same machine as the first five runs in #591: 3/3 expectations passed in each run.

Limits: three runs, one eval case, one machine. The runner writes `evals/results/documentation-and-adrs.eval-1.grading.json` per eval id, so only the last run's grading file survives; the three pass lines are in the terminal output. If you would rather drop the expectation than reshape it, that is a one-line change to this branch.


abhisheksharma2411 (commented):
Hi @mavericksea-ai — I verified every factual claim in the description against `origin/main` rather than taking them on trust, and they all hold:

| Claim | Checked |
|---|---|
| SKILL.md never says "timeless" | ✅ zero matches |
| The wording comes from `references/definition-of-done.md` | ✅ and it's a **repo-root** reference, not one of this skill's — the skill has no `references/` at all, only `SKILL.md` |
| The skill defines the lifecycle | ✅ `PROPOSED → ACCEPTED → (SUPERSEDED or DEPRECATED)` at :96 |
| The template carries Status and Date | ✅ `## Status` :53-54, `## Date` :55-56 |

So an expectation was grading the model against a document the skill doesn't point at, and the variance in #591 is explained. The replacement grades something the skill states in two independent places. That's the right direction, and it's the kind of fix that's easy to get wrong by tightening the *skill* to match the eval instead.

Two small things, neither blocking.

**The template and the lifecycle disagree about `Proposed`.** The template line is:

```markdown
## Status
Accepted | Superseded by ADR-XXX | Deprecated
```

No `Proposed` — but the lifecycle at :96 starts there, and your new expectation lists it. A model following the template literally will never emit `Proposed`, so the expectation is slightly wider than what the template offers. Harmless for grading (the other three are all reachable), but it's a real inconsistency in SKILL.md that this PR happens to surface. Might be worth a one-line follow-up adding `Proposed | ` to the template, so the two places agree.

**The fixture supplies no date.** `decision-context.md` has no date anywhere, so the model has to supply one. That's fine given the template requires the field — just worth being sure the grader reads "and a date" as "the Date section is present and populated" rather than expecting a particular value, since there's nothing in the fixture to anchor it to.

Both are observations rather than objections. The one-line change stands on its own. 👍

mavericksea-ai:
Thanks for checking every claim against main. Both notes are fair.

On `Proposed`: agreed, the template's Status line and the lifecycle disagree. I kept SKILL.md out of this PR because #591 settled that the eval was off, not the skill, but adding `Proposed | ` to the template is a one-line follow-up I'm happy to open if the maintainers want it.

On the date: yes, the intent is that the Date field is present and filled in, not a particular value. All three runs produced dated ADRs and were graded that way. If you'd prefer the expectation to say so, "a populated Date field" is an easy change on this branch.

I also re-authored the commit under this account, so the branch was force-pushed; the one-line change is the same.

PR #600, opened 2026-09-23
fix(constraints): floor-guard reference sees untracked files, deletions and maximum thresholds

### LINKED ISSUE #599: constraint-driven-development: the supplied floor guard passes on untracked suppressions, deleted test files and loosened maximum budgets
`skills/constraint-driven-development/references/floor-guard.md` (main at bcab6a1) supplies a guard and tells agents to adapt it rather than write their own. Its comments promise to include untracked files, catch deleted test files and weakened thresholds, and stay quiet when constraints tighten. I ran the fenced implementation unchanged, with Node, against fourteen small fixtures, each a fresh temporary git repository with one source file, one test file and a CONSTRAINTS.md (a floor bullet, a coverage minimum, a bundle maximum). Seven behave as promised, seven do not:

| Change | Expected exit | Actual exit |
|---|---:|---:|
| Unchanged repository | 0 | 0 |
| Suppression added in tracked file | 1 | 1 |
| Suppression in new untracked file | 1 | **0** |
| Skipped test in new untracked file | 1 | **0** |
| Same suppression in staged new file | 1 | 1 |
| Existing test file deleted, unstaged | 1 | **0** |
| Existing test file deleted, staged | 1 | **0** |
| Assertion removed from surviving test file | 1 | 1 |
| Minimum coverage lowered 80% to 60% | 1 | 1 |
| Minimum coverage raised 80% to 90% | 0 | 0 |
| Maximum bundle size raised 200 KB to 300 KB | 1 | **0** |
| Maximum bundle size reduced 200 KB to 150 KB | 0 | **1** |
| Floor rule deleted | 1 | **0** |
| Tracked exception row added | 1 | 1 |

Four causes, all in the plumbing the reference says to keep unchanged when adapting the language patterns:

1. **Untracked files are discarded.** `git diff --no-index /dev/null <file>` exits 1 whenever there are differences, which is the normal case. The shared `git()` wrapper treats every nonzero exit as failure and returns null, so the caller substitutes an empty diff. Every added-line rule is blind to new files, whatever the patterns say.
2. **Deleted files lose their identity.** The parser takes the filename only from the `+++` line. For a deletion that line is `+++ /dev/null`, and the slice yields `ev/null`, which never matches a test filename. There is no explicit deletion rule either, and staging does not change it.
3. **Numeric direction is hardcoded.** A decreasing number is treated as weaker. That is right for a minimum (coverage) and inverted for a maximum (size, latency, error budget): raising a maximum passes, tightening one fails.
4. **Removed floor bullets are never checked.** A comment promises it; the loop only reports a numeric decrease paired with an added constraints line.

To see whether this reaches generated projects, I ran the skill's constraints setup three times in fresh Claude Code sessions (2.1.278, claude-opus-5) on a small Node project and probed each generated guard in a fresh git repository. All three blocked a deleted regression test and a new untracked suppression: when adapting, the model repaired the plumbing without being asked. One public verbatim copy of the reference (domengabrovsek/claude, `skills/constraints/scripts/floor-guard.mjs`) keeps the original logic and misses both. So the defect lives in the shipped reference and in copies of it, not necessarily in every adapted guard.

The repository's own deterministic gate passes at this commit (140 checks, 88/88 routing prompts), which is expected: it checks the Markdown, not what the program does when run. I searched issues and PRs for `floor-guard`, `untracked` and `threshold` and read the comments on #472, which introduced the reference; #584, #585 and #588 are about references being unavailable or their links, not this. If it was reported somewhere I did not look, apologies.

<details>
<summary>Reproduction (Python 3, Node, git; no network, no model calls, temporary directories only)</summary>

Save the reference file beside the script as `floor-guard.source.md`, then `python3 reproduce.py`. It extracts the fenced block, builds each fixture, runs the guard with `--base HEAD`, records every subprocess call to `results.json`, and prints the table above.

````python
import pathlib,re,subprocess,tempfile,json
HERE=pathlib.Path(__file__).resolve().parent
text=(HERE/'floor-guard.source.md').read_text()
code=re.search(r'```js\n(.*?)\n```',text,re.S).group(1)+'\n'
(HERE/'floor-guard.extracted.mjs').write_text(code)
cases=[
 ('unchanged',0,lambda p:None),
 ('tracked-suppression',1,lambda p:(p/'app.js').write_text('// @ts-ignore\nmodule.exports=1;\n')),
 ('untracked-suppression',1,lambda p:(p/'new.js').write_text('// @ts-ignore\nmodule.exports=1;\n')),
 ('untracked-skipped-test',1,lambda p:(p/'new.test.js').write_text("test.skip('regression',()=>{});\n")),
 ('staged-new-suppression',1,lambda p:(p/'new.js').write_text('// @ts-ignore\nmodule.exports=1;\n')),
 ('deleted-test-file',1,lambda p:(p/'app.test.js').unlink()),
 ('staged-deleted-test-file',1,lambda p:(p/'app.test.js').unlink()),
 ('removed-assertion',1,lambda p:(p/'app.test.js').write_text("const assert=require('assert');\n")),
 ('coverage-relaxed',1,lambda p:(p/'CONSTRAINTS.md').write_text((p/'CONSTRAINTS.md').read_text().replace('| Coverage | >= 80% |','| Coverage | >= 60% |'))),
 ('coverage-tightened',0,lambda p:(p/'CONSTRAINTS.md').write_text((p/'CONSTRAINTS.md').read_text().replace('| Coverage | >= 80% |','| Coverage | >= 90% |'))),
 ('bundle-limit-relaxed',1,lambda p:(p/'CONSTRAINTS.md').write_text((p/'CONSTRAINTS.md').read_text().replace('| Bundle | <= 200 KB |','| Bundle | <= 300 KB |'))),
 ('bundle-limit-tightened',0,lambda p:(p/'CONSTRAINTS.md').write_text((p/'CONSTRAINTS.md').read_text().replace('| Bundle | <= 200 KB |','| Bundle | <= 150 KB |'))),
 ('floor-rule-deleted',1,lambda p:(p/'CONSTRAINTS.md').write_text((p/'CONSTRAINTS.md').read_text().replace('- No skipped tests.\n',''))),
 ('tracked-new-exception',1,lambda p:(p/'CONSTRAINTS.md').write_text((p/'CONSTRAINTS.md').read_text()+'| W1 | temporary waiver | owner | 2026-10-01 |\n')),
]
results=[]
for name,expected,mutate in cases:
 with tempfile.TemporaryDirectory(prefix='floor-') as temp:
  p=pathlib.Path(temp);commands=[]
  def run(argv):
   r=subprocess.run(argv,cwd=p,text=True,capture_output=True)
   commands.append(dict(argv=argv,exit=r.returncode,stdout=r.stdout,stderr=r.stderr))
   return r
  for argv in [['git','init','-q'],['git','config','user.name','Local Test'],['git','config','user.email','test@example.invalid']]:assert run(argv).returncode==0
  (p/'app.js').write_text('module.exports=1;\n')
  (p/'app.test.js').write_text("const assert=require('assert');\nassert.equal(require('./app'),1);\n")
  (p/'CONSTRAINTS.md').write_text('# Constraints\n\n## Floor\n- No skipped tests.\n\n| Dimension | Rule |\n|---|---|\n| Coverage | >= 80% |\n| Bundle | <= 200 KB |\n\n## Exceptions\n')
  assert run(['git','add','.']).returncode==0;assert run(['git','commit','-qm','baseline']).returncode==0
  mutate(p)
  if name.startswith('staged-'):assert run(['git','add','-A']).returncode==0
  actual=run(['node',str(HERE/'floor-guard.extracted.mjs'),'--base','HEAD'])
  results.append(dict(name=name,expected_exit=expected,actual_exit=actual.returncode,matches=actual.returncode==expected,commands=commands))
(HERE/'results.json').write_text(json.dumps(results,indent=2))
for r in results:print(f"{r['name']}: expected {r['expected_exit']}, actual {r['actual_exit']}, {'MATCH' if r['matches'] else 'MISMATCH'}")
print('Summary:',sum(r['matches'] for r in results),'match;',sum(not r['matches'] for r in results),'mismatch')
````

</details>

The fix is contained: keep stdout when `--no-index` exits 1, take old and new file identities from both diff headers and treat `+++ /dev/null` as a deletion, model whether a threshold is a minimum or a maximum, check removed floor bullets as the comment says, and exit 2 when the guard cannot reach a verdict rather than 0. A PR with that fix and these fourteen cases as its test file follows this issue. If you would rather reduce the reference to a description than ship executable guidance, say so and I will close it.

### PR DESCRIPTION
**Pull request by mavericksea-ai**
Fixes #599.

`references/floor-guard.md` promises to include untracked files, catch deleted test files and weakened thresholds, and stay silent when constraints tighten. Run unchanged it did none of those for ordinary local git changes; #599 has the fourteen-case table and the four causes. This PR fixes the plumbing and adds the cases as a test. The three language patterns are untouched.

## What changed in the reference

- **Untracked files.** `git diff --no-index /dev/null <file>` exits 1 whenever the sides differ. The `git()` wrapper now keeps stdout on that exit, and every other failure (no merge base, a diff that cannot be read) exits 2 instead of reading as clean, which the contract already required.
- **Deletions.** Both diff headers name the file, so `+++ /dev/null` keeps the old path. A deleted test file is its own finding (`test-deleted`); the assertion rule skips files that no longer exist so a deletion is reported once.
- **Threshold direction.** Each number in a CONSTRAINTS.md rule carries a direction read from the words around it: `>=`, `at least`, `must not fall` mark a minimum, loosened by going down; `<=`, `at most`, `under`, `must not grow` mark a maximum, loosened by going up. A number with no readable direction is reported whenever it changes (`threshold-changed`), since the guard cannot tell tightening from loosening and staying quiet is the wrong default.
- **Rule identity.** A rule is a floor bullet (its text before the first colon) or a table row (its first cell); the old key was the text before the first pipe, which is empty for every Markdown row. A removed rule with no replacement is reported (`rule-removed`), except an Exceptions row, whose removal tightens.

## Test

`scripts/floor-guard-reference-test.js` extracts the fenced block from the reference and runs it with real git against fourteen fixtures (fresh repositories with a source file, a test file and a CONSTRAINTS.md carrying a floor bullet, a coverage minimum, a bundle maximum and an Exceptions table) plus three edge cases: a removed exception row is silent, a missing merge base exits 2, an undirected number is reported when it changes. On the previous reference 8 of 17 fail; all 17 pass here. The workflow runs it beside the other script tests.

## Verification

- `node --test scripts/floor-guard-reference-test.js`: 17 pass.
- `node scripts/validate-skills.js`: 25 skills, 0 errors; `node scripts/validate-reference-links.js`: passed; `node --test scripts/lib/skill-lint-test.js`: 35 pass.
- The reproduction script from #599, run against this reference: 14 match, 0 mismatch.

## Limits

The reference stays regex-shallow on purpose, as its own note says. Direction words are read from a short window around each number, so an unusual phrasing falls into `threshold-changed` (loud) rather than being misread. Removed-rule detection compares keys, so a reworded bullet is reported as removed and added; that is a review prompt, not a false pass. A removed line whose own text begins with `-- ` is read as a diff header, as it was before.

### CODE REVIEW

**Review by abhisheksharma2411 — COMMENTED** [review]
Hi @mavericksea-ai — this is a good fix, and the framing in #599 (fourteen cases, four causes) is the right way to have approached it. The `git diff --no-index /dev/null` trick for untracked files is neat, and making every *other* git failure exit 2 rather than reading as clean is the correctness call that matters most here — a guard that fails open is worse than no guard.

Your 17 tests pass for me on `4ee1eb6`. I went after the threshold comparison specifically, because direction-from-surrounding-words is the kind of thing that works on the cases you wrote and surprises you on the fifteenth. Three results, one of which I think is worth fixing before merge.

### 1. A threshold deleted from a multi-threshold rule is silent — false negative

This is the one I'd prioritise, because it fails in the unsafe direction.

```js
before.forEach((b, i) => {
  const n = after[i];
  if (verdict || !n || n.n === b.n) return;   // ← `!n` returns early
```

When the edit removes a number, `after[i]` is `undefined` and the loop returns quietly.

```
- Latency: p95 under 200 ms and p99 under 900 ms.
+ Latency: p95 under 200 ms.

floor-guard: clean          ← exit 0
```

The p99 maximum is gone and the rule still matches by key, so nothing fires. Deleting the *whole* rule is caught (`rule-removed`, verified), so this is specifically the "rule survives, one of its thresholds doesn't" case. Given the reference's own stance — *"staying quiet is the wrong default"* when direction can't be read — a vanished threshold seems like it should get the same treatment.

### 2. Positional pairing misfires when a number is inserted — false positive

Thresholds are compared by index, so any number added earlier in the text shifts the alignment:

```
- Coverage: at least 80 percent.
+ Coverage: per ADR 7, at least 80 percent.

[threshold-loosened]   ← exit 1
```

`80` is now compared against `7`, inherits `min` from the *before* side, `7 < 80`, loosened. The rule didn't change. Citing an ADR, adding an issue number, a date, a footnote marker — all trip it.

Pairing on `(direction, position-among-same-direction)` would survive this, or matching on the surrounding token rather than the index. Either is more work than the `!n` fix, so I'd understand deferring it — but it's worth knowing, because the failure mode is a contributor being told they lowered a bar they didn't touch, and the fix they'll reach for is deleting the guard.

### 3. Renaming a rule reads as `rule-removed` — correct, misleading message

```
- Coverage: at least 80 percent.
+ Code coverage: at least 80 percent.

[rule-removed]   ← exit 1
```

I don't think this is a bug — the key *is* the text before the colon, so a renamed rule genuinely can't be matched, and blocking is the safe default. But `rule-removed` on a rule that's plainly still there sends the reader hunting for a deletion that didn't happen. A separate flag, or just appending "(or its label changed — rename in one commit, change thresholds in another)" to the message, would save that trip.

### How I checked

Extracted the fenced block the same way your test does and ran it against fresh repos with real git, with a control in each direction so a pass isn't a coincidence:

```
ok    lowering a minimum is caught                            (exit 1, control)
ok    deleting an entire rule is caught                       (exit 1, control)
FAIL  deleting one threshold from a rule                      (expected 1, got 0)
FAIL  a number added before the threshold                     (expected 0, got 1)
FAIL  renaming a rule label                                   (expected 0, got 1)
```

Happy to push these as three more cases in `floor-guard-reference-test.js` if you'd like — say the word and they're yours, no credit needed. Or if you'd rather land this as-is and take (1) as a follow-up, that seems reasonable too; the plumbing fixes here are clearly a net improvement over what's on `main` regardless.

mavericksea-ai:
Thanks @abhisheksharma2411  for going after the threshold comparison. You were right on all three, and each one reproduced before I changed anything.

Pushed a follow-up commit:

1. A threshold with no counterpart after the edit is now reported as `threshold-removed`, so your p99 case exits 1.
2. Numbers are paired within their direction, minimum with minimum and maximum with maximum, so the `per ADR 7` insertion no longer shifts the pairing and exits 0. An added threshold still counts as tightening and stays silent.
3. The rename still blocks, as you suggested, and the output now adds a line saying a `rule-removed` finding can also mean the label changed.

Your cases are in `floor-guard-reference-test.js`, credited in the commit, plus a control that adding a threshold is silent: 21 pass, and your three fail on the previous commit. One trade-off of pairing by direction: a number whose direction reading changes (say "80 percent" becoming "at least 80 percent") is now reported as removed rather than compared. That's loud rather than a false pass, but worth knowing.

I also re-authored the first commit under this account, so the branch was force-pushed; that commit's content is unchanged.

abhisheksharma2411:
Pulled the follow-up and re-ran my three cases — all fixed, and the `threshold-removed` finding is a better answer than what I suggested, since it names *what* went missing instead of just failing the pair comparison. Thanks for the credit in the test file.

On the trade-off you flagged:

> a number whose direction reading changes (say "80 percent" becoming "at least 80 percent") is now reported as removed rather than compared

I think that's the right call, and worth keeping deliberately rather than treating as a wart. `80 percent` has no direction the guard can read, and the code already reports a directionless number *whenever it changes* on the same reasoning — it can't tell tightening from loosening, so silence is the wrong default. Reporting the transition as removed is the same principle applied to the transition itself: the rule genuinely changed shape, and a human should look once. Loud-and-rare beats quiet-and-wrong here.

The one thing I'd suggest is that the message say so, the way you did for the rename case. Something like "a `threshold-removed` finding can also mean a number gained or lost its direction words — compare the two lines before assuming a threshold was deleted." Otherwise the first person to hit it will read "removed" literally, go looking for a deletion that isn't there, and the fix they reach for is loosening the guard.

Nothing blocking from me. Nice turnaround.

mavericksea-ai:
Added in c1a5cf4: when a threshold-removed finding appears, the output now says it can also mean a number gained or lost its direction words, and to compare the two lines before assuming a deletion. One new case (">= 80%" becoming "80%"); 22 pass. Thanks for the careful review.

